PR #71, opened 2020-06-02
Fix a typo in Multi-GPU guide

salah3x:
In the distributed training guide,  line [367](https://github.com/keras-team/keras-io/blob/6e6ffd12e95c068f0290c9589aa1d6ce721fcb64/guides/distributed_training.py#L367), it should be `since it runs on a single machine and does **not** need to communicate with other machines` instead of `since it runs on a single machine and does need to communicate with other machines`


salah3x:
This closes #62

PR #79, opened 2020-06-05
metric learning eg

matpalm:
A minimal example of metric learning using cosine similarity as logits for a softmax. Unsure the level of description expected on keras.io regarding the approach of using sims as logits, have included what is the minimal info.

### DIFF
--- a/examples/vision/metric_learning.py
+++ b/examples/vision/metric_learning.py
@@ -1,25 +1,23 @@
 """
-Title: Metric learning using cross entropy.
+Title: Metric learning using crossentropy.
 Author: [Mat Kelcey](https://twitter.com/mat_kelcey)
 Date created: 2020/06/05
 Last modified: 2020/06/05
-Description: Example of using metric learning using cross entropy on synthetic data.
+Description: Example of using metric learning using crossentropy on synthetic data.
 """
 """
 ## Overview
 
-Metric learning aims to train models that can embed inputs into a high dimensional space
+Metric learning aims to train models that can embed inputs into a high-dimensional space
 such that "similar" inputs, as defined by the training scheme, are located close to each
 other. These models once trained can produce embeddings for downstream systems where such
 similarity is useful; examples include as a ranking signal for search or as a form of
 pretrained embedding model for another supervised problem.
 
-For a more detailed overview of metric learning see..
+For a more detailed overview of metric learning see:
 
-* [What is metric
-learning?](http://contrib.scikit-learn.org/metric-learn/introduction.html)
-* ["Using cross entropy for metric learning"
-tutorial](https://www.youtube.com/watch?v=Jb4Ewl5RzkI)
+* [What is metric learning?](http://contrib.scikit-learn.org/metric-learn/introduction.html)
+* ["Using crossentropy for metric learning" tutorial](https://www.youtube.com/watch?v=Jb4Ewl5RzkI)
 """
 
 """
@@ -31,10 +29,8 @@ import matplotlib.pyplot as plt
 import numpy as np
 import tensorflow as tf
 from PIL import Image
-from tensorflow.keras.layers import *
-from tensorflow.keras.losses import binary_crossentropy
-from tensorflow.keras.models import Model
-from tensorflow.keras.optimizers import Adam
+from tensorflow import keras
+from tensorflow.keras import layers
 
 """
 ## Dataset
@@ -45,7 +41,7 @@ enough that we can quickly demonstrate concepts.
 """
 
 # Height / Width of all images generated.
-HW = 32
+height_width = 32
 
 
 def random_colour():
@@ -55,10 +51,10 @@ def random_colour():
 
 def random_instance(colour):
     # Start with a black background.
-    img = np.zeros((HW, HW, 3), dtype=np.float32)
+    img = np.zeros((height_width, height_width, 3), dtype=np.float32)
     while True:
         # Pick box extents, ensuring they are ordered.
-        x1, x2, y1, y2 = np.random.randint(0, HW, 4)
+        x1, x2, y1, y2 = np.random.randint(0, height_width, 4)
         if x1 > x2:
             x1, x2 = x2, x1
         if y1 > y2:
@@ -78,25 +74,29 @@ We can visualise a grid of 25 randomly generated examples to get a sense of thes
 
 
 def show_collage(examples):
-    HWB = HW + 2  # Box height / width plus a 2 pixel buffer.
-    n_rows, n_cols = examples.shape[:2]
+    box_size = height_width + 2
+    num_rows, num_cols = examples.shape[:2]
 
     collage = Image.new(
-        mode="RGB", size=(n_cols * HWB, n_rows * HWB), color=(250, 250, 250)
+        mode="RGB",
+        size=(num_cols * box_size, num_rows * box_size),
+        color=(250, 250, 250),
     )
-    for row_idx in range(n_rows):
-        for col_idx in range(n_cols):
+    for row_idx in range(num_rows):
+        for col_idx in range(num_cols):
             array = (np.array(examples[row_idx, col_idx]) * 255).astype(np.uint8)
-            collage.paste(Image.fromarray(array), (col_idx * HWB, row_idx * HWB))
+            collage.paste(
+                Image.fromarray(array), (col_idx * box_size, row_idx * box_size)
+            )
 
     # Double size for visualisation.
-    collage = collage.resize((2 * n_cols * HWB, 2 * n_rows * HWB))
+    collage = collage.resize((2 * num_cols * box_size, 2 * num_rows * box_size))
     return collage
 
 
 # Show a collage of 5x5 random images.
 examples = [random_instance(random_colour()) for _ in range(25)]
-examples = np.stack(examples).reshape((5, 5, HW, HW, 3))
+examples = np.stack(examples).reshape((5, 5, height_width, height_width, 3))
 show_collage(examples)
 
 """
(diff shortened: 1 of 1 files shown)


matpalm:
Yes the toy data is a double edged sword; It's not realistic but very fast to train and get a consistent result. I'll take a look at (a possibly subset of) CIFAR-10 & CelebA but can't promise a 10sec training time!

examples/vision/metric_learning.py
```diff
@@ -0,0 +1,255 @@
...
+
+"""
+## Training loop
+
+We train this model with a custom training loop that first embeds both anchors and
+positives and then uses their pairwise dot products as logits for a softmax. On a Google
+Colab CPU instance this step takes approximately 10 seconds.
+"""
+
+optimiser = Adam(learning_rate=1e-3)
```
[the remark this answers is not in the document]
matpalm: god save the queen

matpalm:
hmm. moving the loss calculation into a train_step call gives me a problem with shape inference around the way it uses the batch size to create a set of labels...

specifically `labels = tf.range(similarities.shape[0])` gives me a problem with None; previously the way it was run the size of the batch was available; now it's still None at this point. i realise it's unusual to want to know the batch size, but this technique requires it.. I'll have a poke around, there's probably a different way to obtain / express this value..

matpalm:
with continue to look, but for now see https://gist.github.com/matpalm/9bfbd138caf8f264c451fdbf1b3ae34a for the error

fchollet:
> specifically labels = tf.range(similarities.shape[0]) gives me a problem with None; previously the way it was run the size of the batch was available; now it's still None at this point. i realise it's unusual to want to know the batch size, but this technique requires it.. I'll have a poke around, there's probably a different way to obtain / express this value..

What's happening is that your training loop is now compiled. It will run faster as a result ;)

If you ever run into `None` shape entires when running compiled code, switch to using `tf.shape(tensor)` instead. That's the symbolic shape, which will never be None at runtime. Kind of a gotcha :/

matpalm:
oh! i tried `tensor.get_shape()` which mistakenly remembered being the same as `tf.shape(tensor)` oops. but, works now! should have time to try another dataset tomorrow. thanks for the thorough review!

matpalm:
switched to CIFAR-10. PTAL.

examples/vision/metric_learning.py
```diff
@@ -0,0 +1,314 @@
...
+            for class_idx in range(num_classes):
+                examples_for_class = class_idx_to_train_idxs[class_idx]
+                anchor_idx = random.choice(examples_for_class)
+                positive_idx = random.choice(examples_for_class)
+                while positive_idx == anchor_idx:
+                    positive_idx = random.choice(examples_for_class)
+                yield x_train[anchor_idx], x_train[positive_idx]
+
+    return (
+        tf.data.Dataset.from_generator(
```
fchollet, line 138: If you're going to use a generator, I recommend using the `Sequence` class, like here: https://keras.io/examples/vision/oxford_pets_image_segmentation/#prepare-sequence-class-to-load-amp-vectorize-batches-of-data

examples/vision/metric_learning.py
```diff
@@ -0,0 +1,314 @@
...
+"""
+
+confusion_matrix = np.zeros((num_classes, num_classes))
+
+# For each class.
+for class_idx in range(num_classes):
+    # Consider 10 examples.
+    example_idxs = class_idx_to_test_idxs[class_idx][:10]
+    for y_test_idx in example_idxs:
+        # And count the classes of it's near neighbours.
```
fchollet, line 294: its

examples/vision/metric_learning.py
```diff
@@ -0,0 +1,314 @@
...
+
+We define a custom model with a `train_step` that first embeds both anchors and positives
+and then uses their pairwise dot products as logits for a softmax.
+"""
+
+
+class EmbeddingModel(keras.Model):
+    def train_step(self, data):
+        data = tf.reshape(data, (2, num_classes, height_width, height_width, 3))
+        anchors, positives = data[0], data[1]
```
matpalm, line 166: the slice call here for the anchors and positives complains if the leading dim for data is None
https://gist.github.com/matpalm/77b06312d383cc79c0a660dcccfaefee

calling a reshape explicitly removes this problem, but seems clumsy :/

examples/vision/metric_learning.py
matpalm: done. but there's one quirk still, around slicing. see my other comment.

matpalm:
Switched to Sequence but one piece I couldn't resolve again around undecided shape.

examples/vision/metric_learning.py
fchollet: This is an issue we're looking at. For the time being you should be able to do:

```
if isinstance(data, tuple):
    data = data[0]
```

matpalm: cool. that works.

matpalm:
OK. I think that's it. Thanks again for the review; it's like a things-you-didn't-know-were-in-Keras guided tour!

examples/vision/metric_learning.py
omalleyt12: I think using the recently exposed (in `tf-nightly`) `tf.keras.utils.unpack_x_y_sample_weight` is the best solution here:

```python
def train_step(self, data):
  x, _, _ = tf.keras.utils.unpack_x_y_sample_weight(data)
```

This will always work for all data formats `Model.fit` accepts

matpalm: I'm not sure. For me that reads as unpack y and weights and ignore them, but a key thing with this example is there is no y, and I know that's confused a number of people I've introduced to these techniques. The vast majority of use of non vanilla pieces in this example; Sequence, custom train_step, etc are around not have vanilla (x, y). I'd hate to let even one reference in.

examples/vision/metric_learning.py
```diff
@@ -0,0 +1,316 @@
...
+For this example we will be using the
+[CIFAR-10](https://www.cs.toronto.edu/~kriz/cifar.html) dataset.
+"""
+
+from tensorflow.keras.datasets import cifar10
+
+(x_train, y_train), (x_test, y_test) = cifar10.load_data()
+
+x_train = x_train.astype("float32") / 255.0
+y_train = np.squeeze(y_train)
```
fchollet, line 49: Why the squeeze? There should be no empty dimensions

examples/vision/metric_learning.py
```diff
@@ -0,0 +1,316 @@
...
+"""
+Metric learning provides training data not as explicit `(X, y)` pairs but instead uses
+multiple instances that are related in the way we want to express similarity. In our
+example we will use instances of the same class to represent similarity; a single
+training instance will not be one image, but a pair of images of the same class. When
+referring to the images in this pair we'll use the common metric learning names of the
+`anchor` (a randomly chosen image) and the `positive` (another randomly chosen image of
+the same class).
+"""
+
```
fchollet, line 99: Please connect these 2 text blocks

examples/vision/metric_learning.py
```diff
@@ -0,0 +1,316 @@
...
+            # We use these similarities as logits for a softmax. The labels for
+            # this call are just the sequence [0, 1, 2, ..., num_classes] since we
+            # want the main diagonal values, which correspond to the anchor/positive
+            # pairs, to be high. This loss will move embeddings for the
+            # anchor/positive pairs together and move all other pairs apart.
+            sparse_labels = tf.range(num_classes)
+            loss = self.compiled_loss(sparse_labels, similarities)
+
+            # Calculate gradients and apply via optimizer.
```
fchollet, line 194: Lines below this line do not need to be under the `GradientTape` scope

examples/vision/metric_learning.py
```diff
@@ -0,0 +1,316 @@
...
+## Embedding model
+
+We define a custom model with a `train_step` that first embeds both anchors and positives
+and then uses their pairwise dot products as logits for a softmax.
+"""
+
+
+class EmbeddingModel(keras.Model):
+    def train_step(self, data):
+        # Note: Workaround for open issue, to be removed.
```
fchollet, line 165: Another thing to consider would be to have your generator return a dict `{"anchors": ..., "positives": ...}`, that way you don't need this and your code will be more explicit

examples/vision/metric_learning.py
```diff
@@ -0,0 +1,316 @@
...
+from tensorflow.keras.datasets import cifar10
+
+(x_train, y_train), (x_test, y_test) = cifar10.load_data()
+
+x_train = x_train.astype("float32") / 255.0
+y_train = np.squeeze(y_train)
+x_test = x_test.astype("float32") / 255.0
+y_test = np.squeeze(y_test)
+
+x_train.shape, y_train.shape, x_test.shape, y_test.shape
```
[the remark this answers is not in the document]
matpalm: It just of debug so might remove them

examples/vision/metric_learning.py
matpalm: `y_train` is `(50000, 1)` & `y_test` is `(10000, 1)`. just want `(50000,)` to make the enumerate later a bit easier.

examples/vision/metric_learning.py
```diff
@@ -0,0 +1,316 @@
...
+            x[1, class_idx] = x_train[positive_idx]
+        return x
+
+
+"""
+We can visualise a batch in another collage. The top row shows randomly chosen anchors
+from the 10 classes, the bottom row shows the corresponding 10 positives.
+"""
+
+for examples in AnchorPositivePairs(num_batchs=1):
```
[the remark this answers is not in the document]
matpalm: `TypeError: 'AnchorPositivePairs' object is not an iterator` ?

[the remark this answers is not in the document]
matpalm: ahh. `examples = next(iter(AnchorPositivePairs(num_batchs=1)))` works. i'll go with that.

examples/vision/metric_learning.py
matpalm: looks like would stil need the workaround though; in that case it returns a single element tuple, with the dictionary in it

matpalm:
no problem, thanks for the review! the contrib guide says to add the assets generated by `add_example` as a final step before the merge. did you still want those?

PR #91, opened 2020-06-15
Add ocr example

AakashKumarNain:


examples/vision/captcha_ocr.py
```diff
@@ -0,0 +1,502 @@
...
+valid_data_generator = DataGenerator(
+    data=validation_data,
+    labels=validation_labels,
+    char_map=char_to_labels,
+    batch_size=batch_size,
+    img_width=img_width,
+    img_height=img_height,
+    downsample_factor=downsample_factor,
+    max_length=max_length,
+    shuffle=False,
```
fchollet, line 334: By line 334 you've downloaded and prepared the data. That's a lot of code for data.

Here's what I recommend:

- make list of image paths and list of their labels
- filter them
- convert them to 2 datasets -- you can use `image_dataset_from_directory` with the `labels` argument, or do it by hand like this: https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/keras/preprocessing/image_dataset.py#L209-L233
- map `tf.strings.split` and an `IndexLookup` layer on the labels dataset to convert the strings to integers

This takes at most 20 lines and it will be more performant

examples/vision/captcha_ocr.py
```diff
@@ -0,0 +1,502 @@
...
+    dataset, test_size=0.1, random_state=seed
+)
+
+training_data = training_data.reset_index(drop=True)
+validation_data = validation_data.reset_index(drop=True)
+print("Number of training samples: ", len(training_data))
+print("Number of validation samples: ", len(validation_data))
+
+# Map characters to numbers
+char_to_labels = {char: idx for idx, char in enumerate(characters)}
```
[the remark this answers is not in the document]
AakashKumarNain: This is a great suggestion. But IIRC, it's in `experimental` phase as of now, right?

examples/vision/captcha_ocr.py
```diff
@@ -0,0 +1,502 @@
...
+
+"""
+## Model
+"""
+
+
+class CTCLayer(layers.Layer):
+    def __init__(self, name=None):
+        super().__init__(name=name)
+        self.loss_fn = keras.backend.ctc_batch_cost
```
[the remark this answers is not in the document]
AakashKumarNain: Oh okay. I think we should provide warnings for the same then

AakashKumarNain:
Thanks @fchollet for the review. I will make the necessary changes

examples/vision/captcha_ocr.py
AakashKumarNain: This functionality is in nightly. Also, the current implementation(the dev branch in nightly) checks if the `labels` argument is an instance of `list` or `tuple`, no checks for `None`. So, if you pass `labels=None`, it will throw error

fchollet: Sorry, I mean `label_mode=None`. You can leave `labels` to its default value.

It's fine to use functionality that's in tf-nightly and experimental. It will soon be in the core API.

AakashKumarNain: I have incorporated these changes. Though instead of using `image_from_directory`, I did it manually but it hardly is 20-25 LOC

examples/vision/captcha_ocr.py
```diff
@@ -0,0 +1,502 @@
...
+        # On test time, just return the computed loss
+        return loss
+
+
+def build_model():
+    # Inputs to the model
+    input_img = layers.Input(
+        shape=(img_width, img_height, 1), name="input_data", dtype="float32"
+    )
+    labels = layers.Input(name="input_label", shape=[max_length], dtype="float32")
```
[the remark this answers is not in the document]
AakashKumarNain: If I use `shape=(None,)`, and infer the shape in CTC layer using `tf.shape(y_true)[0]`, it throws error. Here is the code:
```python
labels = layers.Input(name='input_label', shape=(None,), dtype='float32')
...
output = CTCLayer(name='ctc_loss')(labels, x)
model = Model([input_img, labels], output)

class CTCLayer(layers.Layer):
    def __init__(self, name=None):
        super().__init__(name=name)
        self.loss_fn = keras.backend.ctc_batch_cost

    def call(self, y_true, y_pred):
        # Compute the training-time loss value and add it
        # to the layer using `self.add_loss()`.
        input_length = tf.shape(y_true)[0]
        label_length = tf.shape(y_pred)[0]
        loss = self.loss_fn(y_true, y_pred, input_length, label_length)
        self.add_loss(loss)
        
        # On test time, just return the computed loss
        return loss
```

If I compile this, then it throws this error:
```
ValueError: squeeze_dims[0] not in [0,0). for '{{node ctc_loss/Squeeze_4}} = Squeeze[T=DT_INT32, squeeze_dims=[-1]](ctc_loss/strided_slice_7)' with input shapes: [].

```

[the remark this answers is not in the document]
srihari-humbarwadi: If i am not wrong `input_length = tf.shape(y_true)[0]` returns the number of samples in the batch, and `tf.shape(y_true)[1]` should give you the input_length which is what you need, similarly for `label_length`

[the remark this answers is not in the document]
AakashKumarNain: My bad. That was a typo but it doesn't change anything because both are `None`

AakashKumarNain:
@fchollet I have made all the changes except for removing the `keras.backend` part. I need to check if I need to change anything if I am directly using `tf.nn` instead of `keras.backend`

examples/vision/captcha_ocr.py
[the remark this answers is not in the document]
AakashKumarNain: I looked at the backend code for CTC. It does a heavy lifting. Plus there is no official code example showcasing how to use `tf.nn.ctc_loss`. I have always used it via keras backend. If you have any particular code example I should look at, please let me know

examples/vision/captcha_ocr.py
```diff
@@ -0,0 +1,334 @@
...
+from collections import Counter
+from sklearn.model_selection import train_test_split
+
+import tensorflow as tf
+from tensorflow import keras
+from tensorflow.keras import layers
+
+seed = 1234
+np.random.seed(seed)
+tf.random.set_seed(seed)
```
fchollet, line 39: You don't need to seed these code examples.

examples/vision/captcha_ocr.py
```diff
@@ -0,0 +1,334 @@
...
+    vocabulary=list(characters), num_oov_indices=0, mask_token=None
+)
+
+# Mapping numbers back to original characters
+num_to_char = layers.experimental.preprocessing.StringLookup(
+    vocabulary=char_to_num.get_vocabulary(), mask_token=None, invert=True
+)
+
+# Splitting data into training and validation sets
+X_train, X_valid, y_train, y_valid = train_test_split(
```
fchollet, line 115: Use lowercase for all variable names (here: `x_train, x_val`)

examples/vision/captcha_ocr.py
```diff
@@ -0,0 +1,334 @@
...
+    # 1. Read image
+    img = tf.io.read_file(img_path)
+    img = tf.io.decode_png(img, channels=3)
+    # 2. Convert to grayscale
+    img = tf.image.rgb_to_grayscale(img)
+    # 3. Convert to float32 in [0, 1] range
+    img = tf.image.convert_image_dtype(img, tf.float32)
+    # 4. Resize to the desired size
+    img = tf.image.resize(img, [img_height, img_width])
+    # 5. Transpose the image
```
fchollet, line 130: Why?

examples/vision/captcha_ocr.py
```diff
@@ -0,0 +1,334 @@
...
+X_train, X_valid, y_train, y_valid = train_test_split(
+    np.array(images), np.array(labels), test_size=0.1, random_state=seed, shuffle=True
+)
+
+
+def encode_single_sample(img_path, label):
+    # 1. Read image
+    img = tf.io.read_file(img_path)
+    img = tf.io.decode_png(img, channels=3)
+    # 2. Convert to grayscale
```
fchollet, line 124: Can't do the grayscale conversion as part of image loading?

examples/vision/captcha_ocr.py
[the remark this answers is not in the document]
fchollet: Ok, we can keep it this way for now

examples/vision/captcha_ocr.py
```diff
@@ -0,0 +1,334 @@
...
+    output_text = []
+    for res in results:
+        res = tf.strings.reduce_join(num_to_char(res)).numpy().decode("utf-8")
+        output_text.append(res)
+    return output_text
+
+
+#  Let's check results on some validation samples
+for batch in valid_data_generator.take(1):
+    X_data = batch["input_image"]
```
fchollet, line 315: image_batch / labels_batch?

examples/vision/captcha_ocr.py
```diff
@@ -0,0 +1,334 @@
...
+        self.add_loss(loss)
+
+        # On test time, just return the computed loss
+        return y_pred
+
+
+def build_model():
+    # Inputs to the model
+    input_img = layers.Input(
+        shape=(img_width, img_height, 1), name="input_image", dtype="float32"
```
fchollet, line 208: "images" is shorter (and `labels` below)

examples/vision/captcha_ocr.py
```diff
@@ -0,0 +1,334 @@
...
+
+"""shell
+curl -LO https://github.com/AakashKumarNain/CaptchaCracker/raw/master/captcha_images.tar.gz
+tar -zxf captcha_images.tar.gz
+"""
+
+
+"""
+The dataset contains 1070 captcha files as png images. The label for each sample is the
+name of the file (excluding the '.png' part). We will load the data as numpy arrays. The label
```
fchollet, line 55: NumPy

fchollet:
Note that the motivation for using `tf.nn` APIs is that they're documented and better-maintained. The backend APIs are undocumented, have not been touched in some time, and will eventually be deprecated. But for now, this will do.

examples/vision/captcha_ocr.py
AakashKumarNain: Valid question. The images are in this dataset have a shape `(50, 200, 3)`. If we don't transpose, then after  the two conv blocks and reshape, we would have a shape of `(12, 3200)` (as compared to `(50, 768)` when using transposed version). The number of features that the RNN can attend to is very small and the network fails to converge/perform very poorly.  Please let me know if you want me to add this as a comment

fchollet: Yes, please add a comment -- otherwise it's non-obvious. Basically you want the time dimension to correspond to the width of the image

examples/vision/captcha_ocr.py
```diff
@@ -0,0 +1,327 @@
...
+    vocabulary=list(characters), num_oov_indices=0, mask_token=None
+)
+
+# Mapping numbers back to original characters
+num_to_char = layers.experimental.preprocessing.StringLookup(
+    vocabulary=char_to_num.get_vocabulary(), mask_token=None, invert=True
+)
+
+# Splitting data into training and validation sets
+x_train, x_valid, y_train, y_valid = train_test_split(
```
fchollet, line 107: Please do this by hand with NumPy to avoid the dependency on sklearn. It's only 4 lines or so

examples/vision/captcha_ocr.py
```diff
@@ -0,0 +1,327 @@
...
+"""
+## Let's test-drive it
+"""
+
+
+# Get the prediction model by extracting layers till the output layer
+prediction_model = keras.models.Model(
+    model.get_layer(name="image").input, model.get_layer(name="dense2").output
+)
+print(prediction_model.summary())
```
fchollet, line 289: Just `.summary()`, no need for `print()`

AakashKumarNain:
Thanks for the review. I have made all the changes. 

> Note that the motivation for using tf.nn APIs is that they're documented and better-maintained. The backend APIs are undocumented, have not been touched in some time, and will eventually be deprecated. But for now, this will do.

Totally agree with you. My only concern here is that for some APIs, we haven't provided clear cut code examples for their usage (for example CTC related functionalities). So, both things are required: removing the backend and improving the documentation in `nn.` package

AakashKumarNain:
Thanks for the review

PR #99, opened 2020-06-17
Create speaker_recognition_using_cnn.py

fbadine:
A proposition for a new example that does speaker classification or speaker recognition using 1D Convolutional network


examples/generative/speaker_recognition_using_cnn.py
```diff
@@ -0,0 +1,486 @@
...
+"""
+
+keras.backend.clear_session()
+
+np.random.seed(42)
+tf.random.set_seed(42)
+
+
+def build_model(input_shape, num_classes):
+    model = keras.models.Sequential(
```
fchollet, line 352: In general, prefer using the Functional API. Your model is pretty deep, so consider adding residual connections. Note that it is generally not necessary to pass the `name` argument to very layer; usually it's only useful for input and output layers.

examples/generative/speaker_recognition_using_cnn.py
```diff
@@ -0,0 +1,486 @@
...
+
+print("Training noise size: {}".format(train_noises.shape[0]))
+print("Validation noise size: {}".format(valid_noises.shape[0]))
+
+"""
+## Data Generator
+"""
+
+
+class SpeechGenerator(keras.utils.Sequence):
```
fchollet, line 199: Consider using `tf.data` to load the data:

- start from file paths and their labels
- read_file https://www.tensorflow.org/api_docs/python/tf/io/read_file
- decode_wav https://www.tensorflow.org/api_docs/python/tf/audio/decode_wav
- fft https://www.tensorflow.org/api_docs/python/tf/signal/fft

Check out the code for `image_dataset_from_directory` for some inspiration.

This makes the example more scalable and more future-proof. It will also remove the dependency on scipy and librosa.

fbadine:
Thank you for the review!
I will work on the changes

examples/generative/speaker_recognition_using_cnn.py
fbadine: All worked fine for reading audio waves using tf.audio.decode_wav for all the speakers’ samples in the dataset. In fact, all of those are sampled at sr=16000 
However for noise samples, those are at a different sampling rate (22050) this is why I used librosa which allows resampling while loading audio.
As an alternative, I believe we have 2 solutions:
- use tensorflow_io resample function (a requirement to have tfio)
- resample all audio files (6 files overall) using ffmpeg for example before using the tutorial.

Which one do you think is better for the case of this example?

Thanks!

fchollet: What would be your preference?

Since there's no need for the resampling to be online, I would think the second option (use of ffmpeg on the files on disk before making the dataset) is the simplest. But up to you!

fbadine: That’s exactly what I did today.
I resampled them on disk cause it’s easier + online resampling is not important for the purpose of the example.
Thanks!

examples/generative/speaker_recognition_using_cnn.py
```diff
@@ -0,0 +1,520 @@
...
+noise_paths, _, _ = tf.python.keras.preprocessing.dataset_utils.index_directory(
+    DATASET_NOISE_PATH, "inferred", FILE_FORMATS, shuffle=False
+)
+
+# In this section:
+# - We load all noise samples which should have been resampled to 16000
+# - We split those noise samples to chuncks of 16000 samples which
+#   correspond to 1 second duration each
+
+i = tf.constant(0)
```
fchollet, line 144: Use explicit variable names rather `i` or `sr`

examples/generative/speaker_recognition_using_cnn.py
```diff
@@ -0,0 +1,520 @@
...
+                    os.path.join(DATASET_ROOT, folder),
+                    os.path.join(DATASET_AUDIO_PATH, folder),
+                )
+
+"""
+## Noise Preparation
+"""
+
+# Let's first get the list of all noise files
+noise_paths, _, _ = tf.python.keras.preprocessing.dataset_utils.index_directory(
```
fchollet, line 135: This is a private API, do not use it

examples/generative/speaker_recognition_using_cnn.py
```diff
@@ -0,0 +1,520 @@
...
+    label_mode="int",
+    class_names=None,
+    desired_channels=-1,
+    desired_samples=-1,
+    batch_size=32,
+    shuffle=True,
+    seed=None,
+    validation_split=None,
+    subset=None,
+    follow_links=False,
```
fchollet, line 197: Please restrict the options to what you will actually use here. This is a code example, not a new API.

This could be a single-use function -- or it could even be in-lined in the outer file.

examples/generative/speaker_recognition_using_cnn.py
```diff
@@ -0,0 +1,520 @@
...
+    validation_split=VALID_SPLIT,
+    subset="validation",
+)
+
+valid_ds = valid_ds.map(
+    lambda x, y: (audio_to_fft(x, noises, scale=1.0), y),
+    num_parallel_calls=tf.data.experimental.AUTOTUNE,
+)
+
+print(model.evaluate(valid_ds))
```
fchollet, line 516: Please add some form of results visualization (or audition)

examples/generative/speaker_recognition_using_cnn.py
```diff
@@ -0,0 +1,520 @@
...
+"""
+
+import os
+import shutil
+import numpy as np
+
+import tensorflow as tf
+from tensorflow import keras
+
+from tensorflow.python.keras.preprocessing import dataset_utils
```
fchollet, line 37: This is a private API, do not use it. Reimplement the bits you will need

examples/generative/speaker_recognition_using_cnn.py
```diff
@@ -0,0 +1,520 @@
...
+
+# Where the dataset is located
+# Get the data from https://www.kaggle.com/kongaevans/speaker-recognition-dataset/download
+DATASET_ROOT = "/home/fadi/Downloads/16000_pcm_speeches"
```
fchollet, line 41: Prefer using `HOME + /Downloads` so that others will be able to run the code without modifying the Python file, just by downloading the data

examples/generative/speaker_recognition_using_cnn.py
fbadine: Thanks for the review!
This is done in the latest update

examples/generative/speaker_recognition_using_cnn.py
fbadine: Sorry, I did not know I shouldn't use it. I implemented the bits that I want.

examples/generative/speaker_recognition_using_cnn.py
fbadine: I simplified the functions and reduced the arguments to the minimum needed

examples/generative/speaker_recognition_using_cnn.py
fbadine: I took random samples, ran model.predict, displayed the true and predicted speakers along with possibility to listen to the audios

examples/generative/speaker_recognition_using_cnn.py
fbadine: Done

examples/generative/speaker_recognition_using_cnn.py
fchollet: In general, do not import anything from `tensorflow.python`, stick to the public API

examples/generative/speaker_recognition_using_cnn.py
```diff
@@ -0,0 +1,540 @@
...
+
+- We prepare a dataset of speech samples from different speakers, with the speaker as label.
+- We add background noise to these samples to augment our data.
+- We take the FFT of these samples.
+- We train a 1D convnet to predict the correct speaker given a noisy FFT speech sample.
+
+Note:
+
+- This example should be run with TensorFlow 2.3 or higher, or `tf-nightly`.
+- The noise samples in the dataset need to be resampled to a sampling rate of 16000 Hz
```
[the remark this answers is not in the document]
fbadine: Thanks for the review!

As discussed in https://github.com/keras-team/keras-io/pull/99#discussion_r443256765, I did those offline using ffmpeg command line. On colab I use !ffmpeg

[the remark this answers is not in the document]
fbadine: But since the script assumes we have a dataset downloaded from Kaggle and then splits it into audio and noise folder, maybe it makes more sense now to resample as part of the python script.
Otherwise, the code will ignore any noise that is not at 16000 while the user might not have noticed that resampling is needed beforehand.
What do you think? Do you have any suggestion on what to use? 
Thanks!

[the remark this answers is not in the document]
fchollet: Please include the resampling. It doesn't matter if it's command line code or Python code. You can put command lines in your script by using a shell block (see tutobooks docs)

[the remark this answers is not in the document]
fbadine: Ok. Will do that.
Thanks!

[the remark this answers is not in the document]
fbadine: Command line included

examples/generative/speaker_recognition_using_cnn.py
```diff
@@ -0,0 +1,540 @@
...
+            )
+    else:
+        print("Sampling rate for {} is incorrect. Ignoring it".format(noise_paths[i]))
+
+    return index + 1, noises
+
+
+cond = lambda index, noises: tf.less(index, len(noise_paths))
+
+_, noises = tf.while_loop(cond, read_and_process_noise, [index, noises])
```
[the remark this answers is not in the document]
fbadine: I used the code that you suggested, changing only .append(sample) to .extend(sample)

examples/generative/speaker_recognition_using_cnn.py
```diff
@@ -0,0 +1,540 @@
...
+        len(noise_paths), noises.shape[0], noises.shape[1] // SAMPLING_RATE
+    )
+)
+
+"""
+## Data Generation
+"""
+
+
+def audio_dataset_from_directory(
```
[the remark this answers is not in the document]
fbadine: Removed it and used your suggestion

examples/generative/speaker_recognition_using_cnn.py
```diff
@@ -0,0 +1,540 @@
...
+            elif folder in NOISE_FOLDERS:
+                shutil.move(
+                    os.path.join(DATASET_ROOT, folder),
+                    os.path.join(DATASET_NOISE_PATH, folder),
+                )
+            else:
+                shutil.move(
+                    os.path.join(DATASET_ROOT, folder),
+                    os.path.join(DATASET_AUDIO_PATH, folder),
+                )
```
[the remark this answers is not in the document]
fbadine: I added the directory structure (before and after) and removed an unnecessary if statement but not more.
Does it still seem complex for the example? Or is it good?

examples/generative/speaker_recognition_using_cnn.py
```diff
@@ -0,0 +1,540 @@
...
+# The folders in which we will put the audio samples and the noise samples
+AUDIO_SUBFOLDER = "audio"
+NOISE_SUBFOLDER = "noise"
+
+DATASET_AUDIO_PATH = os.path.join(DATASET_ROOT, AUDIO_SUBFOLDER)
+DATASET_NOISE_PATH = os.path.join(DATASET_ROOT, NOISE_SUBFOLDER)
+
+# Folders containing the noise samples in the
+# original dataset
+NOISE_FOLDERS = ["other", "_background_noise_"]
```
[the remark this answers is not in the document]
fbadine: Removed NOISE_FOLDERS however I need those when creating the audio and noise directories in order to know which directories are considered as noise. So I hardcoded them

examples/generative/speaker_recognition_using_cnn.py
```diff
@@ -0,0 +1,540 @@
...
+
+# Supported audio file format
+FILE_FORMATS = ".wav"
```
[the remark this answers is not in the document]
fbadine: Remove both of them

examples/generative/speaker_recognition_using_cnn.py
```diff
@@ -0,0 +1,465 @@
...
+        len(noise_paths), len(os.listdir(DATASET_NOISE_PATH))
+    )
+)
+
+"""
+Resample all noise samples to 16000 Hz
+"""
+
+"""shell
+!dataset_noise_path=~/Downloads/16000_pcm_speeches/noise; \
```
fchollet, line 183: Don't prefix lines with `!`, the `!` is added automatically when running the example

examples/generative/speaker_recognition_using_cnn.py
```diff
@@ -0,0 +1,465 @@
...
+noises = tf.stack(noises)
+
+print(
+    "{} noise files were split into {} noise samples where each is {} sec. long".format(
+        len(noise_paths), noises.shape[0], noises.shape[1] // SAMPLING_RATE
+    )
+)
+
+"""
+## Data Generation
```
fchollet, line 228: Dataset generation

examples/generative/speaker_recognition_using_cnn.py
fbadine: Thank you for the review and approval.
One quick question before I upload the remaining files:
Will the script add ! To each line? In this case I think it won’t work cause it is a series of shell commands that I have split into lines using \ in order for the line not to be extremely huge

fchollet: Yeah, that's unfortunate. I would suggest using `os.system` to run the command line in a Python cell in this case.

fbadine: Thanks! Will do this.
Shall I still upload all files? Or only .py?

fchollet: Sure, please add the generated files.

examples/generative/speaker_recognition_using_cnn.py
```diff
@@ -0,0 +1,460 @@
...
+    "Found {} files belonging to {} directories".format(
+        len(noise_paths), len(os.listdir(DATASET_NOISE_PATH))
+    )
+)
+
+"""
+Resample all noise samples to 16000 Hz
+"""
+
+command = "for dir in `ls -1 " + DATASET_NOISE_PATH + "`; do "
```
fchollet, line 182: Note that you can put the entire command on a single text string:

```
command = (
    "for dir in `ls -1 " + DATASET_NOISE_PATH + "`; do "
    "for file in `ls -1 " + DATASET_NOISE_PATH + "/$dir/*.wav`; do "
    etc
)
```

fbadine: Thanks! I will do that and upload the files again.

fbadine:
Thanks for all the help and tips to enhance the example!!

fbadine:
I just noticed that the Github and Colab links on the https://keras.io/examples/audio/speaker_recognition_using_cnn/ give a page not found error
I assume it’s because the example was created in generative and now is under ‘audio’ so the hyperlinks point to ‘generative’
Do the files need to be generated again?

PR #133, opened 2020-07-04
Update vae.py: do not use global encoder and decoder

PetterS:
They are attributes of the `VAE` class.


venergiac:
maybe 

            z_mean, z_log_var, z = self.encoder(data)
            reconstruction = self.decoder(z)

PR #159, opened 2020-07-21
Adding an example of time series classification from scratch

hfawaz:
This example shows how to do time series classification from scratch, starting from raw
CSV time series files on disk, without leveraging pre-trained weights or a pre-made Keras
Application model.


examples/timeseries/timeseries_classification_from_scratch.py
```diff
@@ -0,0 +1,231 @@
...
+Author: [hfawaz](https://github.com/hfawaz/)
+Date created: 2020/07/21
+Last modified: 2020/07/21
+Description: Training a time series classifier from scratch on the Coffee dataset from the UCR/UEA archive.
+"""
+"""
+## Introduction
+
+This example shows how to do time series classification from scratch, starting from raw
+CSV time series files on disk, without leveraging pre-trained weights or a pre-made Keras
```
fchollet, line 12: We have no Keras Applications for timeseries, so the second part of this sentence is not necessary.

examples/timeseries/timeseries_classification_from_scratch.py
```diff
@@ -0,0 +1,231 @@
...
+Application model. We demonstrate the workflow on Coffee dataset from the [UCR/UEA
```
fchollet, line 13: Fix markdown link (break the line before the link)

examples/timeseries/timeseries_classification_from_scratch.py
```diff
@@ -0,0 +1,231 @@
...
+"""
+
+epochs = 1000
+mini_batch_size = 8
+
+callbacks = [
+    keras.callbacks.ModelCheckpoint(
+        "best_model.h5", save_best_only=True, monitor="loss"
+    ),
+    keras.callbacks.ReduceLROnPlateau(
```
fchollet, line 179: Are you sure this callback is necessary? How about using a learning rate schedule object instead?

examples/timeseries/timeseries_classification_from_scratch.py
```diff
@@ -0,0 +1,231 @@
...
+model.compile(
+    optimizer=keras.optimizers.Adam(), loss="binary_crossentropy", metrics=["accuracy"],
+)
+hist = model.fit(
+    x_train,
+    y_train,
+    batch_size=mini_batch_size,
+    epochs=epochs,
+    callbacks=callbacks,
+    validation_data=(x_test, y_test),
```
fchollet, line 192: Prefer using validation_split for validation, and test the model on the test data as a separate step after training

examples/timeseries/timeseries_classification_from_scratch.py
```diff
@@ -0,0 +1,231 @@
...
+enc = preprocessing.OneHotEncoder(categories="auto")
+
+enc.fit(np.concatenate((y_train, y_test), axis=0).reshape(-1, 1))
+y_train = enc.transform(y_train.reshape(-1, 1)).toarray()
+y_test = enc.transform(y_test.reshape(-1, 1)).toarray()
+
+"""
+## Build a model
+
+We build a Fully Convolutonal Neural Network originally proposed in [this
```
fchollet, line 115: Fix typos and fix markdown link formatting

examples/timeseries/timeseries_classification_from_scratch.py
```diff
@@ -0,0 +1,231 @@
...
+using a simple reshaping via numpy.
+This will allow us to construct a model that is easily applicable to multivariate time
+series.
+"""
+
+x_train = x_train.reshape((x_train.shape[0], x_train.shape[1], 1))
+x_test = x_test.reshape((x_test.shape[0], x_test.shape[1], 1))
+
+"""
+Finally, in order to have a standard labeling of the data we will use the enconding from
```
fchollet, line 102: You don't need to. You can use sparse_categorical_crossentropy instead. Just count how many categories you have in the data beforehand

examples/timeseries/timeseries_classification_from_scratch.py
```diff
@@ -0,0 +1,231 @@
...
+plt.figure()
+for c in classes:
+    c_x_train = x_train[y_train == c]
+    plt.plot(c_x_train[0], label="class" + str(c))
+plt.legend(loc="best")
+plt.show()
+plt.close()
+
+"""
+## Standardizing the data
```
fchollet, line 72: Use preprocessing layers for this step https://keras.io/guides/preprocessing_layers/

examples/timeseries/timeseries_classification_from_scratch.py
```diff
@@ -0,0 +1,231 @@
...
+
+import tensorflow as tf
+from tensorflow import keras
+from tensorflow.keras import layers
+import numpy as np
+import matplotlib.pyplot as plt
+from sklearn import preprocessing
+
+"""
+## Load the data: the Coffee dataset
```
fchollet, line 31: You don't explain what the dataset is about. Please add an explanation and provide basic data exploration (how many samples for training and for validation? How many classes in the dataset? Are the classes balanced?)

examples/timeseries/timeseries_classification_from_scratch.py
```diff
@@ -0,0 +1,231 @@
...
+allows us to demonstrate effectively how to use ConvNets for time series classification.
+In this file, the first column corresponds to the label.
+"""
+
+
+def readucr(filename):
+    data = np.loadtxt(filename, delimiter="\t")
+    Y = data[:, 0]
+    X = data[:, 1:]
+    return X, Y
```
fchollet, line 47: Use lowercase variable names everywhere

examples/timeseries/timeseries_classification_from_scratch.py
```diff
@@ -0,0 +1,231 @@
...
+The implementation is based on the TF2.0 version provided
+[here](https://github.com/hfawaz/dl-4-tsc/).
+
+"""
+
+
+def make_model(input_shape, num_classes):
+    input_layer = keras.layers.Input(input_shape)
+
+    # this is the first 1D convolutional block
```
fchollet, line 126: Nit: all comments should start with a capital letter and end with a period.

examples/timeseries/timeseries_classification_from_scratch.py
```diff
@@ -0,0 +1,231 @@
...
+
+model = make_model(input_shape=x_train.shape[1:], num_classes=y_train.shape[1])
+keras.utils.plot_model(model, show_shapes=True)
+
+"""
+## Train the model
+
+"""
+
+epochs = 1000
```
fchollet, line 172: How many epochs do you actually need?

examples/timeseries/timeseries_classification_from_scratch.py
```diff
@@ -0,0 +1,231 @@
...
+mini_batch_size = 8
```
fchollet, line 173: Just batch_size (prefer something larger than 8)

examples/timeseries/timeseries_classification_from_scratch.py
hfawaz: Indeed it is possible to replace it with a learning rate schedule object. 
However, since we are implementing the FCN model proposed originally in this [paper](https://arxiv.org/abs/1611.06455), we want to conserve the same model as much as possible.

examples/timeseries/timeseries_classification_from_scratch.py
hfawaz: At the end of the notebook, we comment the train/val loss plots: 

_We can see how the training loss decreases and reaches almost zero after 100 epochs. 
Nevertheless, by observing the validation loss we can see how the network still needs training until it reaches almost perfect loss on both the validation and the training loss after 400 epochs._

examples/timeseries/timeseries_classification_from_scratch.py
hfawaz: We used 12 instead of 8. 
Since the training set contains 28 instances, with ```validation_split =0.2```, we are bounded to low batch size values.

examples/timeseries/timeseries_classification_from_scratch.py
hfawaz: By looking at the Normalization preprocessing layer, it says: 
_The batch axis, 0, is always summed over (```axis=0``` is not allowed)._
However in this case, we want to apply znormalization to each time series independently, similar toInstanceNormalization.
So should I write a custom preprocessing layer for this one and include it in the file ?

examples/timeseries/timeseries_classification_from_scratch.py
```diff
@@ -0,0 +1,241 @@
...
+import matplotlib.pyplot as plt
+from sklearn.metrics import accuracy_score
+
+"""
+## Load the data: the Coffee dataset
+
+### Description
+The dataset we are using here is called Coffee.
+The data comes from the UCR archive.
+The dataset contains 28 training instances and another 28 testing instances.
```
fchollet, line 35: I had not caught that. In this case I'm afraid the problem setup is inappropriate. The model used here has something like 250,000 floating point parameters. But there are only 28 training data samples. For such a small problem you would use a very small model (e.g [Conv1D(8), Conv1D(num_classes), GlobalMaxPooling()]), not a large model like that. Or even a model that isn't deep-learning based. Deep learning is pattern recognition and as such it requires a reasonable amount of training data.

I would recommend switching to using a problem where each class has at least a few hundreds of samples. Then the problem will be more meaningful.

hfawaz: I agree. In fact, I was implementing in this tutorial, a very standard architecture with a very small dataset. 
I will therefore switch to another dataset with few hundred samples per class. 
It should not require big changes to the code. 
Should I open a new pull request or settle for a new commit here ?

fchollet: You could just update this PR when you have a new dataset. For the dataset, the bigger the better. Thank you!

examples/timeseries/timeseries_classification_from_scratch.py
```diff
@@ -0,0 +1,247 @@
...
+
+
+def readucr(filename):
+    data = np.loadtxt(filename, delimiter="\t")
+    y = data[:, 0]
+    x = data[:, 1:]
+    return x, y.astype(int)
+
+
+root_url = "https://raw.githubusercontent.com/hfawaz/cd-diagram/master/FordA/"
```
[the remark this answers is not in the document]
hfawaz: The compressed archive contains 128 datasets. 
This tutorial is for one of the 128 datasets called FordA, so in order to avoid downloading the whole archive, I uploaded FordA on GitHub to use it in this tutorial. 
The archive is not that large (260 MB), so do you think it is better to download and unzip the archive online in the notebook ? 
Note: the authors of the archive require people searching for a password in a paper to unlock the archive.

examples/timeseries/timeseries_classification_from_scratch.py
```diff
@@ -0,0 +1,247 @@
...
+    conv1 = keras.layers.Conv1D(filters=128, kernel_size=8, padding="same")(input_layer)
+    # Following the convolution, we apply a batchnormalization in order to
+    # help the network converge quickly while adding an implicit regularization.
+    conv1 = keras.layers.BatchNormalization()(conv1)
+    # Finally we add the relu activation function to inject some non-linearity
+    # to the convolutions' output.
+    conv1 = keras.layers.Activation(activation="relu")(conv1)
+
+    # A second convolutional block is applied.
+    conv2 = keras.layers.Conv1D(filters=256, kernel_size=5, padding="same")(conv1)
```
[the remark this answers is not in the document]
hfawaz: Indeed, perhaps KerasTuner would give some insights for choosing the best hyperparameters. 
I haven't tried it though. I simply used the same model/hyperparameters published [here](https://ieeexplore.ieee.org/abstract/document/7966039).
Nevertheless it would be interesting to see what KerasTuner would give on this dataset. 
Should I apply it "offline" or include it in this tutorial ?

examples/timeseries/timeseries_classification_from_scratch.py
```diff
@@ -0,0 +1,247 @@
...
+plt.show()
+plt.close()
+
+"""
+## Standardizing the data
+
+Our time series are already in a single length (176). However, their values are
+usually in various ranges. This is not ideal for a neural network;
+in general we should seek to make the input values normalized. Here, we will
+standardize values to have a mean equal to zero and a standard deviation equal to one.
```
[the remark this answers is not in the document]
hfawaz: In fact, I was applying what is recommended by current state-of-the-art for time series classification ([link](https://link.springer.com/article/10.1007/s10618-016-0483-9)). 
Either way, the current dataset (FordA) is already normalized (each instance has a mean equal to zero and a standard deviation equal to one).  
So given that the dataset is already z-normalized, should I remove the normalization procedure from the tutorial?

examples/timeseries/timeseries_classification_from_scratch.py
hfawaz: This comment should be addressed after solving [this](https://github.com/keras-team/keras-io/pull/159#discussion_r475484573).

examples/timeseries/timeseries_classification_from_scratch.py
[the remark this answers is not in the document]
fchollet: Ok, sounds good, let's keep your GitHub link then!

examples/timeseries/timeseries_classification_from_scratch.py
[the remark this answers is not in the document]
fchollet: Yes, you can remove it and add a note (e.g. a link to the article) saying that some form of normalization is usually required

examples/timeseries/timeseries_classification_from_scratch.py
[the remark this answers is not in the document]
fchollet: Please use it offline to find the best model parameters. You can then add a comment in the code explaining how you found these parameters.

examples/timeseries/timeseries_classification_from_scratch.py
```diff
@@ -0,0 +1,237 @@
...
+plt.close()
+
+"""
+## Standardizing the data
+
+Our time series are already in a single length (176). However, their values are
+usually in various ranges. This is not ideal for a neural network;
+in general we should seek to make the input values normalized.
+For this specific dataset, the data is already z-normalized: each time series sample
+have a mean equal to zero and a standard deviation equal to one. This type of
```
fchollet, line 84: has

examples/timeseries/timeseries_classification_from_scratch.py
```diff
@@ -0,0 +1,237 @@
...
+Now we shuffle the training set because we will be using the `validation_split` option
+later when training.
+"""
+
+idx = np.random.permutation(len(x_train))
+x_train = x_train[idx]
+y_train = y_train[idx]
+
+"""
+Standardize the labels to positive integers.
```
fchollet, line 116: Add a comment that the expected labels will then be 0 and 1 (integers)

examples/timeseries/timeseries_classification_from_scratch.py
```diff
@@ -0,0 +1,237 @@
...
+
+"""
+## Build a model
+
+We build a Fully Convolutional Neural Network originally proposed in
+[this paper](https://arxiv.org/abs/1611.06455).
+The implementation is based on the TF2.0 version provided
+[here](https://github.com/hfawaz/dl-4-tsc/).
+The following hyperparameters (kernel_size, filters, the usage of BatchNorm) were found
+using the random search from [KerasTuner](https://github.com/keras-team/keras-tuner).
```
fchollet, line 130: "found via random search using KerasTuner"

examples/timeseries/timeseries_classification_from_scratch.py
```diff
@@ -0,0 +1,237 @@
...
+
+"""
+
+
+def make_model(input_shape):
```
fchollet, line 135: I would recommend removing the comments in the make_model function. They aren't essential.

examples/timeseries/timeseries_classification_from_scratch.py
```diff
@@ -0,0 +1,237 @@
...
+model = make_model(input_shape=x_train.shape[1:])
+keras.utils.plot_model(model, show_shapes=True)
+
+"""
+## Train the model
+
+"""
+
+epochs = 500
+batch_size = 128
```
fchollet, line 183: Because you only have 3601*0.8 training samples, you end up with 22 gradient update steps with a batch size of 128. This is very little, hence why you need to train for so many epochs. I would recommend using a smaller batch size, e.g. 32.

examples/timeseries/timeseries_classification_from_scratch.py
```diff
@@ -0,0 +1,237 @@
...
+plt.ylabel(metric, fontsize="large")
+plt.xlabel("epoch", fontsize="large")
+plt.legend(["train", "val"], loc="best")
+plt.show()
+plt.close()
+
+"""
+We can see how the training accuracy increases and reaches almost 0.9 after 10 epochs. 
+Nevertheless, by observing the validation accuracy we can see how the network still needs
+training until it reaches almost 0.9 for both the validation and the training accuracy
```
fchollet, line 235: This suggests that the network hasn't started overfitting. After how many epochs do you start seeing overfitting?

hfawaz: I have removed early stopping to observe the overfitting phenomena (with a batch size of 32 as you suggested). 
It seems like after the 200th epoch, the validation accuracy stops improving while the training accuracy continues on increasing. 
This would suggest that overfitting starts at epoch 200. 
I updated the description accordingly. 
You can observe the validation and training accuracy below:
image

examples/timeseries/timeseries_classification_from_scratch.py
```diff
@@ -0,0 +1,237 @@
...
+
+"""
+## Evaluate model on test data
+"""
+
+model = keras.models.load_model("best_model.h5")
+
+y_pred = model.predict(x_test).argmax(axis=1)
+
+print("Test accuracy", accuracy_score(y_pred, y_test))
```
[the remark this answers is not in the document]
hfawaz: Indeed I was using `model.evaluate`, however when I used `sparse_categorical_crossentropy` with `models.load_model("best_model.h5")` I observed weird results.
It seems like the fix would be to compile the model with `sparse_categorical_accuracy` instead of `accuracy` ([link](https://github.com/tensorflow/tensorflow/issues/39341#issuecomment-677947225)).

hfawaz:
Great! 
Thanks for the code review and glad to be a contributor to Keras.

PR #168, opened 2020-07-30
Added Kaggle pneumonia example

mjang2000:


examples/vision/xray_classification_with_tpus.py
```diff
@@ -0,0 +1,489 @@
...
+    train_ds,
+    steps_per_epoch=TRAIN_IMG_COUNT // BATCH_SIZE,
+    epochs=EPOCHS,
+    validation_data=val_ds,
+    validation_steps=VAL_IMG_COUNT // BATCH_SIZE,
+    class_weight=class_weight,
+)
+
+"""
+# 7. Finetune the model
```
fchollet, line 396: Fine-tuning is normally done when you have a model trained on a different dataset, that you want to adapt to a new dataset. See for instance this guide https://keras.io/guides/transfer_learning/

Here you are training on the same dataset, with a different learning rate. You could simply add a learning rate schedule in the call to `fit()` above. Please remove this section, and instead  add a learning rate schedule to your original optimizer. See: https://keras.io/api/optimizers/learning_rate_schedules/

examples/vision/xray_classification_with_tpus.py
```diff
@@ -0,0 +1,489 @@
...
+if we can achieve a higher accuracy.
+"""
+
+"""
+# 9. Predict and evaluate results
+
+Let's evaluate the model on our test data!
+"""
+
+loss, acc, prec, rec = model.evaluate(test_ds)
```
fchollet, line 479: Since you have many quantities, prefer using `evaluate(test_ds, return_dict=True)`

examples/vision/xray_classification_with_tpus.py
```diff
@@ -0,0 +1,489 @@
...
+
+"""
+We see that our accuracy on our test data is 83%, which is lower than the accuracy for
+our validating set. This may indicate overfitting. Try finetuning the model further to
+decrease overfitting to the training and validation sets.
```
fchollet, line 484: Training longer will increase overfitting, not decrease it.

examples/vision/xray_classification_with_tpus.py
```diff
@@ -0,0 +1,489 @@
...
+"""
+# 9. Predict and evaluate results
+
+Let's evaluate the model on our test data!
+"""
+
+loss, acc, prec, rec = model.evaluate(test_ds)
+
+"""
+We see that our accuracy on our test data is 83%, which is lower than the accuracy for
```
fchollet, line 482: Please add an example of doing inference (`predict`) on a single test image.

examples/vision/xray_classification_with_tpus.py
```diff
@@ -0,0 +1,489 @@
...
+"""
+
+AUTOTUNE = tf.data.experimental.AUTOTUNE
+GCS_PATH = "gs://kds-63fdbf6403ae1d5e971e6cdbd4104b02e914cf6e90a178bd0e3990a8"
+BATCH_SIZE = 16 * strategy.num_replicas_in_sync
+IMAGE_SIZE = [180, 180]
+EPOCHS = 25
+
+"""
+# 2. Load the data
```
fchollet, line 62: For data loading, you can use:

```python
image_size = (180, 180)
batch_size = 32

train_ds = tf.keras.preprocessing.image_dataset_from_directory(
    "train",
    image_size=image_size,
    batch_size=batch_size,
)
val_ds = tf.keras.preprocessing.image_dataset_from_directory(
    "val",
    image_size=image_size,
    batch_size=batch_size,
)
test_ds = tf.keras.preprocessing.image_dataset_from_directory(
    "test",
    image_size=image_size,
    batch_size=batch_size,
)
class_names = train_ds.class_names
```

examples/vision/xray_classification_with_tpus.py
```diff
@@ -0,0 +1,489 @@
...
+"""
+The images originally have values that range from [0, 255]. CNNs work better with smaller
+numbers so we will scale this down.
+"""
+
+
+def decode_img(img):
+    # convert the compressed string to a 3D uint8 tensor
+    img = tf.image.decode_jpeg(img, channels=3)
+    # Use `convert_image_dtype` to convert to floats in the [0,1] range.
```
fchollet, line 150: Prefer starting the model with a `Rescaling` layer: https://keras.io/api/layers/preprocessing_layers/image_preprocessing/rescaling/

This makes your model portable.

examples/vision/xray_classification_with_tpus.py
```diff
@@ -0,0 +1,489 @@
...
+First, let's use buffered prefetching so we can yield data from disk without having I/O
+become blocking.
+"""
+
+
+def prepare_for_training(ds, cache=True, shuffle_buffer_size=1000):
+    # This is a small dataset, only load it once, and keep it in memory.
+    # use `.cache(filename)` to cache preprocessing work for datasets that don't
+    # fit in memory.
+    if cache:
```
fchollet, line 199: In general for an image dataset it is not recommend to cache it in memory. Prefer downloading the files to local disk and reading them from disk without caching.

examples/vision/xray_classification_with_tpus.py
```diff
@@ -0,0 +1,489 @@
...
+we're building a convolution neural network, we'll create a convolution block and a dense
+layer block.
+
+The architecture for this CNN has been inspired by this
+[article](https://towardsdatascience.com/deep-learning-for-detecting-pneumonia-from-x-ray-images-fc9a3d9fdba8).
+[article](https://towardsdatascience.com/deep-learning-for-detecting-pneumonia-from-x-ray-images-fc9a3d9fdba8).
+"""
+
+
+def conv_block(filters):
```
fchollet, line 265: Prefer using the functional API, e.g.

```python
inputs = keras.Input(shape=(IMAGE_SIZE[0], IMAGE_SIZE[1], 3))
x = preprocessing.Rescaling(1./255)(inputs)
x = layers.Conv2D(16, 3, activation="relu", padding="same")(x)
x = layers.Conv2D(16, 3, activation="relu", padding="same")(x)
...
x = conv_block(x, 32)
etc
```

examples/vision/xray_classification_with_tpus.py
```diff
@@ -0,0 +1,489 @@
...
+"""
+From exploring the data and the model, I noticed that the training for the model has a
+slow start. However, after 25 epochs, the model slowly starts to converge.
+"""
+
+history = model.fit(
+    train_ds,
+    steps_per_epoch=TRAIN_IMG_COUNT // BATCH_SIZE,
+    epochs=EPOCHS,
+    validation_data=val_ds,
```
fchollet, line 390: Don't pass `steps_per_epoch` and `validation_steps`, you don't need to

examples/vision/xray_classification_with_tpus.py
mjang2000: I've tried using `image_dataset_from_directory` but it seems that it doesn't work with TPUs

examples/vision/xray_classification_with_tpus.py
mjang2000: Adding a Rescaling layer changes the dimension so fixed the issue by indexing the image tensor.
```
    inputs = tf.keras.Input(shape=(IMAGE_SIZE[0], IMAGE_SIZE[1], 3)),
    x = tf.keras.layers.experimental.preprocessing.Rescaling(1./255)(inputs)[0]
    x = tf.keras.layers.Conv2D(16, 3, activation='relu', padding='same')(x)
```

examples/vision/xray_classification_with_tpus.py
mjang2000: I left the `conv_block` and `dense_block` as is and changed the `build_model` to the functional API. Please let me know if I should change the blocks as well

examples/vision/xray_classification_with_tpus.py
mjang2000: For TPUs, I don't think we can access local files. All files have to be access through a GCS path. This might be why `image_dataset_from_directory` does not work with TPUs.

examples/vision/xray_classification_with_tpus.py
mjang2000: image

mjang2000: Error message:
```
InvalidArgumentError: NodeDef expected inputs 'string' do not match 0 inputs specified; Op<name=TensorSliceDataset; signature=components: -> handle:variant; attr=Toutput_types:list(type),min=1; attr=output_shapes:list(shape),min=1; is_stateful=true>; NodeDef: {{node TensorSliceDataset}} [Op:MapDataset]
```
Only the accelerator was changed. No change in code.

mjang2000:
Thank you for the helpful feedback!

examples/vision/xray_classification_with_tpus.py
```diff
@@ -0,0 +1,476 @@
...
+As expected, we have two labels for our images.
+"""
+
+CLASS_NAMES = np.array(
+    [
+        str(tf.strings.split(item, os.path.sep)[-1].numpy())[2:-1]
+        for item in tf.io.gfile.glob(str(GCS_PATH + "/chest_xray/train/*"))
+    ]
+)
+CLASS_NAMES
```
fchollet, line 109: In general, this file works both as a standalone Python script and as a notebook, so if you want to print something, use `print(...)`, e.g. here `print('Class names: %s' % (CLASS_NAMES,))`

examples/vision/xray_classification_with_tpus.py
fchollet: Ok, that makes sense. Please include an explanation about how large image datasets should not be cached in memories, but here you do it because you want to train on TPU and the dataset is not very large.

examples/vision/xray_classification_with_tpus.py
fchollet: This is not expected. Please file an issue so we can track this.

mjang2000:
TPU `image_dataset_from_directory` issue posted on GitHub: https://github.com/tensorflow/tensorflow/issues/42010

examples/vision/xray_classification_with_tpus.py
```diff
@@ -0,0 +1,479 @@
...
+
+for f in train_list_ds.take(5):
+    print(f.numpy())
+
+"""
+Run the following cell to see how many images we have in our training dataset and how
+many images we have in our validation set. Verify that the ratio of images is 80:20.
+"""
+
+TRAIN_IMG_COUNT = tf.data.experimental.cardinality(train_list_ds).numpy()
```
fchollet, line 93: I believe you can now use `train_list_ds.cardinality` instead of the experimental function (please check)

examples/vision/xray_classification_with_tpus.py
```diff
@@ -0,0 +1,479 @@
...
+print("Training images count: " + str(TRAIN_IMG_COUNT))
+
+VAL_IMG_COUNT = tf.data.experimental.cardinality(val_list_ds).numpy()
+print("Validating images count: " + str(VAL_IMG_COUNT))
+
+"""
+As expected, we have two labels for our images.
+"""
+
+CLASS_NAMES = np.array(
```
fchollet, line 103: Why make it an array?

In general avoid casting strings to `np.array` because the memory consumption is extremely high and you will OOM even for relatively small lists of strings. If you need a tensor-like object, prefer using `tf.constant`

In this case you can just leave it as list

examples/vision/xray_classification_with_tpus.py
```diff
@@ -0,0 +1,479 @@
...
+def exponential_decay(lr0, s):
+    def exponential_decay_fn(epoch):
+        return lr0 * 0.1 ** (epoch / s)
+
+    return exponential_decay_fn
+
+
+exponential_decay_fn = exponential_decay(0.015, 20)
+
+lr_scheduler = tf.keras.callbacks.LearningRateScheduler(exponential_decay_fn)
```
fchollet, line 413: If you're just doing exponential decay, you can use https://keras.io/api/optimizers/learning_rate_schedules/exponential_decay/
instead of a callback (preferred)

examples/vision/xray_classification_with_tpus.py
```diff
@@ -0,0 +1,479 @@
...
+)
+img = decode_img(img)
+plt.imshow(img / 255)
+
+img_array = tf.keras.preprocessing.image.img_to_array(img)
+img_array = tf.expand_dims(img_array, 0)  # Create batch axis
+
+predictions = model.predict(img_array)
+score = predictions[0]
+print("This scan is %.2f percent pneumonia." % (100 * score))
```
fchollet, line 475: You don't seem to be using CLASS_NAMES anywhere, maybe you could use it here?

examples/vision/xray_classification_with_tpus.py
```diff
@@ -0,0 +1,479 @@
...
+"""
+## Build the CNN
+
+To make our model more modular and easier to understand, let's define some blocks. As
+we're building a convolution neural network, we'll create a convolution block and a dense
+layer block.
+
+The architecture for this CNN has been inspired by this
+[article](https://towardsdatascience.com/deep-learning-for-detecting-pneumonia-from-x-ray-images-fc9a3d9fdba8).
+[article](https://towardsdatascience.com/deep-learning-for-detecting-pneumonia-from-x-ray-images-fc9a3d9fdba8).
```
fchollet, line 244: Duplication

examples/vision/xray_classification_with_tpus.py
```diff
@@ -0,0 +1,479 @@
...
+"""
+
+"""
+## Finetune the model
+
+Finetuning is an art when it comes to Machine Learning, and there are many ways to adjust
+the model in efforts to improve it. Finetuning is beyond the scope of this notebook, but
+check out this
+[article](https://www.pyimagesearch.com/2019/06/03/fine-tuning-with-keras-and-deep-learnin
+g/) for more information.
```
fchollet, line 380: Please fix link formatting break

examples/vision/xray_classification_with_tpus.py
```diff
@@ -0,0 +1,479 @@
...
+
+    model.compile(optimizer="adam", loss="binary_crossentropy", metrics=METRICS)
+
+"""
+From exploring the data and the model, I noticed that the training for the model has a
+slow start. However, after 25 epochs, the model slowly starts to converge.
+"""
+
+"""
+## Finetune the model
```
fchollet, line 374: You probably want to adjust the copy of this section to match the code. Also, keep in mind that fine-tuning is something that you do on a different dataset (transfer learning), so it's not applicable in this context

examples/vision/xray_classification_with_tpus.py
```diff
@@ -0,0 +1,479 @@
...
+"""
+
+checkpoint_cb = tf.keras.callbacks.ModelCheckpoint("xray_model.h5", save_best_only=True)
+
+early_stopping_cb = tf.keras.callbacks.EarlyStopping(
+    patience=10, restore_best_weights=True
+)
+
+"""
+We also want to finetune our learning rate. Too high of a learning rate will cause the
```
fchollet, line 398: "finetune" -> "tune"

mjang2000:
Awesome, thank you for your help!

mjang2000:
I tried to run the autogen script, but it seems that the GCS path is causing some issues.

```
2020-08-07 21:45:41.228611: I tensorflow/core/platform/cloud/google_auth_provider.cc:180] Attempting an empty bearer token since no token was retrieved from files, and GCE metadata check was skipped.
```

examples/vision/xray_classification_with_tpus.py
```diff
@@ -0,0 +1,453 @@
...
+    print("Device:", tpu.master())
+    tf.config.experimental_connect_to_cluster(tpu)
+    tf.tpu.experimental.initialize_tpu_system(tpu)
+    strategy = tf.distribute.experimental.TPUStrategy(tpu)
+except:
+    strategy = tf.distribute.get_strategy()
+print("Number of replicas:", strategy.num_replicas_in_sync)
+
+"""
+We need a Google Cloud link to our data to load the data using a TPU.
```
fchollet, line 34: Please add a couple lines about the pre-requisites to run on TPU: must be on Colab with the TPU runtime selected.

fchollet:
> I tried to run the autogen script, but it seems that the GCS path is causing some issues.

Does that prevent the script from running? Is the file somehow private?

> It also seems that the NB isn't being generated on TPU. Might there be a way to add the NB without using the autogen.py script?

You can instead run the command on a VM with a GPU. It general we need to be able to re-run the notebook in the future and we won't have TPU access.

mjang2000:
Got it to work! It runs on TPU as well.

PR #188, opened 2020-08-07
RealNVP

giorgiozannini:
We are three italian data scientists to be and we would love to know what you think if our project could be a good example!


examples/generative/Density estimation using RealNVP
```diff
@@ -0,0 +1,245 @@
...
+"""
+## Model training
+"""
+
+loss_history = []
+# Adam optimizer
+optimizer = tf.keras.optimizers.Adam(learning_rate=0.00005)
+
+# 2000 epochs
+for i in notebook.tqdm(range(2000)):
```
fchollet, line 197: Let's turn this logic into a more idiomatic Keras workflow -- please factor this into a `train_step` method on a Model subclass -- see https://keras.io/guides/customizing_what_happens_in_fit/

giorgiozannini:
Thanks you very much! We will apply those changes as soon as possible!
Best regards,
Giorgio Zannini Quirini, Daniele Sanna, Giorgio Maria Mandolini

examples/generative/Density estimation using RealNVP.py
```diff
@@ -0,0 +1,251 @@
...
+## Setup
+
+"""
+
+# libraries to import
+import tensorflow as tf
+from tensorflow import keras
+from tensorflow.keras import layers
+from tensorflow.keras import regularizers
+from sklearn.preprocessing import StandardScaler
```
fchollet, line 41: Don't use sklearn. You can do this with a Normalization layer instead. https://keras.io/api/layers/preprocessing_layers/core_preprocessing_layers/normalization/

examples/generative/Density estimation using RealNVP.py
```diff
@@ -0,0 +1,251 @@
...
+data = scaler.fit_transform(data)
+
+"""
+## Affine coupling layer
+"""
+
+# Creating a custom layer with keras API
+
+
+nweights = 256
```
fchollet, line 65: Use fully-spelled out variable names everywhere, for better readability. Here, you can use output_dim

examples/generative/Density estimation using RealNVP.py
```diff
@@ -0,0 +1,251 @@
...
+    )(s_layer_4)
+    return keras.Model(inputs=input, outputs=[s_layer_5, t_layer_5])
+
+
+Coupling(2).summary()
+"""
+## Real NVP
+"""
+# class containing the whole set of operations
+class Realnvp(keras.Model):
```
fchollet, line 111: RealNVP

examples/generative/Density estimation using RealNVP.py
```diff
@@ -0,0 +1,251 @@
...
+    def __init__(self, layers_list, masks, distr):
+        super(Realnvp, self).__init__()
+
+        # number of coupling layers
+        self.num_cl = len(masks)
```
fchollet, line 116: Use fully-spelled out variable names everywhere

examples/generative/Density estimation using RealNVP.py
```diff
@@ -0,0 +1,251 @@
...
+        # distribution of the latent space
+        self.distr = distr
+        # masks to divide into 1:d and d+1:D
+        self.masks = masks
+        # s and t custom layers of before
+        self.layers_list = layers_list
+
+    # custom function defining the forward operation
+    def forward(self, y):
```
fchollet, line 125: Put this in a `call()` method instead, so you can call the model

examples/generative/Density estimation using RealNVP.py
```diff
@@ -0,0 +1,251 @@
...
+plt.title("model loss")
+plt.ylabel("loss")
+plt.xlabel("epoch")
+
+# from data to latent space
+z, _ = model.inverse(data)
+
+# from latent space to data
+samples = distr.sample(3000)
+x, _ = model.forward(samples)
```
fchollet, line 232: Use `model.predict(samples)` (which will work once you put your logic in `call`)

examples/generative/Density estimation using RealNVP.py
```diff
@@ -0,0 +1,251 @@
...
+            print(epoch, "loss", round(logs.get("loss"), 4))
+
+
+# Training model
+history = model.fit(
+    data,
+    batch_size=len(data),
+    epochs=7500,
+    callbacks=[Loss_every_100_epochs()],
+    verbose=0,
```
fchollet, line 215: Just set verbose=1 or 2

examples/generative/Density estimation using RealNVP.py
```diff
@@ -0,0 +1,251 @@
...
+
+# Adam optimizer
+optimizer = tf.keras.optimizers.Adam(learning_rate=0.0001)
+
+model = Realnvp(layers_list, masks, distr)
+
+model.compile(optimizer=optimizer)
+
+
+class Loss_every_100_epochs(keras.callbacks.Callback):
```
fchollet, line 203: This isn't a useful callback, just use the built-in progress bar

examples/generative/Density estimation using RealNVP.py
```diff
@@ -0,0 +1,251 @@
...
+## Initialization
+"""
+
+# initializing the masks, the layers and the prior distribution for the real nvp
+num_cl = 6
+masks = np.array([[0, 1], [1, 0]] * (num_cl // 2), dtype="float32")
+
+layers_list = [Coupling(2) for i in range(num_cl)]
+
+tfd = tfp.distributions
```
fchollet, line 187: No need for this assignment, you can use `tfp.distributions` directly

examples/generative/Density estimation using RealNVP.py
```diff
@@ -0,0 +1,251 @@
...
+        y, logdet = self.inverse(x)
+        return self.distr.log_prob(y) + logdet
+
+    def log_loss(self, x):
+        return -tf.reduce_mean(self.log_likelihood(x))
+
+    def train_step(self, data):
+
+        with tf.GradientTape() as tape:
+            loss = self.log_loss(data)
```
fchollet, line 167: You don't use the forward pass in the loss computation?

examples/generative/Density estimation using RealNVP.py
```diff
@@ -0,0 +1,251 @@
...
+            s, t = self.layers_list[i](x_masked)
+            s *= reversed_mask
+            t *= reversed_mask
+            x = x_masked + reversed_mask * (x * tf.exp(s) + t)
+            log_det_for += tf.reduce_sum(s, [1])
+
+        return x, log_det_for
+
+    # custom function defining the forward operation
+    def inverse(self, x):
```
fchollet, line 141: You can refactor `call` so that both `forward` and `inverse` are covered by the same `call` method. They only differ very little

examples/generative/Density estimation using RealNVP.py
```diff
@@ -0,0 +1,251 @@
...
+        2, activation="tanh", kernel_regularizer=regularizers.l2(reg)
+    )(s_layer_4)
+    return keras.Model(inputs=input, outputs=[s_layer_5, t_layer_5])
+
+
+Coupling(2).summary()
+"""
+## Real NVP
+"""
+# class containing the whole set of operations
```
fchollet, line 110: All comments throughout the file should start with a capital letter and end with a period. Please remove any comment whose content is redundant with what the code says

giorgiozannini:
Sorry for the late reply, we will fix the code ASAP!

examples/generative/Density estimation using RealNVP.py
```diff
@@ -0,0 +1,218 @@
...
+                * (x * tf.exp(direction * s) + direction * t * tf.exp(gate * s))
+                + x_masked
+            )
+            log_det_inv += gate * tf.reduce_sum(s, [1])
+
+        return x, log_det_inv
+
+    # Log likelihood of the normal distribution plus the log determinant of the jacobian.
+    def log_loss(self, x):
+        y, logdet = self.call(x)
```
fchollet, line 142: You can just use `self(x)` here

examples/generative/Density estimation using RealNVP.py
```diff
@@ -0,0 +1,216 @@
...
+loss_tracker = keras.metrics.Mean(name="loss")
+
+optimizer = keras.optimizers.Adam(learning_rate=0.0001)
+
+model = RealNVP(layers_list, masks, distribution, loss_tracker)
+
+model.compile(optimizer=optimizer)
+
+history = model.fit(
+    normalized_data, batch_size=len(normalized_data), epochs=2000, verbose=2
```
fchollet, line 182: I also didn't catch this the first time, but you don't do any validation during training. The only model verification you have is a graph at the end. This leads to the question: how did you pick the number of epochs to train for?

Consider monitoring test performance on a subset of the training data. You could do this by implementing `test_step` and passing `validation_split=0.1` to `fit()`

It's also rather weird that the batch size is the data size. What's the motivation for this? Why not use mini batch training?

examples/generative/Density estimation using RealNVP.py
```diff
@@ -0,0 +1,211 @@
...
+"""
+
+plt.figure(figsize=(15, 10))
+plt.plot(history.history["loss"])
+plt.title("model loss")
+plt.ylabel("loss")
+plt.xlabel("epoch")
+
+# From data to latent space.
+z, _ = model(normalized_data)
```
fchollet, line 191: Use `predict` here, since you don't need to differentiate through the call to `model()`

examples/generative/Density estimation using RealNVP.py
```diff
@@ -0,0 +1,211 @@
...
+
+        g = tape.gradient(loss, self.trainable_variables)
+        self.optimizer.apply_gradients(zip(g, self.trainable_variables))
+        self.loss_tracker.update_state(loss)
+
+        return {"loss": self.loss_tracker.result()}
+
+    def test_step(self, data):
+        loss = self.log_loss(data)
+        self.loss_tracker.update_state(loss)
```
fchollet, line 163: One slight issue with the use of the loss tracker here (and also in train_step) is that it would need to be reset before each epoch and before the call to `evaluate`. Otherwise you just keep adding data to it forever, and that won't be accurate.

The simplest way to achieve this is to make sure the loss tracker shows up in `model.metrics`, in which case it will be reset at all the appropriate times. The easiest way to do this is to implement

```
@property
def metrics(self):
   """List of the model's metrics.

   We make sure the loss tracker is listed as part of `model.metrics`
   so that `fit()` and `evaluate()` are able to `reset()` the loss tracker
   at the start of each epoch and at the start of an `evaluate()` call.
   """
   return [self.loss_tracker]
```

examples/generative/Density estimation using RealNVP.py
```diff
@@ -0,0 +1,217 @@
...
+        self.masks = np.array(
+            [[0, 1], [1, 0]] * (num_coupling_layers // 2), dtype="float32"
+        )
+
+        self.loss_tracker = keras.metrics.Mean(name="loss")
+
+        self.layers_list = [Coupling(2) for i in range(num_coupling_layers)]
+
+    @property
+    def metrics(self):
```
fchollet, line 125: Please include the docstring I provided, it's important for readers to understand why this method exists.

examples/generative/Density estimation using RealNVP.py
fchollet: Please update this

examples/generative/Density estimation using RealNVP.py
fchollet: Please update this

examples/generative/Density estimation using RealNVP.py
```diff
@@ -0,0 +1,222 @@
+"""
+Title: Density estimation using Real NVP
```
fchollet, line 2: Also, please rename the file to `real_nvp.py`

examples/generative/real_nvp.py
```diff
@@ -0,0 +1,222 @@
...
+"""
+
+plt.figure(figsize=(15, 10))
+plt.plot(history.history["loss"])
+plt.title("model loss")
+plt.ylabel("loss")
+plt.xlabel("epoch")
+
+# From data to latent space.
+z, _ = model(normalized_data)
```
fchollet, line 202: Use `predict` here

examples/generative/real_nvp.py
```diff
@@ -0,0 +1,222 @@
...
+        output_dim, activation="relu", kernel_regularizer=regularizers.l2(reg)
+    )(s_layer_3)
+    s_layer_5 = keras.layers.Dense(
+        2, activation="tanh", kernel_regularizer=regularizers.l2(reg)
+    )(s_layer_4)
+
+    return keras.Model(inputs=input, outputs=[s_layer_5, t_layer_5])
+
+
+Coupling(2).summary()
```
fchollet, line 99: This line is not necessary

fchollet:
One last thing: this example is still in need of copy editing. Please reapply edits from https://github.com/keras-team/keras-io/pull/188/commits/c951bc2e26eb58ea5758b161c1055c4ebe39325d (which, for the most part, remain missing)

giorgiozannini:
Thank you so much! It was a pleasure learning from you. We will run the
command immediately.
Best regards,
Giorgio Zannini Quirini, Giorgio Maria Mandolini, Daniele Sanna

giorgiozannini:
We ran the commands. Please let us know if everything is ok or if we need
to do anything else.

fchollet:
Please update the PR with the files.

giorgiozannini:
Sorry to bother you again, but what files should we send here? The real_nvp.py, the generated folders with images and the .md files? Thanks again for your patience.

fchollet:
All files that were created by the command: the ipynb one, the md one, and the image ones.

giorgiozannini:
Thank you very much again for this opportuinity!

PR #245, opened 2020-09-03
fix #243

### LINKED ISSUE #243: Time Series Example "timeseries_anomaly_detection" is broken
With the commit aa45363b1d7e2cb82a675a86ad89abe9bae59390 the File examples/timeseries/timeseries_anomaly_detection.py don’t plot a correct graph at the End, it should show the anomalies on the original graph as expected from comment "Let's overlay the anomalies on the original test data plot." but it doesn’t.

RobAltena:
Fix #243 
For the example to work as a python script all charts need to be isolated and shown with `plt.show()`.

Signed-off-by: Robert Altena <Rob@Ra-ai.com>

### DIFF
--- a/examples/timeseries/timeseries_anomaly_detection.py
+++ b/examples/timeseries/timeseries_anomaly_detection.py
@@ -64,8 +64,9 @@ print(df_daily_jumpsup.head())
 
 We will use the following data for training.
 """
-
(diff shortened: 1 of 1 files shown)


fchollet (approved):
Thank you for the PR!

PR #251, opened 2020-09-18
3D image classification

hasibzunair:
Adds a 3D image classification training and test pipeline using [MosMedData: Chest CT Scans with COVID-19 Related Findings](https://www.medrxiv.org/content/10.1101/2020.05.20.20100362v1). Works on Colab and local jupyter.

- Uses a dataset of CT scans
- Uses Conv3D layers
- On the fly data augmentation of the 3D volumetric data


examples/vision/3D_image_classification.py
```diff
@@ -0,0 +1,454 @@
...
+
+from scipy.ndimage import zoom
+
+
+def read_nifti_file(filepath):
+    # read file
+    scan = nib.load(filepath)
+    # get raw data
+    scan = scan.get_fdata()
+    # rotate
```
fchollet, line 79: Why?

examples/vision/3D_image_classification.py
```diff
@@ -0,0 +1,454 @@
...
+left_labels = np.array([0 for _ in range(len(left_scans))])
+
+# merge classes
+inputs = np.concatenate((right_scans, left_scans), axis=0)
+labels = np.concatenate((right_labels, left_labels), axis=0)
+labels = keras.utils.to_categorical(labels, 2)
+print("Size of raw data and labels:", inputs.shape, labels.shape)
+
+# split data in the ratio 70-30 for training and testing
+from sklearn.model_selection import train_test_split
```
fchollet, line 204: You don't need sklearn for this, you can just do it by hand

examples/vision/3D_image_classification.py
```diff
@@ -0,0 +1,454 @@
...
+
+# for the CT scans having cancerous regions in the right side
+# assign 1, similarly for left assign 0.
+right_labels = np.array([1 for _ in range(len(right_scans))])
+left_labels = np.array([0 for _ in range(len(left_scans))])
+
+# merge classes
+inputs = np.concatenate((right_scans, left_scans), axis=0)
+labels = np.concatenate((right_labels, left_labels), axis=0)
+labels = keras.utils.to_categorical(labels, 2)
```
fchollet, line 200: They're binary, you don't need categorical encoding

examples/vision/3D_image_classification.py
```diff
@@ -0,0 +1,454 @@
...
+plt.imshow(img[:, :, 5], cmap="gray")
+
+"""
+Since a CT scan has many slices, let's visualize a montage of the slices.
+"""
+
+
+def plot_slices(data):
+    # plot a montage of 20 CT slices
+    a, b = 2, 10
```
fchollet, line 153: Everywhere, please use fully-spelled out variable names for better readability

examples/vision/3D_image_classification.py
```diff
@@ -0,0 +1,454 @@
...
+print(
+    "After train and test split:",
+    x_train.shape,
+    y_train.shape,
+    x_test.shape,
+    y_test.shape,
+)
+
+"""
+## Preprocessing and data augmentation
```
fchollet, line 218: Everything from this point on (normalization, data augmentation) should be done with tf.data.Dataset, not Numpy. In particular the augmentation should be done on the fly so as not to use too much RAM

examples/vision/3D_image_classification.py
```diff
@@ -0,0 +1,454 @@
...
+
+    outputs = layers.Dense(units=2, activation="softmax")(x)
+
+    # define the model
+    model = keras.Model(inputs, outputs)
+    return model
+
+
+# free up RAM
+keras.backend.clear_session()
```
fchollet, line 369: You don't need this in this example, it will clear very little. It's only justified if you have a for loop where you create dozens of models

examples/vision/3D_image_classification.py
```diff
@@ -0,0 +1,454 @@
...
+
+    x = layers.Conv3D(filters=128, kernel_size=3, activation="relu")(x)
+    x = layers.MaxPool3D(pool_size=2)(x)
+    x = layers.BatchNormalization()(x)
+
+    x = layers.Flatten()(x)
+    x = layers.Dense(units=256, activation="relu")(x)
+    x = layers.Dropout(0.1)(x)
+
+    outputs = layers.Dense(units=2, activation="softmax")(x)
```
fchollet, line 361: 1, sigmoid (binary labels)

examples/vision/3D_image_classification.py
```diff
@@ -0,0 +1,454 @@
...
+    x = layers.MaxPool3D(pool_size=2)(x)
+    x = layers.BatchNormalization()(x)
+
+    x = layers.Conv3D(filters=128, kernel_size=3, activation="relu")(x)
+    x = layers.MaxPool3D(pool_size=2)(x)
+    x = layers.BatchNormalization()(x)
+
+    x = layers.Flatten()(x)
+    x = layers.Dense(units=256, activation="relu")(x)
+    x = layers.Dropout(0.1)(x)
```
fchollet, line 359: Use a higher value, at least 0.3

examples/vision/3D_image_classification.py
```diff
@@ -0,0 +1,454 @@
...
+Description: Train a 3D convolutional neural network to classify location of cancer.
+"""
+"""
+## Introduction
+
+This example will show the steps needed to build a 3D image classification
+model using a 3D convolutional neural network (CNN) to predict the location
+of cancerous regions in CT scans.
+
+## Why 3D CNNs at all?
```
fchollet, line 15: It would be useful to provide paper references here

fchollet (commented):
Thank you for the PR! This will make a great new example 👍

examples/vision/3D_image_classification.py
hasibzunair: The files are in Nifti format which consists of other metadata about the CT scan/study. `nib.load(filepath)` loads the file and then the image data is extracted using `scan.get_fdata()` as a Numpy array. This is what is normally done. Maybe I should add these as comments on the script, no?

examples/vision/3D_image_classification.py
hasibzunair: I could use tf.data.Dataset without augmentation. There is an issue to convert to Numpy from Tensor and back. Those specific functions are required for augmenting 3D volumes. Any pointers on how to do this?

hasibzunair: I figured it out! Will be adding the changes soon.

examples/vision/3D_image_classification.py
hasibzunair: Done!

examples/vision/3D_image_classification.py
hasibzunair: Done!

examples/vision/3D_image_classification.py
hasibzunair: Done!

examples/vision/3D_image_classification.py
hasibzunair: Done!

examples/vision/3D_image_classification.py
hasibzunair: Removed!

examples/vision/3D_image_classification.py
hasibzunair: Fixed!

examples/vision/3D_image_classification.py
hasibzunair: Removed!

examples/vision/3D_image_classification.py
hasibzunair: Done!

examples/vision/3D_image_classification.py
```diff
@@ -0,0 +1,469 @@
...
+model = get_model((128, 128, 128, 1), 2)
+model.summary()
+
+"""
+## Train model
+"""
+
+# compile model
+initial_learning_rate = 0.0001
+lr_schedule = tf.keras.optimizers.schedules.ExponentialDecay(
```
fchollet, line 401: Since you import `keras`, you can replace `tf.keras` with `keras` throughout this file (saves 3 char)

examples/vision/3D_image_classification.py
```diff
@@ -0,0 +1,469 @@
...
+
+    outputs = layers.Dense(units=1, activation="sigmoid")(x)
+
+    # define the model
+    model = keras.Model(inputs, outputs, name="3dcnn")
+    return model
+
+
+# build model
+model = get_model((128, 128, 128, 1), 2)
```
fchollet, line 392: Pass keyword args for better readability

examples/vision/3D_image_classification.py
```diff
@@ -0,0 +1,469 @@
...
+
+
+
+To make the model easier to understand, blocks are defined. Since this is a
+3D CNN, 3D convolutions are used. The architecture of the 3D CNN used in this example
+is based on this [paper](https://arxiv.org/abs/2007.13224).
+"""
+
+
+def get_model(img_size, num_classes):
```
fchollet, line 359: You don't use num_classes. If you want to use it, you need an if statement

```
if num_classes == 2:
   outputs = layers.Dense(units=1, activation="sigmoid")(x)
else:
   outputs = layers.Dense(units=num_classes, activation="softmax")(x)
```

examples/vision/3D_image_classification.py
```diff
@@ -0,0 +1,469 @@
...
+
+    x = layers.Conv3D(filters=64, kernel_size=3, activation="relu")(x)
+    x = layers.MaxPool3D(pool_size=2)(x)
+    x = layers.BatchNormalization()(x)
+
+    x = layers.Conv3D(filters=128, kernel_size=3, activation="relu")(x)
+    x = layers.MaxPool3D(pool_size=2)(x)
+    x = layers.BatchNormalization()(x)
+
+    x = layers.Flatten()(x)
```
fchollet, line 380: Consider using a global pooling layer (GlobalAveragePooling2D), otherwise you end up with very large vectors

examples/vision/3D_image_classification.py
```diff
@@ -0,0 +1,469 @@
...
+batch_size = 2
+# augment the on the fly during training
+train_dataset = train_dataset.shuffle(len(x_train)).map(train_augment).batch(batch_size)
+# only rescale
+test_dataset = test_dataset.shuffle(len(x_test)).map(test_augment).batch(batch_size)
+
+# data = train_dataset.take(1)
+# print(data)
+# images, labels = list(data)[0]
+# images.shape, labels.shape
```
fchollet, line 329: Debugging code?

examples/vision/3D_image_classification.py
```diff
@@ -0,0 +1,469 @@
...
+    elif augmentation == "normal":
+        pass
+    else:
+        print("No augmentation selected!")
+    # normalize
+    volume = tf_normalize(volume)
+    return volume, label
+
+
+def test_augment(volume, label):
```
fchollet, line 302: It's not augmentation, just preprocessing, so name it `test_preprocessing` (resp. train)

examples/vision/3D_image_classification.py
```diff
@@ -0,0 +1,469 @@
...
+    return volume
+
+
+@tf.function(input_signature=[tf.TensorSpec((128, 128, 128), tf.float64)])
+def tf_blur(volume):
+    augmented_volume = tf.numpy_function(blur, [volume], tf.float64)
+    return augmented_volume
+
+
+def train_augment(volume, label):
```
fchollet, line 283: Let's simplify these 2 functions:

- normalization should be done with TF ops, not numpy. Only rotation and blur require scipy.
- no need for conditionals
- It seems very restrictive to hardcode image dimensions in the tensorspec, remove them.
- the scipy functions can be inlined inside their tf wrapper

fchollet (commented):
Thank you for the update!

examples/vision/3D_image_classification.py
```diff
@@ -0,0 +1,469 @@
...
+"""
+### Make predictions on a single CT scan
+"""
+
+# load best weights
+model.load_weights("3d_image_classification.h5")
+prediction = model.predict(np.expand_dims(x_test[0], axis=0))[0]
+scores = [1 - prediction[0], prediction[0]]
+
+CLASS_NAMES = ["left", "right"]
```
fchollet, line 464: Use a lowercase name

examples/vision/3D_image_classification.py
```diff
@@ -0,0 +1,464 @@
...
+"""
+
+# define data loaders
+train_loader = tf.data.Dataset.from_tensor_slices((x_train, y_train))
+test_loader = tf.data.Dataset.from_tensor_slices((x_test, y_test))
+
+batch_size = 2
+# augment the on the fly during training
+train_dataset = (
+    train_loader.shuffle(len(x_train)).map(train_preprocessing).batch(batch_size)
```
fchollet, line 307: Also add prefetching `.prefetch()`

examples/vision/3D_image_classification.py
```diff
@@ -0,0 +1,453 @@
...
+    train_dataset,
+    validation_data=test_dataset,
+    epochs=epochs,
+    shuffle=True,
+    verbose=2,
+    callbacks=[checkpoint_cb, early_stopping_cb],
+)
+
+"""
+It is important to note that the number of samples is very small (only 40) and we don't specify a random seed.
```
fchollet, line 413: Is it possible to train on more samples? Here the model has many orders of magnitude more parameters than we have samples.

examples/vision/3D_image_classification.py
```diff
@@ -0,0 +1,453 @@
...
+checkpoint_cb = keras.callbacks.ModelCheckpoint(
+    "3d_image_classification.h5", save_best_only=True
+)
+early_stopping_cb = keras.callbacks.EarlyStopping(monitor="val_acc", patience=10)
+
+# train the model, doing validation at the end of each epoch.
+epochs = 100
+model.fit(
+    train_dataset,
+    validation_data=test_dataset,
```
fchollet, line 405: If you're going to use it as validation dataset, please name it `validation_dataset` to avoid confusion.

fchollet (commented):
Looking good! I have applied some copyedits, please pull them first.

examples/vision/3D_image_classification.py
hasibzunair: @fchollet, there is space limitation. I could add the 40 scans as a GitHub release file to be used easily used on a single notebook. I used Dropbox previously and then got blocked for downloading it many times. I was not aware of that! How about reducing the complexity of the model?

examples/vision/3D_image_classification.py
hasibzunair: Done!

examples/vision/3D_image_classification.py
hasibzunair: Named everything validation instead of test.

examples/vision/3D_image_classification.py
fchollet: Have you tried the model to convergence on the full dataset (or a subset) before? What results did you get?

hasibzunair: On the full data I get an accuracy of 70%, for the subset it is around 60%. In both cases there is large variance.

fchollet: Please add this information to the text here:

- How many samples are in the full dataset
- Results obtained on the full dataset
- Some quantification of the variance in each case (results will be in the range...)

hasibzunair: I used another dataset for the example which give slightly more reliable results. Also added the details.

examples/vision/3D_image_classification.py
```diff
@@ -0,0 +1,482 @@
...
+Findings](https://www.medrxiv.org/content/10.1101/2020.05.20.20100362v1). This dataset
+consists of lung CT scans with COVID-19 related findings, as well as without such
+findings.
+
+We will be using the associated radiological findings of the CT scans as labels to build
+a classifier to predict presence of viral pneumonia.
+Hence, the task is a binary classification problem.
+"""
+
+# download url of normal CT scans
```
fchollet, line 54: One last nit: make sure all comments in the file start with a capital and end with a period.

fchollet (approved):
Looks good to me! Please address the last nit and then you can add the autogenerated files. Almost ready to merge 👍

examples/vision/3D_image_classification.py
```diff
@@ -0,0 +1,482 @@
...
+
+2D CNNs are commonly used to process RGB images (3 channels). A 3D CNN is simply the 3D
+equivalent:
+it takes as input a 3D volume or a sequence of 2D frames (e.g. slices in a CT scan),
+3D CNNs are a powerful model for learning representations for volumetric data.
+
+## References
+
+- [A survey on Deep Learning Advances on Different 3D
+DataRepresentations](https://arxiv.org/pdf/1808.01462.pdf)
```
fchollet, line 22: Also, fix this link formatting, it won't render properly. Links should be on one line.

examples/vision/3D_image_classification.py
hasibzunair: Could you please check before I add the generated files?

examples/vision/3D_image_classification.py
hasibzunair: Done!

hasibzunair:
@fchollet, I added the generated files. Please check.

fchollet (approved):
I'll take it from here. Thank you for the great contribution! 👍

fchollet:
Update: the validation accuracy for this example does not exceed that of random predictions (~0.5). We should fix that.

The 80+% accuracy you got on the full dataset was the validation accuracy, right?

hasibzunair:
This is the training log of another run of the same script.
image

Yes, it was the validation accuracy. Training was about 95+%.

fchollet:
Do you get any better results when removing the intermediate Dense layer and lowering the number of conv channels to 64?

hasibzunair:
It was almost similar in both the datasets I experimented with. The training time reduced, when I used GAP layer instead of Dense (thanks for suggesting that!).

hasibzunair:
Here are two other runs, on Colab and local machine. 

Screenshot from 2020-09-25 19-57-10

Screenshot from 2020-09-26 00-16-16

In any case, it was always better than random predictions at the end of training. Not sure what the issue is. Let me know if I should look into it.

fchollet:
We need to do something to get better (& more reliable) metrics here, before we can publish. Maybe a smaller model, maybe a different dataset, maybe a different label set. In terms of making the model smaller, you could remove the Dense(512) layer as well as making the conv layers much smaller (fewer filters).

hasibzunair:
I'll get back with some updates sometime next week.

hasibzunair:
@fchollet, I made a new PR with the changes. Please check.

PR #252, opened 2020-09-18
Update text_extraction_with_bert.ipynb

tomerk:
Add a step to install Huggingface transformers when running on colab


fchollet (commented):
Thank you for the PR. Please edit the Python file instead. Don't include the pip install statement, but you can include text that says "This example requires the Transformers library, which you can install via `pip install transformers`"

PR #265, opened 2020-09-25
Update masked_language_modeling colab and github links

liopic:
A recent change https://github.com/keras-team/keras-io/commit/7c7d09ba9651dd8a05ccb70b2a0520d0a34c68cc updated some filenames, but the links to "view in colab" and "github source" were not updated.


fchollet (approved):
LGTM, thanks

PR #272, opened 2020-10-02
Fix link to question

Offpics:
This PR fixes broken link for question section "What do "sample", "batch", and "epoch" mean?"


fchollet (approved):
LGTM, thanks for the PR

PR #288, opened 2020-10-25
Correcting filter size typo

dimitreOliveira:
At the "Build the generators" the last layer is said to have `filter size of 1` but actually, it also has `filter size of 7` as the first layer, what may have caused the type is that it has `stride size of 1`.

Also the code implementation is used `filter size of 7`.


fchollet (commented):
LGTM, thanks for the PR!

Please replicate the fix in the `.md` and `.ipynb` files.

dimitreOliveira:
You're welcome @fchollet , the pleasure is mine.

Just updated the missing files.

fchollet (approved):

PR #293, opened 2020-11-01
Make code style consistent

soroushj:
There was a code style inconsistency in an example for the Functional API.
This change makes the example easier to read.


fchollet (commented):
Thanks for the PR. The code is auto-formatted by Black, not formatted by hand. This statement fits on a single line because it is short enough.

PR #300, opened 2020-11-05
Add missing parameter in code example

albertvillanova:


MarkDaoust (approved):
Thanks.

MarkDaoust (changes requested):
This change is good.

Can you also make the matching changes to the generated IPYNB and MD files?

Thanks.

MarkDaoust (approved):

PR #304, opened 2020-11-07
replace global batch_size with self.batch_size

Fluxionsdx:
Change two instances of batch_size to self.batch size in dataset class get_item() method; only works as written because there's a globally defined batch_size variable


fchollet (commented):
Thank you for the PR. Please replicate the fix in the corresponding `.md` and `.ipynb` files.

fchollet (approved):
LGTM, thank you

PR #313, opened 2020-11-16
Fix the number of classes in the Oxford Pets segmentation example

xianlopez:
I've seen the number of classes set in the Oxford Pets image segmentation example is 4, whilst the ground truth there has only three different labels. I imagine this was due to the fact that the labels are 1, 2 and 3, so inputing that directly to the sparse cross entropy loss would require a network with 4 outputs per pixel (as if there were a label 0).

Subtracting 1 to the labels allows to keep the network how it has to be, with three outputs.


fchollet (commented):
Sounds good, please regenerate the autogenerated files via the `add_example` command.

xianlopez:
Amended the commit to avoid black format error.

xianlopez:
I ran the `add_example` function and added a new commit with the autogenerated changes.

fchollet (approved):
Thank you, LGTM

PR #318, opened 2020-11-30
Update serialization_and_saving.py

philyoun:
1. Trivial bracket typo.
2. Changed the order.
3. I can see the sentence 'Below is an example of ... from he SavedModel format without overwriting the config methods.' on the page, but somehow I can't find it in the sourcecode. Hope 'he -> the'.

I read it many times, it's been a great help!
So I can resume the training from the where I left even it's a subclassed model with a lot of custom-layers and a bunch of compilation information, if I use the SavedFormat and define the `get_config()`?

### DIFF
--- a/guides/ipynb/serialization_and_saving.ipynb
+++ b/guides/ipynb/serialization_and_saving.ipynb
@@ -104,7 +104,7 @@
     "\n",
     "- The model's architecture/config\n",
     "- The model's weight values (which were learned during training)\n",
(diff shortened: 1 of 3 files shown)


MarkDaoust (approved):

MarkDaoust:
>  I can't find it in the sourcecode. Hope 'he -> the'.

The site might just be out of sync, that will probably be fixed with this next time the site is updated.

PR #320, opened 2020-12-02
Probable error in example "TemporalSoftmax"

coobit:
Isn't softmax divides by a (Sum over inputs_exp) and not by a (Sum over raw inputs)


fchollet (approved):
LGTM, that's right, it was a typo.

PR #325, opened 2020-12-07
Structured data example

ksalama:


ksalama:
@fchollet - Thank you for merging the PR. 

I noticed that it is not present in the 'catalog' page: https://keras.io/examples/

Is it something that you do from your side, or do I need to do a PR for this?

PR #334, opened 2020-12-15
[TF issue #27120] Clarify usage of Keras models and Keras optimizers with TensorFlow Functions

monicadsong:
This commit clarifies documentation in https://www.tensorflow.org/guide/keras/writing_a_training_loop_from_scratch#speeding-up_your_training_step_with_tffunction
about a common user pattern of passing in Keras optimizers or Keras models as arguments into a tf.function-decorated custom training loop. Confusion regarding this pattern is documented in https://github.com/tensorflow/tensorflow/issues/27120.

### DIFF
--- a/guides/writing_a_training_loop_from_scratch.py
+++ b/guides/writing_a_training_loop_from_scratch.py
@@ -2,7 +2,7 @@
 Title: Writing a training loop from scratch
 Author: [fchollet](https://twitter.com/fchollet)
 Date created: 2019/03/01
-Last modified: 2020/04/15
+Last modified: 2020/12/15
 Description: Complete guide to writing low-level training & evaluation loops.
 """
 """
@@ -292,6 +292,12 @@ for epoch in range(epochs):
 
 """
 Much faster, isn't it?
+
+Caution: Note how `x` and `y` are the only arguments passed into `train_step`. 
+Passing in Keras optimizers or Keras models to functions that are decorated with
+`tf.function` may lead to an error. If you would like to call `train_step` with
+a new optimizer or new model, it is recommended to retrace the `train_step` or
+create a new `tf.function`-decorated training loop for that model or optimizer.
 """
 
 """
--- a/guides/writing_a_training_loop_from_scratch.py
+++ b/guides/writing_a_training_loop_from_scratch.py
@@ -295,7 +295,8 @@ Much faster, isn't it?
 
 Caution: Note how `x` and `y` are the only arguments passed into `train_step`. 
 Passing in Keras optimizers or Keras models to functions that are decorated with
-`tf.function` may lead to an error. If you would like to call `train_step` with
+`tf.function` may lead to `ValueError: tf.function-decorated function tried to
+create variables on non-first call`. If you would like to call `train_step` with
 a new optimizer or new model, it is recommended to retrace the `train_step` or
 create a new `tf.function`-decorated training loop for that model or optimizer.
 """
--- a/guides/writing_a_training_loop_from_scratch.py
+++ b/guides/writing_a_training_loop_from_scratch.py
@@ -297,9 +297,9 @@ Caution: Note how `x` and `y` are the only arguments passed into `train_step`.
 Passing in Keras optimizers or Keras models to functions that are decorated with
 `tf.function` may lead to `ValueError: tf.function-decorated function tried to
 create variables on non-first call`. If you see this error, you are most likely
-using the same traced graph with a new optimizer or new model. In this case, you
-should retrace the `train_step` or create a new `tf.function`-decorated training
-loop for that model or optimizer.
+trying to pass in a new optimizer or new model to the same graph. In this case,
+you should force retracing of the `train_step` or create a new 
+`tf.function`-decorated training loop for that model or optimizer.
 """
 
 """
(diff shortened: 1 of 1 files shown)


guides/writing_a_training_loop_from_scratch.py
```diff
@@ -292,6 +292,13 @@ def test_step(x, y):
 
 """
 Much faster, isn't it?
+
+Caution: Note how `x` and `y` are the only arguments passed into `train_step`.
```
fchollet, line 296: In general, a tf.function can take as argument a model or optimizer, as long as that model or optimizer creates its variables the first time the tf.function is called (or before the tf.function is called). The issue you see only arises if variables are created at a later call.

What code were you trying to run?

monicadsong: There is this example taken from here: https://github.com/tensorflow/tensorflow/issues/27120#issuecomment-609111883. It seems like there are users who want to use the same trace from a tf.function on different optimizers or models, and they see the error message in the title of the issue and are stuck.
```
opt1 = tf.keras.optimizers.Adam(learning_rate = 1e-2)
opt2 = tf.keras.optimizers.Adam(learning_rate = 1e-2) # could use 1e-3

@tf.function
def train_step_with_opt(a, x, y, optimizer):
    print('retracing')
        L = tf.reduce_sum(tf.square(a*x - y))

    tr_weights = [a]
    gradients = tape.gradient(L, tr_weights)
    optimizer.apply_gradients(zip(gradients, tr_weights))

    return a

a = tf.Variable(2.)
x = tf.Variable([-1.,-1.,-1.], dtype = tf.float32)
y = tf.Variable([2.,2.,2.], dtype = tf.float32)

train_step_with_opt(a, x, y, opt1) # works, but retraced twice for a single initial call
train_step_with_opt(a, x, y, opt2) # fails
```

mdanatg: Let me investigate the original issue before submitting this PR. I see Alex started looking at it, but not sure how far he got. At any rate, passing an optimizer as argument should just work, and I'd like to avoid recommending the contrary.

MarkDaoust: Hi, 

I was also surprised by this. I looked into it a bit.

And I found that confusion is coming from the fact that "non-first" is global. Not per cache-key. 

https://colab.corp.google.com/drive/1_ncNEb30ojg_9ICJCeZo4EfZ1rl211R5#scrollTo=3wT6slGK1vsQ

So in https://github.com/tensorflow/tensorflow/issues/27120 everyone is getting tripped up when they pass a *second* uninitialized layers/models/optimizer to a single tf.function definition.

It's obvious why this should fail:

```
@tf.function
def my_function(input):
  return layers.Dense(10)
```

But this failure is surprising (if you don't know about it):

```
dense1 = tf.keras.layers.Dense(3)
dense2 = tf.keras.layers.Dense(3)

@tf.function
def fun(dense, input):
  return dense(input)
  
fun(dense1, input)
fun(dense2, input)
```

A seperate @tf.function instance per object works fine (that's how @tf.function on a method works)

MarkDaoust: I had a chat with @mdanatg about this, and I think the conclusion we came to was that the main reason this is confusing is that applying an `@tf.function` behaves differently on a regular python function vs. a method on a class.

Python doesn't really see these as different things, why does tensorflow.

```
@tf.function
def fun(obj, input):
  ...

class MyClass(tf.Module):
  method=fun
```

```
class MyClass(tf.Module):
  @tf.function
  def method(self, inputs):
    ...

fun = MyClass.method
```

monicadsong: Ah yes, reminds me of this slide from https://www.youtube.com/watch?v=MSXouZPyTrc
Screen Shot 2020-12-16 at 8 43 29 AM

MarkDaoust: @monicadsong, 

So now that I've finally caught up with you in understanding the issues here, I'm starting to think that we should add links in the other direction. One of the goals in the Keras guides is always to keep things clean and simple. I think this doc should link to guide/function, and that we should describe those issues in detail there.

For the warning here I'm leaning towards

"""
Caution: Keep `tf.function` usage simple. Common sources of errors include: 1. Passing Keras objects as arguments; 2. Creating Keras objects inside the `tf.function`; or 3. Changing the contents of objects/namespaces after defining a `tf.function`. Instead: 1. Define classes and use `tf.function` on methods, or 2. Create a separate `tf.function`s per object. See the [tf.function guide](https://www.tensorflow.org/guide/function)  for details.
"""

There's no good way to get the list into the callout box.

WDYT?

mdanatg: Agree than 2 and 3 should be avoided for now. We need to think about the best workaround to recommend in general (there may not always be the same).

We really don't have a good answer for point 1 right now. I do recommend passing all the inputs (objects or not) as arguments, and that works correctly so long as they don't try to create variables. If you don't pass them as arguments, you risk other silent failures. The alternative point 1 also has caveats because tf.function doesn't monitor changes to `self`.
I would much rather prioritize fixing the root causes around point 1 than risking a potentially confusing advice.

fchollet: SGTM. I think we can also call out that these are not errors per se, but could lead to edge cases related to variable creation order...

monicadsong:
Closing this, as we've decided to make changes only to the tensorflow.org Function guide, thanks!

PR #342, opened 2020-12-24
Fix model subclassing guide link

greeeg:
This PR updates the guide slug used in the `api/models` page in order to redirect to an existing page.
Indeed, it looks like the `model_subclassing` guide does not exist anymore and causes a 404.


fchollet (approved):
LGTM, thanks

PR #347, opened 2021-01-04
Create an example for implementing the Vision Transformer (ViT) model

ksalama:
Hello @fchollet - Kindly review this PR. I created an example to demonstrate the implementation of the [Vision Transformer](https://arxiv.org/pdf/2010.11929.pdf) model on the CIFAR-100 classification dataset.


examples/vision/image_classification_with_vision_transformer.py
```diff
@@ -0,0 +1,294 @@
...
+## Introduction
+
+This example implements the [Vision Transformer (ViT)](https://arxiv.org/abs/2010.11929)
+model, by Alexey Dosovitskiy et al., for CIFAR-100 image classification.
+The ViT applies the transformer architecture with self-attentions to sequences of
+image patches, without using convolutional networks.
+
+The example requires TensorFlow 2.4 or higher, and
+[TensorFlow Addons](https://www.tensorflow.org/addons/overview),
+which can be installed using the following command:
```
8bitmp3, line 19: Suggestion:

- Not sure if the extra commas are necessary here.

```suggestion
This example implements the [Vision Transformer (ViT)](https://arxiv.org/abs/2010.11929)
model by Alexey Dosovitskiy et al. for image classification on the CIFAR-100 dataset.
The ViT applies the Transformer architecture with self-attentions on sequences of
image patches without using convolutional networks.

The example requires TensorFlow 2.4 or higher and
[TensorFlow Addons](https://www.tensorflow.org/addons/overview),
which can be installed using the following command:
```

8bitmp3: Just to double check - is this 100% your work? If you have used some open-sourced code, maybe it'd be a good idea to reference them and give the original authors credit (even if you used their, say, PyTorch code to implement your TensorFlow version).

examples/vision/image_classification_with_vision_transformer.py
```diff
@@ -0,0 +1,294 @@
...
+        self.mlp = keras.Sequential(mlp_layers)
+
+    def call(self, inputs):
+        return self.mlp(inputs)
+
+
+"""
+## Experiment 1: Train the baseline classification model
+
+We use an untrained ResNet50 architecture as our baseline model.
```
8bitmp3, line 116: Suggestion: ResNet-50
```suggestion
We use an untrained ResNet-50 architecture as our baseline model.
```

examples/vision/image_classification_with_vision_transformer.py
```diff
@@ -0,0 +1,294 @@
...
+    # Create the Keras model.
+    model = keras.Model(inputs=inputs, outputs=outputs)
+    return model
+
+
+resnet_classifier = create_resnet_classifier()
+history = run_experiment(resnet_classifier)
+
+"""
+After 100 epochs, the RestNet50 classification model achieves around 66% top 5
```
8bitmp3, line 142: Suggestion:

```suggestion
After 100 epochs the RestNet-50 classification model achieves around 66% top 5
```

Also, What is top 5 referring to here?

examples/vision/image_classification_with_vision_transformer.py
```diff
@@ -0,0 +1,294 @@
...
+accuracy on the test data.
+"""
+
+"""
+## Experiment 2: Train Vision Transformer model
```
8bitmp3, line 147: Suggestion:

```suggestion
## Experiment 2: Train the Vision Transformer model
```

examples/vision/image_classification_with_vision_transformer.py
```diff
@@ -0,0 +1,294 @@
...
+The ViT model consists of multiple layers of the Transformer block,
+which uses the `layers.MultiHeadAttention` layer as a self-attention mechanism
+for the sequence of patches. The Transformer blocks produces a
+`[batch_size, num_patches, projection_dim]` tensor, which is processed via an
+MLP head with softmax to produce the final class probabilities output.
+
+Unlike the technique described in the [paper](https://arxiv.org/abs/2010.11929),
+which prepends a learnable embedding to the sequence of encoded patches to serve
+as the image representation, the outputs of the final Transformer block are
+aggregated, using `layers.GlobalAveragePooling1D()`, and used as the image
```
8bitmp3, line 243:
```suggestion
aggregated with `layers.GlobalAveragePooling1D()` and then used as the image
```

examples/vision/image_classification_with_vision_transformer.py
```diff
@@ -0,0 +1,294 @@
...
+    model = keras.Model(inputs=inputs, outputs=outputs)
+    return model
+
+
+vit_classifier = create_vit_classifier()
+history = run_experiment(vit_classifier)
+
+"""
+After 100 epochs, the ViT classification model achieves more than 72% top 5
+accuracy on the test data. You can try to train the model for more epochs,
```
8bitmp3, line 291: Sorry, I'm not well-versed in Transformers - what does top 5 refer to, if I or some other user didn't read the original paper?

examples/vision/image_classification_with_vision_transformer.py
```diff
@@ -0,0 +1,294 @@
...
+use larger number of Transformer layer, or increase the projection dimensions
```
8bitmp3, line 292: Suggestion:

```suggestion
use larger number of Transformer layer or increase the projection dimensions
```

Also, by "use [a] larger number of... layer" do you mean "increase the number of Transformer layers" or "increase the width of the network"? I think it may be a good idea to make it clear here in case someone wants to improve on this work.

examples/vision/image_classification_with_vision_transformer.py
ksalama: Yes indeed, it is 100% my work. The original implementation of the paper's authors is in JAX. I found other implementations online, but this one is different in many ways, as I try to simplify, use existing APIs, and follow the best practices used in other examples. In addition, I made a few changes to the original implementation.

examples/vision/image_classification_with_vision_transformer.py
ksalama: Done

examples/vision/image_classification_with_vision_transformer.py
ksalama: Done

examples/vision/image_classification_with_vision_transformer.py
ksalama: Done

examples/vision/image_classification_with_vision_transformer.py
ksalama: This is the [Top K Accuracy](https://keras.io/api/metrics/accuracy_metrics/#topkcategoricalaccuracy-class) metric, in which a prediction is counted true if it is 1 of the top k classes with the highest probabilities. The standard accuracy metric is the same as Top K where K=1.

examples/vision/image_classification_with_vision_transformer.py
ksalama: Done

examples/vision/image_classification_with_vision_transformer.py
ksalama: Done

examples/vision/image_classification_with_vision_transformer.py
8bitmp3: OK, re: 

> Also, by "use [a] larger number of... layer" do you mean "increase the number of Transformer layers" or "increase the width of the network"? 

Do you think it'd help explain this bit? I'm not well-versed in Transformers, so apologies if I misunderstood this bit.

examples/vision/image_classification_with_vision_transformer.py
8bitmp3: that's amazing 👍

examples/vision/image_classification_with_vision_transformer.py
8bitmp3: I'm learning so much, thank you!

ksalama:
@8bitmp3 Thank you for your comments. I have applied your suggestions.

8bitmp3:
> After 100 epochs, the ViT classification model achieves more than 72% top 5
accuracy on the test data. You can try to train the model for more epochs,
use larger number of Transformer layer, or increase the projection dimensions

Based on the paper ("Visual Transformers: Token-based Image Representation and Processing for Computer Vision" https://arxiv.org/pdf/2006.03677v3.pdf), the authors used different ResNet architectures as backbones (-18, -50, -101, etc) for the classification task (they also experimented with segmentation and used "pyramids" but it's not the focus of your Keras example here, I think).

image

So, I think it may help new users to go into a bit more depth to clarify this:

> "use larger number of Transformer layer"

given that you used ReNet-50 as your baseline.

Anyway, it's just a thought 👍

ViT is a seriously cool paper.

examples/vision/image_classification_with_vision_transformer.py
```diff
@@ -0,0 +1,294 @@
...
+print(f"x_test shape: {x_test.shape} - y_test shape: {y_test.shape}")
+
+"""
+## Define hyperparameters
+"""
+
+learning_rate = 0.003
+weight_decay = 0.001
+batch_size = 512
+hidden_units = [128]
```
fchollet, line 55: If you want to make this a list, use at least 2 entries (two dense layers)

examples/vision/image_classification_with_vision_transformer.py
```diff
@@ -0,0 +1,294 @@
...
+num_epochs = 100
+dropout_rate = 0.5
+
+"""
+## Compile, train, and evaluate the mode
+"""
+
+
+def run_experiment(model):
+    k = 5
```
fchollet, line 65: You can just define this in the metric constructor. Changing the value requires changing the function body anyway

examples/vision/image_classification_with_vision_transformer.py
```diff
@@ -0,0 +1,294 @@
...
+    model.compile(
+        optimizer=tfa.optimizers.AdamW(
+            learning_rate=learning_rate, weight_decay=weight_decay
+        ),
+        loss=keras.losses.SparseCategoricalCrossentropy(),
+        metrics=[keras.metrics.SparseTopKCategoricalAccuracy(k)],
+    )
+
+    history = model.fit(x=x_train, y=y_train, batch_size=batch_size, epochs=num_epochs)
```
fchollet, line 74: We should use some validation data to monitor how training goes. Here you can use `validation_split`. I just noticed that this is also true for your previous two examples: we should use part of the training data to creation a validation dataset to pass as `validation_data` in `fit`. Otherwise we don't know the out-of-sample performance of the model until we run it on the test data.

examples/vision/image_classification_with_vision_transformer.py
```diff
@@ -0,0 +1,294 @@
...
+    ],
+    name="data_augmentation",
+)
+
+"""
+## Implement Multilayer perceptron (MLP) as a layer
+"""
+
+
+class MLP(layers.Layer):
```
fchollet, line 100: This is just a sequential. It would be simpler to write a function that returns a Sequential, e.g.

```
def mlp(hidden_units, dropout_rate):
    model = Sequential()
    for units in hidden_units:
          model.add(layers.Dense(units, activation=tf.nn.gelu))
          model.add(layers.Dropout(dropout_rate))
    return model
```

or alternatively (this is better):

```
def mlp(x, hidden_units, dropout_rate):
    for units in hidden_units:
          x = layers.Dense(units, activation=tf.nn.gelu)(x)
          x = layers.Dropout(dropout_rate)(x)
    return x
```

examples/vision/image_classification_with_vision_transformer.py
```diff
@@ -0,0 +1,294 @@
...
+    # Create the Keras model.
+    model = keras.Model(inputs=inputs, outputs=outputs)
+    return model
+
+
+resnet_classifier = create_resnet_classifier()
+history = run_experiment(resnet_classifier)
+
+"""
+After 100 epochs the RestNet-50 classification model achieves around 66% top 5
```
fchollet, line 142: The baseline for ResNet50 on CIFAR100 is 67% top-1 accuracy, so this result seems low. What's the top-1 accuracy?

examples/vision/image_classification_with_vision_transformer.py
```diff
@@ -0,0 +1,294 @@
...
+
+
+def run_experiment(model):
+    k = 5
+    model.compile(
+        optimizer=tfa.optimizers.AdamW(
+            learning_rate=learning_rate, weight_decay=weight_decay
+        ),
+        loss=keras.losses.SparseCategoricalCrossentropy(),
+        metrics=[keras.metrics.SparseTopKCategoricalAccuracy(k)],
```
fchollet, line 71: The standard metric for CIFAR100 is accuracy rather than top-5 accuracy. Top-5 accuracy is used on ImageNet because there are 1000 classes.

examples/vision/image_classification_with_vision_transformer.py
```diff
@@ -0,0 +1,294 @@
...
+"""
+## Introduction
+
+This example implements the [Vision Transformer (ViT)](https://arxiv.org/abs/2010.11929)
+model by Alexey Dosovitskiy et al. for image classification on the CIFAR-100 dataset.
+The ViT applies the Transformer architecture with self-attentions on sequences of
+image patches without using convolutional networks.
+
+The example requires TensorFlow 2.4 or higher and
+[TensorFlow Addons](https://www.tensorflow.org/addons/overview),
```
fchollet, line 18: Specify that TFA is used for the AdamW optimizer only

examples/vision/image_classification_with_vision_transformer.py
```diff
@@ -0,0 +1,294 @@
...
+
+"""
+## Compile, train, and evaluate the mode
+"""
+
+
+def run_experiment(model):
+    k = 5
+    model.compile(
+        optimizer=tfa.optimizers.AdamW(
```
fchollet, line 67: What do we lose by using a regular Adam optimizer instead?

examples/vision/image_classification_with_vision_transformer.py
```diff
@@ -0,0 +1,294 @@
...
+    # Create the Keras model.
+    model = keras.Model(inputs=inputs, outputs=outputs)
+    return model
+
+
+vit_classifier = create_vit_classifier()
+history = run_experiment(vit_classifier)
+
+"""
+After 100 epochs, the ViT classification model achieves more than 72% top 5
```
fchollet, line 290: Was the ViT model benchmarked on CIFAR100 by someone before? What accuracy did it achieve?

fchollet (commented):
Demonstrating the Vision Transformer model is a great idea and will make a great code example. Thank you!

examples/vision/image_classification_with_vision_transformer.py
ksalama: Done

examples/vision/image_classification_with_vision_transformer.py
ksalama: Done

examples/vision/image_classification_with_vision_transformer.py
ksalama: Done. Used validation_split = 0.15

examples/vision/image_classification_with_vision_transformer.py
ksalama: Done

examples/vision/image_classification_with_vision_transformer.py
ksalama: I actually now using just Adam, so I removed the need for tfa

examples/vision/image_classification_with_vision_transformer.py
ksalama: I tried to use a 'similar' optimization mechanism to what they used in the paper. However, I think for the purpose of demonstrating the model, we can just stick with Adam.

examples/vision/image_classification_with_vision_transformer.py
ksalama: It was benchmarked in the paper indeed. However, I didn't try to replicate their results, since they performed pre-training on the JFT-300M dataset before fine-tuning on CIFAR-100. In addition, they used much larger values for the hyperparams (e.g 12 to 32 transformer layers, compared to 4 in this example, 12-16 attention heads compared to 4 in this example, projection size 768-1280 compared to 64 in this example).

examples/vision/image_classification_with_vision_transformer.py
ksalama: Ok I will use top-1 accuracy and update the results.

ksalama:
@fchollet - I played a bit with the hyperparams, and this is the best I managed to achieve so far (100 epochs):
1. ResNet50: around 43% top 1 accuracy and 68% top 5 accuracy
2. ViT: around 45% top 1 accuracy and 75% top 5 accuracy
3. AdamW performs better than Adam (both using weight decay)
4. Using a validation split during training, I noticed that ResNet50 overfits after 20 iterations (training accuracy keeps on improving while validation accuracy stays the same), while both training and validation accuracy keeps improving in ViT. That means training ViT for longer would improve the accuracy.

I guess the main difficulty is to achieve decent results with ResNet50!
I did a bit of research and it seems that the SOTA results are achieved via careful hyperparameter tuning, training scheduling, and data augmentation. 

Now we have a couple of options to proceed with this example:
1. Submit the example as-is
2. Use a different dataset (maybe CIFAR10).
3. Extensive hyperparameter tuning.

I prefer option one as I think the purpose is to show how to implement the architecture, leaving implementing a sophisticated training routine to the user. In addition, we show that ViT outperforms the baseline (without careful tuning in both). I don't mind the second option, but I am not very interested in the third one tbh (In fact, it could be a nice example - on its own - around how to implement a sophisticated training routine to achieve the best result with ResNet50 on CIFAR-100, which doesn't seem to be easy at all!).

Let me know what do you think, or if you have other suggestions.

fchollet:
I think you'll find it easier to demonstrate good ViT results on CIFAR100 than CIFAR10.

I think data augmentation is definitely a must. Try increasing a bit the random rotation factor and maybe add `RandomZoom`. 

From what you describe you might have an overfitting problem. Try removing the MLP classifier -- a single softmax layer with dropout before it should be enough.

One thing you could do is reproduce the same learning rate schedule that was used in the best ResNet50 benchmark.

I don't think we need this example to be SotA, but it shouldn't be too low either. Given that ResNet50 is supposed to achieve 67% acc, I would recommend reaching at least 50%.

So I would recommend trying a few more configurations (no MLP, more augmentation, learning rate scheduling) and adopting the best performing one. Hopefully it will be closer to the benchmark. Then we can submit, regardless of the final results. I completely agree that it's more important to demonstrate the advantage of ViT.

ksalama:
@fchollet - Ok. I spent more time optimizing the ResNet50 model, but I couldn't get above 45% accuracy on the test set. I tried different combinations of removing MLP, learning rate, batch size, Adam, AdamW SGD with momentum, and several data augmentation features (including RandomCrop, RandomZoom, Normalization instead of scaling, etc.).

I found [example 1](https://medium.com/jovianml/classifying-cifar-100-with-resnet-5860a9c2c13f), [example 2](https://zhenye-na.github.io/2018/10/07/pytorch-resnet-cifar100.html), and [example 3](https://github.com/batuhan3526/ResNet50_on_Cifar_100_Without_Transfer_Learning/blob/master/ResNet50_on_Cifar_100_Without_Transfer_Learning.ipynb) in PyTorch using ResNet on CIFAR-100, which all achieved 60%+ accuracy (with no tricks in the learning procedures)!!. I tried the configurations they used, but with no luck! Also tried DenseNet, which produced a bit better results, but not reaching 50% accuracy.

I wonder if you have any idea of why this discrepancy? I am open for any suggestion :-)

Now I will switch to ViT to try to achieve above 50% accuracy, which I think it should be the matter if training for longer

fchollet:
> I wonder if you have any idea of why this discrepancy? I am open for any suggestion :-)

If you can confirm that the architectures are identical, then any such discrepancies are usually caused by:
- Different input preprocessing
- Different initializations (check especially the conv layers)
- Different optimizer configurations

ksalama:
So I found this [tf.keras.applications.resnet.preprocess_input](https://www.tensorflow.org/api_docs/python/tf/keras/applications/resnet/preprocess_input) function, which improved the results of the ResNet model when used: accuracy = 47% - top 5 accuracy = 73%. It says that 'color channel is zero-centered with respect to the ImageNet dataset', but I am not using ImageNet weights, so I am not sure if it makes since to use it.

fchollet:
> It says that 'color channel is zero-centered with respect to the ImageNet dataset', but I am not using ImageNet weights, so I am not sure if it makes since to use it.

You're processing natural images so it makes sense anyway -- the channel statistics are similar. It's recommended to use the `preprocess_input` function of each application regardless of whether you're using pretrained weights.

That still leaves a significant gap though. The fact that the preprocessing function seems to have a large impact might indicate a learning rate issue. Try different learning rates and decay schedules.

ksalama:
@fchollet - I spent a lot of time doing 'manual' hyperparameter tuning, including using different types of data augmentation, optimizers, and learning rate scheduling. Normalization and RandomZooming improved the results, however, other didn't make much difference. After 100 epochs only:
- ResNet50 achieved 48% accuracy and 74% top 5 accuracy
- ViT achieved 52% accuracy and 80% top 5 accuracy (more epochs would lead to further improvements).

I also added this at the end of the tutorial: 
```
"You can try to train the model  for more epochs, use larger number of Transformer layers, or increase  the projection dimensions to achieve better results. Also note that, as mentioned in  the [paper](https://arxiv.org/abs/2010.11929), the quality of the model is affected not only by the architecture choice, but also other parameters, such as training 
schedule, optimizer, weight decay, etc."
```

IMHO, I think trying to squeeze 1-2 accuracy percentages will not increase the value of the example, as the I think the purpose is to demonstrate a simple and an effective implementation of ViT, rather than a sophisticated training/hp tuning mechanism. 

I will try different other configurations (in the background, as I am working on other examples :-)) and update the example if I find better ones, but, as far as i am concerned, the example looks ready.

Let me know your thoughts. And thank you so much for your valuable suggestions!

fchollet:
Thanks for the update. Can you post the training and validation accuracy & loss graphs over the duration of training (epoch by epoch)?

ksalama:
This is the training graph for ResNet50: 
image

This is the graph for ViT: 
image

Also, I create two separate Colab for [ResNet](https://colab.research.google.com/drive/1T2I161N7xVRH8_nxrj-8xc4VugCCOhl2#scrollTo=2bOj1Aw3RdEL) and [ViT](https://colab.research.google.com/drive/1LMJQcBi2ndORSyif1aNYSvaDSyW1idnO#scrollTo=Ih3KSUQT7a3m) (that produced these graphs), in case you would like to tinker with them! Note that, in these notebooks, I used the test sets as the validation_data to show the accuracy graphs graphs.

fchollet:
Looks good, thank you for the information! It seems the ResNet50 starts overfitting quite early, so this might be a regularization / data augmentation problem. I think we can merge the example in this current state anyway, maybe with a caveat comment about this. I will do a last round of reviews today.

fchollet (commented):
So I tried a different config (thanks for sharing the notebook) and I'm getting 61% acc after 18 epochs. I'm confident if we wait for full convergence we'd reach something close to the high-performance baseline of 67%.

See notebook: https://colab.research.google.com/drive/10nYeZgeg1I0r6BAF85Kx4fVCUISAL_LM?usp=sharing

Please apply the following:

- Add names to metrics, for readability
- Drop TFAddons, use Adam with learning rate schedule
- Use a Resizing layer to upscale the images to 256x256
- In `run_experiment`, please add a `ModelCheckpoint` callback with `save_best_only=True`. After training, reload the best model before evaluation on the test data.
- Don't specify input_shape in the Application


Note: I only tried *one* config, because I didn't have time for more (training is slow). What I propose isn't optimal, but it's good enough to sustain comparisons with academic baselines.

examples/vision/image_classification_with_vision_transformer.py
```diff
@@ -0,0 +1,297 @@
...
+    return history
+
+
+"""
+## Using data augmentation
+"""
+
+data_augmentation = keras.Sequential(
+    [
+        layers.experimental.preprocessing.Normalization(),
```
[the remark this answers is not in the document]
fchollet: Tried it -- it works well.

examples/vision/image_classification_with_vision_transformer.py
```diff
@@ -0,0 +1,297 @@
...
+    ],
+    name="data_augmentation",
+)
+
+data_augmentation.layers[0].adapt(x_train)
+
+"""
+## Experiment 1: Train the baseline classification model
+
+We use an untrained ResNet-50 architecture as our baseline model.
```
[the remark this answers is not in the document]
fchollet: Didn't try it but also probably a good idea. Anyhow you can skip this, it's not essential

fchollet:
> In run_experiment, please add a ModelCheckpoint callback with save_best_only=True. After training, reload the best model before evaluation on the test data.

Actually, can you apply this change (as well as making sure that we use validation data in `fit`) to your previous examples? This should be standard.

examples/vision/image_classification_with_vision_transformer.py
```diff
@@ -0,0 +1,297 @@
...
+"""
+## Using data augmentation
+"""
+
+data_augmentation = keras.Sequential(
+    [
+        layers.experimental.preprocessing.Normalization(),
+        layers.experimental.preprocessing.RandomFlip("horizontal"),
+        layers.experimental.preprocessing.RandomRotation(0.02),
+        layers.experimental.preprocessing.RandomZoom(0.2, 0.2),
```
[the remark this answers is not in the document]
ksalama: Done

fchollet:
Update: confirmed that we get to 67%. Based on results so far I suspect that a bit of hparam optimization would get us to 70%, but that's irrelevant. Let's just use the config I use in the notebook.

ksalama:
@fchollet - Thank you so much for your input! I think the key trick was the resizing, right?

I will work on figuring out the changes required for the ViT to be comparable with these updates (as with the resizing, we need need to increase the batch size, projection dimension, the MLP head etc.). Will work one this and hoping we get better results.

ksalama:
@fchollet - I have been working on tuning the hyperparameters for ViT to improve its performance. The best I could get is 55% after 100 epochs on the original 32x32 image size with patch size of 4x4. As shown in the image, the model could benefit from training longer. 

image

I have also tried a few configurations with resizing the image to 255x255 and make the patch size 16x16 with no luck (training is slow, can't increase the batch size due to memory limitation etc.). However, I am still trying out a few other configurations.

I also checked the paper and all the results are based on pre-training on JFT-300M dataset, then fine-tuning on CIFAR-100.
I am not sure which results to expect when training ViT from scratch on CIFAR-100.

Maybe for the scope of this example we remove ResNet and show only the implementation and the results of ViT? Also mentioning that the SOTA results are obtained by pre-training on a large, high-resolution dataset then fine-tuning. 

What are your thoughts?

fchollet:
Thanks for the update. So I think we should do the following:

- Only include the ViT model in the example. This will make the example more focused.
- Call out the fact that the model from the paper was pretrained, hence why we don't achieve the same results by training from scratch.
- Call out the fact that a ResNet50V4 trained from scratch achieves 67% accuracy: hence the ViT model is clearly not ready for practical applications, and we don't recommend anyone use it in practice. We present it for demonstration only.

What do you think?

ksalama:
Thanks for your suggestions. I updated the example accordingly.
Please let me know what you think about the paragraph at the end.

Cheers

fchollet (commented):
Looks good to me, thank you! I applied some copyedits. You can add the generated files.

ksalama:
@fchollet - Thank you for the edits. I am reluctant to draw this conclusion: "So the ViT model isn't quite competitive with regular convnets on this task", since we haven't done extensive hp-tuning nor used the config used in the paper. The author won't like it if they come across this example :-) In fact, I manage to achieve better result the more I play with the parameters.

So I prefer not to conclude with this statement, not to compare with ResNet50 results (they can refer to the paper to see the results and the comparisons).

What do you think?

fchollet:
Keras.io examples are meant to be used as guides for people building practical applications. By publishing an example, we're encouraging people to take that model and apply it in practice on new problems. If a given technique is *not* a best practice, we should disclose this and provide the reader with extra context to enable to judge the technique fairly.

We should definitely compare the results here to a sane CNN baseline (which ResNet50 provides). We can definitely also state that results are likely to improve with more hyperparameter tuning, but we should *not* just skip mentioning the baseline, which could get people to believe that ViT is preferable to a basic CNN.

ksalama:
Understood - I rephrased the conclusion a little bit. Let me know if you are happy with it.
I also generated the notebook and the md files.

Thanks a lot for all your efforts in this one!

fchollet (approved):
LGTM, thank you! 👍

eri820503:
@ksalama , thanks for your example.
There is a class token in original paper and classification head, which is the last MLP layer, is attached from it.
But it seems that your example omits it.

PR #351, opened 2021-01-06
Update oxford_pets_image_segmentation.py

allihaider:
For the two `curl` commands in lines 13 and 14, using URLs with the http protocol downloads files that are not recognized by the subsequent `tar` commands as tar archives. Using the https protocol solves the issue.


fchollet (commented):
Looks good, please update the `md` and `ipynb` files as well.

allihaider:
@googlebot I signed it!

fchollet (approved):

PR #363, opened 2021-02-01
Update Save and load Keras subclassed models

giulatona:
The guide states that to save/load custom layer or a subclassed model, the get_config and optionally from_config methods should be overwritten. However, in the case of subclassed model the definition of from_config is needed and not optional. Updated to reflect the need of definition of from_config for subclassed models.

### DIFF
--- a/guides/serialization_and_saving.py
+++ b/guides/serialization_and_saving.py
@@ -372,7 +372,7 @@ which cannot be serialized into a JSON-compatible config
 but it's completely unsafe and means your model cannot be loaded on a different system.
 
 In order to save/load a model with custom-defined layers, or a subclassed model,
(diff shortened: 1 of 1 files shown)


giulatona:
@googlebot I signed

fchollet (commented):
Thanks for the PR!

It appears that the original text already make it clear that we're talking about custom layers:

"""
In order to save/load a model with **custom-defined layers**, or **a subclassed model**,
you should overwrite the `get_config` and optionally `from_config` methods.
"""

giulatona:
Thank you. I would like to point out however, that I meant that the documentation implies that also for subclassed models it is optional to define from_config. This does not work.

PR #386, opened 2021-03-01
Update working_with_rnns.py

ksood12345:
In the API, it specifies that the output of the Bidirectional RNN is the concatenation of the forward layer output and backward layer output by default.

### DIFF
--- a/guides/working_with_rnns.py
+++ b/guides/working_with_rnns.py
@@ -313,7 +313,7 @@ Under the hood, `Bidirectional` will copy the RNN layer passed in, and flip the
 `go_backwards` field of the newly copied layer, so that it will process the inputs in
 reverse order.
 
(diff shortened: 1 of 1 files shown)


ksood12345:
@googlebot I signed it!

8bitmp3:
@ksood12345 Good point, the default value of the `merge_mode` arg in `tf.keras.layers.Bidirectional()` is indeed `"concat"`.

https://keras.io/api/layers/recurrent_layers/bidirectional/:

> `merge_mode`: Mode by which outputs of the forward and backward RNNs will be combined. One of {'sum', 'mul', 'concat', 'ave', None}. If None, the outputs will not be combined, they will be returned as a list. Default value is 'concat'.

fchollet (approved):
LGTM, thanks

PR #393, opened 2021-03-06
Adding an image classification incorporating mixUp augmentation

sayakpaul:
Hi @fchollet. 

My first-ever example here. I decided to go with something simple yet extremely effective -- the `MixUp` augmentation technique proposed [here](https://arxiv.org/abs/1710.09412).


sayakpaul:
@googlebot I fixed it.

PR #396, opened 2021-03-06
Load model from the correct path

rebelgiri:
The model was saved at the path **model_checkpoints** but it was getting loaded from the path **saved_checkpoints**.
**saved_checkpoints** does not exist.


rebelgiri:
Insignificant issue.

PR #409, opened 2021-03-14
Added siamese network example using Totally looks like dataset

hazemessamm:
Added siamese network example with totally looks like dataset with custom training loop and custom generator.


hazemessamm:
@googlebot I signed it!

fchollet (commented):
Thanks for the PR! This is a very valuable topic.

Please only include the Python file in the PR for now. We'll add the other files when the code is ready.

I recommend further iterating on the text explanations of the example. I will provide a new round of codes later.

hazemessamm:
@fchollet Okay great I will remove the v1 lines (I used them because of my GPU but I found other alternatives),  I will modify everything then I will make another pull request with the new modifications.

PR #426, opened 2021-03-27
Update debugging_tips.py

JakeWilds-Vertikal:
simple typo. Not sure if this will create a PR based on where I'm logged in right now. If it doesn't, I'll try to remember to do this from a dev machine when I'm back.


JakeWilds-Vertikal:
@googlebot I signed it!

fchollet (changes requested):
Thanks for the PR! Please also fix the typo in the generated `ipynb` and `md` files.

JakeWilds-Vertikal:
> Thanks for the PR! Please also fix the typo in the generated `ipynb` and `md` files.

Got those too!

fchollet (approved):
Thanks!

PR #442, opened 2021-04-13
Example on consistency training for vision models

sayakpaul:
This example demonstrates _consistency training_ which is one of the major elements of works like FixMatch, UDA, and Noisy Student Training. It not only yields better performing models having higher topline metrics like Top-1 Accuracy but also provides enhanced robustness to common corruptions and perturbations.

Besides, this example could also be treated as a template for developing semi-supervised or weakly supervised classification methods.


sayakpaul:
Not sure what is happening with the black formatting since I did run `black consistency_training.py` before committing it.

fchollet:
> Not sure what is happening with the black formatting since I did run black consistency_training.py before committing it.

Maybe a discrepancy between `black` versions?

sayakpaul:
> Maybe a discrepancy between `black` versions?

That seems to be the case. After reformatting now when I am trying to make the final push it says everything is up-to-date:

image

8bitmp3:
This is an excellent example and thanks for bringing up the challenges of non-i.i.d. data and what this means for computer vision tasks.

I'd suggest to have the introductory section:

- Explain what the example does in the first sentence (why should users care about this awesome example, since they are busy)
- Mention what can cause the distribution shifts (such as random noise, contrast change, and blurring)
- Mention that normally we assume i.i.d. but these models can have performance degradation (this is from the "A Fourier Perspective on Model Robustness in Computer Vision" paper you mentioned)
- Trim words without altering the meaning (e.g. "conventional... standard..." in one sentence)
- Mention the name of the great paper ("A Fourier Perspective on Model Robustness in Computer Vision" vs "this literature", so that the users don't have to lick to know what it is)
- Other stuff related to syntax and style

```
This example demonstrates how to achieve robustness against data distribution
shifts in computer vision with the help of the
[RandAugment](https://arxiv.org/abs/1909.13719) data augmentation method.

Deep learning models excel in many image recognition tasks when the data is
independent and identically distributed (i.i.d.). However, they can suffer from
performance degradation caused by subtle distribution shifts in the input data 
(such as random noise, contrast change, and blurring), as discussed in
[A Fourier Perspective on Model Robustness in Computer Vision](https://arxiv.org/pdf/1906.08988.pdf)).
In addition, model training procedures, such as standard image classification
training workflows) don't typically enable a model to generalize well.

In this example, we will be training an image classification model enforcing a
sense of *consistency* inside it through the following steps:

- Train a standard image classification model.
- Train an equal or larger model on a noisy version of the dataset augmented
using [RandAugment](https://arxiv.org/abs/1909.13719). 
  - To do this, we begin with obtaining predictions of the previous model on the
  clean images of the dataset. 
  - Then, we use these predictions and train the second model to match these
predictions on the noisy variant of the same images. This is identical to the
workflow in the [*Knowledge Distillation*](https://keras.io/examples/vision/knowledge_distillation/)
example; but, because the student model is equal or larger in size, this process
is also referred to as *self-training*. 
```

8bitmp3:
Regarding the title, because "consistency" may mean different things to different users in different domains, I'd suggest the following to be more precise and capture the right audience (The current title is very vague IMHO)

```suggestion
Title: Robustness against data distribution shifts in supervised learning in computer vision
```

"Robustness against data distribution shifts" is how the authors of "a Fourier Perspective on Model Robustness in Computer Vision" paper would probably describe it.

8bitmp3:
Alternatively, you could alter the description to  with "supervised learning" -> "supervised learning in computer vision" (for precision)

```
Description: Robustness against data distribution shifts in supervised learning in computer vision
```

While leaving the title shorter, such as:

```
Title: Robustness against data distribution shifts
```

sayakpaul:
I would keep the description as you suggested but for the title, I would prefer the current one. Here's why:

Although the example focuses on using consistency training to enhance the robustness of models to common corruptions this example also serves as a template for weakly supervised learning and other forms of consistency regularization. Maybe mentioning this in the text would be more justifiable.

examples/vision/consistency_training.py
```diff
@@ -0,0 +1,391 @@
...
+ensembling* and when the student model is forced to mimic that, it leads to better
+performance. The original work uses [Stochastic Depth](https://arxiv.org/abs/1603.09382)
+and [Dropout](https://jmlr.org/papers/v15/srivastava14a.html) to bring in the ensembling
+part but for this example, we will use [Stochastic Weight Averaging](https://arxiv.org/abs/1803.05407)
+(SWA) which also resembles geometric ensembling.
+"""
+
+# Define the callbacks.
+reduce_lr = tf.keras.callbacks.ReduceLROnPlateau(patience=3)
+es = tf.keras.callbacks.EarlyStopping(patience=10, restore_best_weights=True)
```
fchollet, line 220: Use fully spelled out variable names

examples/vision/consistency_training.py
```diff
@@ -0,0 +1,391 @@
...
+
+# Initialize SWA from tf-hub.
+SWA = tfa.optimizers.SWA
```
fchollet, line 223: How are you using this?

examples/vision/consistency_training.py
```diff
@@ -0,0 +1,391 @@
...
+
+        # Return a dict of performance
+        results = {m.name: m.result() for m in self.metrics}
+        return results
+
+
+"""
+The only difference in this implementation is the way loss is being calculated. Instead
+of weighted the distillation loss and student loss differently we are taking their
+average following Noisy Student Training.
```
fchollet, line 329: Here, it would be worth highlighting the specific line in question

fchollet (commented):
Thanks for the PR! This is a great example.

examples/vision/consistency_training.py
```diff
@@ -0,0 +1,387 @@
...
+
+"""shell
+pip install -q tf-models-official tensorflow-addons
+"""
+
+"""
+## Imports and setup
+"""
+
+from official.vision.image_classification.augment import RandAugment
```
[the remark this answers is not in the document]
sayakpaul: Yes. It's been some time. I got to know about it via the following tutorials:

* https://www.tensorflow.org/tutorials/text/classify_text_with_bert
* https://www.tensorflow.org/tutorials/text/solve_glue_tasks_using_bert_on_tpu

examples/vision/consistency_training.py
sayakpaul: Yes. I had forgotten to use it to wrap the optimizer. It's been fixed now and I have also added a comment above its usage so that it won't go unnoticed.

examples/vision/consistency_training.py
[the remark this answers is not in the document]
fchollet: I've escalated this, hopefully it will get resolved.

sayakpaul:
@fchollet would suggest opening a separate PR since the Black CI is failing?

fchollet:
> @fchollet would suggest opening a separate PR since the Black CI is failing?

We can merge first and fix it afterwards. We should also look into making the failure message more informative, as a follow up.

fchollet (approved):
Looking great -- thanks for the update! We can add the generated files now, then merge, then manually resolve the style issue.

sayakpaul:
Thanks, @fchollet. Should be all set now.

fchollet (approved):
Excellent -- thank you for the great contribution!

PR #451, opened 2021-04-24
Example on semi-supervised learning with SimCLR

beresandras:
This example demonstrates the usage of contrastive pretraining combined with supervised finetuning on the STL10 dataset. The performance of the pretrained + finetuned model exceeds that of its supervised baseline counterpart.

Other features include:
-custom image augmentation layers
-sampling differently-sized batches from the zipped labeled and unlabeled datasets simultaneously

A current issue is that the example exceeds the allowed line-count (386 vs 300). The implementation could be made somewhat shorter by removing training curve visualization and print() and model.summary() statements, they are however helpful for comparison with the baseline solution.


beresandras:
@googlebot I signed it!

examples/vision/semisupervised_simclr.py
```diff
@@ -0,0 +1,743 @@
...
+# load STL10 dataset
+batch_size, train_dataset, labeled_train_dataset, test_dataset = prepare_dataset(
+    steps_per_epoch
+)
+
+"""
+## Custom image augmentation layers
+
+The two most important image augmentations for contrastive learning are the following:
+- cropping: forces the model to encode different parts of the same image similarly
```
fchollet, line 127: Insert line break above start of list so that it renders correctly

examples/vision/semisupervised_simclr.py
```diff
@@ -0,0 +1,743 @@
...
+- color jitter: prevents a trivial color histogram-based solution to the task by
+distorting color histograms
+
+In this example we use random horizontal flips as well.
+
+We implement our image augmentations as custom preprocessing layers. This has the
+following two advantages:
+- the data augmentation will run on the GPU, so the training will not be bottlenecked by
```
fchollet, line 135: Here too

examples/vision/semisupervised_simclr.py
```diff
@@ -0,0 +1,743 @@
...
+# https://github.com/pytorch/vision/blob/master/torchvision/transforms/functional_tensor.py
+
+# however these augmentations:
+# -run on batches of images
+# -run on gpu
+# -can be part of a model
+
+
+# crops and resizes part of the image to the original resolutions
+class RandomResizedCrop(layers.Layer):
```
fchollet, line 153: You can remove this layer and instead use a combination of the layers RandomCrop and Resizing. Saves you a good number of lines

examples/vision/semisupervised_simclr.py
```diff
@@ -0,0 +1,743 @@
...
+        super().__init__(**kwargs)
+
+        # color jitter ranges: (min jitter strength, max jitter strength)
+        self.brightness = brightness
+        self.contrast = contrast
+        self.saturation = saturation
+        self.hue = hue
+
+        # list of applicable color augmentations
+        self.color_augmentations = [
```
fchollet, line 212: All this really does is apply a random affine transform. You can implement a random affine transform in ~15 lines top, which will save you quite a few lines.

examples/vision/semisupervised_simclr.py
```diff
@@ -0,0 +1,743 @@
...
+        # same image should be higher than their similarity with other views
+        contrastive_labels = tf.range(batch_size)
+        self.contrastive_accuracy.update_state(contrastive_labels, similarities)
+        self.contrastive_accuracy.update_state(
+            contrastive_labels, tf.transpose(similarities)
+        )
+
+        # the temperature-scaled similarities are used as logits for cross-entropy
+        # a symmetrized version of the loss is used here
+        loss = (
```
fchollet, line 495: Prefer splitting this into 3-4 lines, it will be more readable

examples/vision/semisupervised_simclr.py
```diff
@@ -0,0 +1,743 @@
...
+
+Contrastive learning avoids this condition by modifying the objective in the following
+way: it pulls representations of augmented versions/views of the same image closer to
+each other (contracting positives), while simultaneously pushing different images away
+from each other (contrasting negatives) in representation space.
+
+One such contrastive approach is [SimCLR](https://arxiv.org/abs/2002.05709), which
+essentially identifies the core components needed to optimize this objective, and
+achieves a high performance by scaling this simple approach. For further reading about
+SimCLR check out [its
```
fchollet, line 45: All links should be on a single line, otherwise they won't render properly

examples/vision/semisupervised_simclr.py
```diff
@@ -0,0 +1,743 @@
...
+Title: Semi-supervised image classification using contrastive pretraining with SimCLR
+Author: [András Béres](www.linkedin.com/in/andras-beres-789190210)
+Date created: 2021/04/24
+Last modified: 2021/04/24
+Description: Using SimCLR for contrastive pretraining for semi-supervised image classification.
+"""
+"""
+## Introduction
+
+### Semi-supervised learning
```
fchollet, line 11: In this section it would be useful to reference this example https://keras.io/examples/vision/simsiam/ and explain what is different & what is similar here

fchollet (commented):
Thanks for the PR! This looks really nice!

> A current issue is that the example exceeds the allowed line-count (386 vs 300)

You can definitely fit this in 300 lines. You can save a lot of lines by simplifying the data augmentation code. And then maybe a few in the plotting code.

8bitmp3 (commented):
@beresandras This is a very cool example. Self-sup learning is 🔥 . 

I'm just reading through the original SimCLR research, can review the rest of the example. Below is a quick suggestion.

examples/vision/semisupervised_simclr.py
```diff
@@ -0,0 +1,743 @@
...
+            layers.Input(shape=(96, 96, 3)),
+            preprocessing.Rescaling(1 / 255),
+            preprocessing.RandomFlip("horizontal"),
+            RandomResizedCrop(scale=(0.5, 1.0), ratio=(3 / 4, 4 / 3)),
+            RandomColorJitter(brightness=0.2, contrast=0.2, saturation=0.2, hue=0.1),
+        ],
+        name="classification_augmenter",
+    )
+
+
```
ebursztein, line 313: I would recommend refactoring get_classification_augmenter() and get_contrastive_augmenter() as the only difference is the ResizeCrop and ColorJitter color. That's a lot duplicated code. 

I would also try to not have the parameters hardcoded so maybe you can have a dictionary object 
`classification_augmentation = {
"shape": (96, 96, 3),
....
"resizedcrop_scale": (0.5, 1.0)
...
}`

examples/vision/semisupervised_simclr.py
```diff
@@ -0,0 +1,743 @@
...
+- we treat each image in the batch as if it had its own class
+- then we have two examples (a pair of augmented views) for each "class"
+- each view's representation is compared to every possible pair's one (for both augmented
+versions)
+- we use the temperature-scaled cosine similarity of compared representations as logits
+- and use categorical cross-entropy as the "classification" loss
+
+The following two metrics are used for monitoring the pretraining performance:
+- **contrastive accuracy**: self-supervised metric, the ratio of cases in which the
+closest representation to a view was its differently augmented version from the same
```
ebursztein, line 412: was different? Unsure what the sentence means.

examples/vision/semisupervised_simclr.py
```diff
@@ -0,0 +1,743 @@
...
+image. Self-supervised metrics can be used for hyperparameter tuning even in the case
+when there are no labeled examples.
+- **linear probing accuracy**: supervised metric, "linear probing" is a widely used
+metric in the self-supervised literature. It simply means the accuracy of a logistic
```
ebursztein, line 416: Please linkify litterature.

ebursztein, line 416: Suggested wording that might be a little clearer:
"Linear probing is a popular metric to evaluate self-supervised classifiers. It is computed as the accuracy of a regression classifier trained on top of the  encoder's features. In our case, this is done by having a single dense layer on top of the frozen encoder. Note that contrary to traditional approach where the classifier is trained after the pre-training phase, in this example,. ....

examples/vision/semisupervised_simclr.py
```diff
@@ -0,0 +1,743 @@
...
+The InfoNCE/NT-Xent/N-pairs loss is used, which can be interpreted in the following way:
+- we treat each image in the batch as if it had its own class
+- then we have two examples (a pair of augmented views) for each "class"
+- each view's representation is compared to every possible pair's one (for both augmented
+versions)
+- we use the temperature-scaled cosine similarity of compared representations as logits
+- and use categorical cross-entropy as the "classification" loss
+
+The following two metrics are used for monitoring the pretraining performance:
+- **contrastive accuracy**: self-supervised metric, the ratio of cases in which the
```
ebursztein, line 411: Add link to reference.

examples/vision/semisupervised_simclr.py
```diff
@@ -0,0 +1,743 @@
...
+only a small amount of labeled examples.
+"""
+
+"""
+## Improving further
+
+### Architecture:
+
+It is shown in the original paper, that increasing the width and depth of the models
+improves performance with a higher rate then for supervised learning. Also, using a
```
ebursztein, line 675: s/then/than/

examples/vision/semisupervised_simclr.py
```diff
@@ -0,0 +1,743 @@
...
+[ResNet50](https://keras.io/api/applications/resnet/#resnet50-function) encoder is quite
+standard in the literature. However keep in mind, that more powerful models will require
+more memory and will limit the maximal batch size you can use.
```
ebursztein, line 678: not only limit the batch size but increase training time.

examples/vision/semisupervised_simclr.py
```diff
@@ -0,0 +1,743 @@
...
+[v3](https://arxiv.org/abs/2104.02057)): uses a momentum-encoder as well, whose weights
+are an exponential moving average of the target encoder
+- [SwAV](https://arxiv.org/abs/2006.09882): uses clustering instead of pairwise comparison
+- [BarlowTwins](https://arxiv.org/abs/2103.03230): uses a cross correlation-based
+objective instead of pairwise comparison
+
+I have implemented **MoCo** and **BarlowTwins** as well in a similar fashion, you can
+find these three implementations together in [this
+repository](https://github.com/beresandras/contrastive-classification-keras). A Colab
+Notebook is included as well.
```
ebursztein, line 730: I would suggested simply saying: Keras implementation  **MoCo** and **BarlowTwins** implementation can be found here. Instead of making a special case for your own workl.

examples/vision/semisupervised_simclr.py
```diff
@@ -0,0 +1,743 @@
...
+
+There is also a new line of works, which optimize a similar objective, however without
+the use of any negatives:
+- [BYOL](https://arxiv.org/abs/2006.07733): momentum-encoder + no negatives
+- [SimSiam](https://arxiv.org/abs/2011.10566): no momentum-encoder + no negatives
```
ebursztein, line 735: Add link to the keras.io example.

examples/vision/semisupervised_simclr.py
```diff
@@ -0,0 +1,743 @@
+"""
+Title: Semi-supervised image classification using contrastive pretraining with SimCLR
+Author: [András Béres](www.linkedin.com/in/andras-beres-789190210)
+Date created: 2021/04/24
+Last modified: 2021/04/24
+Description: Using SimCLR for contrastive pretraining for semi-supervised image classification.
```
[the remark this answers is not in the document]
beresandras: That's a good point, agreed!

examples/vision/semisupervised_simclr.py
beresandras: Actually, finding a good reference for contrastive accuracy is hard, as it is not a widely used metric currently. It is reported in the SimCLR paper (Table 5) but not referenced, I can link to that. I found it useful for the reasons mentioned in the text, and therefore implemented it.

examples/vision/semisupervised_simclr.py
beresandras: Yeah... I reworded it to the following: "the ratio of cases in which the representation of an image is more similar to its differently augmented version's one, than to the representation of any other image in the current batch".

examples/vision/semisupervised_simclr.py
beresandras: Good point, I refactored them to a single function, with customizable augmentation parameters. I would however prefer to keep the input sizes hardcoded, for simplicity.

beresandras:
Thanks for your comments! I tweaked the text based on your suggestions, and also overhauled the data augmentation pipeline. Using RandomTranslation + RandomZoom for cropping now, RandomColorJitter has been replaced with a random affine transformation in color space. I might have to tune the parameters of the data augmentation pipeline a bit further.

examples/vision/semisupervised_simclr.py
```diff
@@ -0,0 +1,631 @@
...
+gather a large dataset to make it work well. However, while the cost of labeling scales
+linearly with the dataset size (labeling each example takes a constant time), model
+performance only scales [sublinearly](https://arxiv.org/abs/2001.08361) with it. This
+means that labeling more and more samples becomes less and less cost-efficient, while
+gathering unlabeled data is generally cheap as it is usually readily available in large
+quantities.
+
+Semi-supervised learning offers to solve this problem by only requiring a partially
+labeled dataset, and by being label-efficient by utilizing the unlabeled examples for
+learning as well.
```
8bitmp3, line 24: I'm just reviewing the text and noticed something - are you wrapping at a certain character level? 

E.g. At `80` this paragraph would look as follows:

```suggestion
Semi-supervised learning offers to solve this problem by only requiring a
partially labeled dataset, and by being label-efficient by utilizing the
unlabeled examples for learning as well.
```

(I'm reviewing the syntax and grammar and, for example, this sentence has "...by ... by ... by" structure - I will send your the suggestions soon.)

examples/vision/semisupervised_simclr.py
ebursztein: hardcoded inputs makes the code harder to reuse for different sizes. You would have to go and manually replace variables through the code. If you like to make it easier to read you could use:
`WIDTH = 256` or something like this at the top of the example so its clear those are constants and you see where they are used.

examples/vision/semisupervised_simclr.py
```diff
@@ -0,0 +1,631 @@
...
+from tensorflow import keras
+from tensorflow.keras import layers
+from tensorflow.keras.layers.experimental import preprocessing
+
+"""
+## Hyperparameterers
+"""
+
+num_epochs = 60
+steps_per_epoch = 200  # defines the batch size implicitly
```
fchollet, line 75: Why do this rather than use an actual batch size?

examples/vision/semisupervised_simclr.py
```diff
@@ -0,0 +1,631 @@
...
+            images = tf.clip_by_value(tf.matmul(images, color_transforms), 0, 1)
+        return images
+
+
+# image augmentation module
+def get_augmenter(min_area, brightness, color_jitter):
+    zoom_factor = 1.0 - tf.sqrt(min_area)
+    return keras.Sequential(
+        [
+            layers.Input(shape=(96, 96, 3)),
```
fchollet, line 193: You might want to make that 96 a global variable `image_size`

examples/vision/semisupervised_simclr.py
```diff
@@ -0,0 +1,631 @@
...
+        self.contrastive_optimizer = contrastive_optimizer
+        self.probe_optimizer = probe_optimizer
+
+        # self.contrastive_loss will be defined as a method
+        self.probe_loss = keras.losses.SparseCategoricalCrossentropy(from_logits=True)
+
+        self.contrastive_accuracy = keras.metrics.SparseCategoricalAccuracy()
+        self.probe_accuracy = keras.metrics.SparseCategoricalAccuracy()
+
+    def reset_metrics(self):
```
fchollet, line 362: Rather, just implement a `metrics` `@property` that lists these 2 metrics

examples/vision/semisupervised_simclr.py
```diff
@@ -0,0 +1,631 @@
...
+            class_logits = self.linear_probe(features)
+            probe_loss = self.probe_loss(labels, class_logits)
+        gradients = tape.gradient(probe_loss, self.linear_probe.trainable_weights)
+        self.probe_optimizer.apply_gradients(
+            zip(gradients, self.linear_probe.trainable_weights)
+        )
+        self.probe_accuracy.update_state(labels, class_logits)
+
+        return {
+            "c_loss": contrastive_loss,
```
fchollet, line 434: If you return a batch-wise scalar here, you will only print the value for the last batch (rather than the average of the value over an epoch, which is standard). Prefer creating 2 loss tracker `keras.metrics.Mean` objects (e.g. in `compile`) and list them in `self.metrics`. Then this line can be simplified as `return {m.name: m.result() for m in self.metrics}`.

See also https://keras.io/guides/customizing_what_happens_in_fit/#going-lowerlevel

fchollet (commented):
Thanks for the update! The code looks very nice -- love it!

Just a few comments. I'll do a last pass for the text copy afterwards.

examples/vision/semisupervised_simclr.py
ebursztein: Then the SimCLR (v1?) paper is fine as a reference. if its hard to find, then its even more useful to link where you found it.

beresandras:
In this round: I added dataset-related hyperparameters (such as image size) as global variables, replaced steps_per_epoch with batch_size, refactored the metrics so that the losses are tracked by a mean metric tracker, and refactored the visualization code to save some lines. I also wrapped all markdown text to 80 chars.

examples/vision/semisupervised_simclr.py
beresandras: Yeah, they weren't wrapped properly after I edited the text manually, its fixed now.

8bitmp3:
@beresandras Bear with me, I'll get back to asap 👍

examples/vision/semisupervised_simclr.py
```diff
@@ -0,0 +1,658 @@
...
+Date created: 2021/04/24
+Last modified: 2021/04/24
+Description: Contrastive pretraining with SimCLR for semi-supervised image classification on the STL-10 dataset.
+"""
+"""
+## Introduction
+
+### Semi-supervised learning
+
+Semi-supervised learning is a machine learning paradigm, that deals with
```
8bitmp3, line 13: Suggestion:

```suggestion
Semi-supervised learning is a machine learning paradigm that deals with
```

examples/vision/semisupervised_simclr.py
```diff
@@ -0,0 +1,658 @@
...
+**partially labeled datasets**. When applying deep learning in the real world,
+one usually has to gather a large dataset to make it work well. However, while
+the cost of labeling scales linearly with the dataset size (labeling each
+example takes a constant time), model performance only scales
+[sublinearly](https://arxiv.org/abs/2001.08361) with it. This means that
+labeling more and more samples becomes less and less cost-efficient, while
+gathering unlabeled data is generally cheap as it is usually readily available
```
8bitmp3, line 20: Suggestion:

```suggestion
gathering unlabeled data is generally cheap, as it is usually readily available
```

examples/vision/semisupervised_simclr.py
```diff
@@ -0,0 +1,658 @@
...
+in large quantities.
+
+Semi-supervised learning offers to solve this problem by only requiring a
+partially labeled dataset, and by being label-efficient by utilizing the
+unlabeled examples for learning as well.
+
+In this example, we will pretrain an encoder with contrastive learning on the
+[STL-10](https://ai.stanford.edu/~acoates/stl10/) semi-supervised dataset using
+no labels at all, and then finetune it using only its labeled subset.
```
8bitmp3, line 29:
```suggestion
no labels at all, and then fine-tune it using only its labeled subset.
```

(check if US English is OK with "labeled" (one "l"))

examples/vision/semisupervised_simclr.py
```diff
@@ -0,0 +1,658 @@
...
+
+Contrastive learning avoids this condition by modifying the objective in the
+following way: it pulls representations of augmented versions/views of the same
+image closer to each other (contracting positives), while simultaneously pushing
+different images away from each other (contrasting negatives) in representation
+space.
+
+One such contrastive approach is [SimCLR](https://arxiv.org/abs/2002.05709),
+which essentially identifies the core components needed to optimize this
+objective, and achieves a high performance by scaling this simple approach.
```
8bitmp3, line 47: Suggestion:

```suggestion
objective, and can achieve high performance by scaling this simple approach.
```

examples/vision/semisupervised_simclr.py
```diff
@@ -0,0 +1,658 @@
...
+
+Another approach is [SimSiam](https://arxiv.org/abs/2011.10566)
+([Keras example](https://keras.io/examples/vision/simsiam/)),
+whose main difference from
+SimCLR is that it does not use any negatives in its loss. Therefore it does not
```
8bitmp3, line 52: Suggestion:

```suggestion
SimCLR is that the former does not use any negatives in its loss. Therefore, it does not
```

examples/vision/semisupervised_simclr.py
```diff
@@ -0,0 +1,658 @@
...
+explicitly prevent the trivial solution, rather avoids it implicitly by
```
8bitmp3, line 53: Suggestion:

```suggestion
explicitly prevent the trivial solution, and, instead, avoids it implicitly by
```

examples/vision/semisupervised_simclr.py
```diff
@@ -0,0 +1,658 @@
...
+architecture design (assymetric encoding paths using a predictor network, and
```
8bitmp3, line 54: Suggestion + typo fix

```suggestion
architecture design (asymmetric encoding paths using a predictor network and
```

examples/vision/semisupervised_simclr.py
```diff
@@ -0,0 +1,658 @@
...
+BatchNorm applied in the final layers).
```
8bitmp3, line 55: Suggestion:

```suggestion
batch normalization (BatchNorm) are applied in the final layers).
```

examples/vision/semisupervised_simclr.py
```diff
@@ -0,0 +1,658 @@
...
+
+For further reading about SimCLR check out
```
8bitmp3, line 57: Suggestion:


```suggestion
For further reading about SimCLR check out
```

examples/vision/semisupervised_simclr.py
```diff
@@ -0,0 +1,658 @@
...
+[its blogpost](https://ai.googleblog.com/2020/04/advancing-self-supervised-and-semi.html),
```
8bitmp3, line 58: Suggestion:

```suggestion
[the official Google AI blog post](https://ai.googleblog.com/2020/04/advancing-self-supervised-and-semi.html),
```

examples/vision/semisupervised_simclr.py
```diff
@@ -0,0 +1,658 @@
...
+for an overview of self-supervised learning across both vision and language
```
8bitmp3, line 59: Suggestion:

```suggestion
and for an overview of self-supervised learning across both vision and language
```

examples/vision/semisupervised_simclr.py
```diff
@@ -0,0 +1,658 @@
...
+check out
+[this blogpost](https://ai.facebook.com/blog/self-supervised-learning-the-dark-matter-of-intelligence/).
```
8bitmp3, line 61: Suggestion:

```suggestion
[this blog post](https://ai.facebook.com/blog/self-supervised-learning-the-dark-matter-of-intelligence/).
```

examples/vision/semisupervised_simclr.py
```diff
@@ -0,0 +1,658 @@
...
+)
+
+"""
+## Self-supervised model for contrastive pretraining
+
+We pretrain an encoder on unlabeled images with a contrastive loss for 30
+epochs. A nonlinear projection head is attached to the top of the encoder, as it
+improves the quality of representations of the encoder.
+
+The InfoNCE/NT-Xent/N-pairs loss is used, which can be interpreted in the
```
8bitmp3, line 303: Suggestion: use active voice

```suggestion
We use the InfoNCE/NT-Xent/N-pairs loss, which can be interpreted in the
```

examples/vision/semisupervised_simclr.py
```diff
@@ -0,0 +1,658 @@
...
+following way:
+
+- we treat each image in the batch as if it had its own class
+- then we have two examples (a pair of augmented views) for each "class"
+- each view's representation is compared to every possible pair's one (for both
+  augmented versions)
+- we use the temperature-scaled cosine similarity of compared representations as
+  logits
+- and use categorical cross-entropy as the "classification" loss
```
8bitmp3, line 312: Suggestion:

```suggestion
1. We treat each image in the batch as if it had its own class.
2. Then, we have two examples (a pair of augmented views) for each "class".
3. Each view's representation is compared to every possible pair's one (for both
  augmented versions).
4. We use the temperature-scaled cosine similarity of compared representations as
  logits.
5. Finally, we use categorical cross-entropy as the "classification" loss
```

examples/vision/semisupervised_simclr.py
```diff
@@ -0,0 +1,658 @@
...
+
+The following two metrics are used for monitoring the pretraining performance:
+
+- [contrastive accuracy (SimCLR Table 5)](https://arxiv.org/abs/2002.05709):
```
8bitmp3, line 316: Suggestion:

```suggestion
- [Contrastive accuracy (SimCLR Table 5)](https://arxiv.org/abs/2002.05709):
```

examples/vision/semisupervised_simclr.py
```diff
@@ -0,0 +1,658 @@
...
+  Self-supervised metric, the ratio of cases in which the representation of an
+  image is more similar to its differently augmented version's one, than to the
+  representation of any other image in the current batch. Self-supervised
+  metrics can be used for hyperparameter tuning even in the case when there are
+  no labeled examples.
+- [linear probing accuracy](https://arxiv.org/abs/1603.08511): Linear probing is
```
8bitmp3, line 322: Suggestion:

```suggestion
- [Linear probing accuracy](https://arxiv.org/abs/1603.08511): Linear probing is
```

examples/vision/semisupervised_simclr.py
```diff
@@ -0,0 +1,658 @@
...
+        plt.legend()
+        plt.title(f"Classification {metric_name} during training")
+        plt.xlabel("epochs")
+        plt.ylabel(f"validation {metric_name}")
+
+
+plot_training_curves(pretraining_history, finetuning_history, baseline_history)
+
+"""
+By comparing the training curves, we can see that by using contrastive
```
8bitmp3, line 562: Suggestion:

- Avoid "By.... by..." with "By.... when..." as follows:

```suggestion
By comparing the training curves, we can see that when using contrastive
```

examples/vision/semisupervised_simclr.py
```diff
@@ -0,0 +1,658 @@
...
+validation loss, which means that the pretrained network was able to generalize
+better when seeing only a small amount of labeled examples.
+"""
+
+"""
+## Improving further
+
+### Architecture:
+
+It is shown in the original paper, that increasing the width and depth of the
```
8bitmp3, line 573: Suggestion:

```suggestion
The experiment in the original paper demonstrated that increasing the width and depth of the
```

examples/vision/semisupervised_simclr.py
```diff
@@ -0,0 +1,658 @@
...
+models improves performance at a higher rate than for supervised learning. Also,
+using a [ResNet50](https://keras.io/api/applications/resnet/#resnet50-function)
```
8bitmp3, line 575: Suggestion:

```suggestion
using a [ResNet-50](https://keras.io/api/applications/resnet/#resnet50-function)
```

examples/vision/semisupervised_simclr.py
```diff
@@ -0,0 +1,658 @@
...
+It has [been](https://arxiv.org/abs/1905.09272)
+[reported](https://arxiv.org/abs/1911.05722) that the usage of BatchNorm layers
+could sometimes degrade performance, as it introduces an intra-batch dependency
+between samples, which is why I did not have used them in this example. In my
+experiments however, using BatchNorm, especially in the projection head,
+improves performance.
+
+### Hyperparameters:
+
+The hyperparameters of this example have been tuned manually for this task and
```
8bitmp3, line 589: Suggestion:

```suggestion
The hyperparameters used in this example have been tuned manually for this task and
```

examples/vision/semisupervised_simclr.py
```diff
@@ -0,0 +1,658 @@
...
+architecture, so without changing those, only marginal gains can be expected
```
8bitmp3, line 590: Suggestion:

```suggestion
architecture. Therefore, without changing them, only marginal gains can be expected
```

examples/vision/semisupervised_simclr.py
```diff
@@ -0,0 +1,658 @@
...
+from further hyperparameter tuning.
+
+However for a different task or model architecture these would need tuning, so
+here are my notes on the most important ones:
+
+- **batch size**: since the objective can be interpreted as a classification
+  over a batch of images (loosely speaking), batch size is actually a more
+  important hyperparameter than usual. The higher, the better.
+- **temperature**: the temperature defines "softness" of the of the softmax
```
8bitmp3, line 599: Suggestion: ```suggestion
- **temperature**: the temperature defines the "softness" of the of the softmax
```

File: examples/vision/semisupervised_simclr.py [hunk 44]
```diff
@@ -0,0 +1,658 @@
...
+
+The hyperparameters of this example have been tuned manually for this task and
+architecture, so without changing those, only marginal gains can be expected
+from further hyperparameter tuning.
+
+However for a different task or model architecture these would need tuning, so
+here are my notes on the most important ones:
+
+- **batch size**: since the objective can be interpreted as a classification
+  over a batch of images (loosely speaking), batch size is actually a more
```
8bitmp3, line 597: Suggestion:

```suggestion
  over a batch of images (loosely speaking), the batch size is actually a more
```

examples/vision/semisupervised_simclr.py
```diff
@@ -0,0 +1,658 @@
...
+  important hyperparameter than usual. The higher, the better.
+- **temperature**: the temperature defines "softness" of the of the softmax
+  distribution that is used in the cross-entropy loss, and is an important
+  hyperparameter. Lower values generally lead to a higher contrastive accuracy.
+  A recent trick (in [ALIGN](https://arxiv.org/pdf/2102.05918.pdf)) is to learn
+  the temperature's value as well (which can be done by defining it as a
+  tf.Variable, and applying gradients on it). This provides a good baseline
+  value, however in my experiments the learned temperature was somewhat lower
+  than optimal, as it is optimized w.r.t. the contrastive loss, which is not a
+  perfect proxy for representation quality.
```
8bitmp3, line 607: Suggestion:

```suggestion
  tf.Variable, and applying gradients on it). Even though this provides a good baseline
  value, in my experiments the learned temperature was somewhat lower
  than optimal, as it is optimized with respect to the contrastive loss, which is not a
  perfect proxy for representation quality.
```

examples/vision/semisupervised_simclr.py
```diff
@@ -0,0 +1,658 @@
...
+- **image augmentation strength**: during pretraining stronger augmentations
+  increase the difficulty of the task, however after a point too strong
+  augmentations will degrade performance. During finetuning stronger
+  augmentations reduce overfitting while in my experience too strong
+  augmentations decrease the performance gains from pretraining. The whole data
+  augmentation pipeline can be seen as an important hyperparameter of the
+  algorithm, implementations of other custom image augmentation layers in Keras
+  can be found in
+  [this repository](https://github.com/beresandras/contrastive-classification-keras).
+- **learning rate schedule**: a constant schedule is used here, however it is
```
8bitmp3, line 617: Suggestion:

```suggestion
- **learning rate schedule**: a constant schedule is used here, but it is
```

examples/vision/semisupervised_simclr.py
```diff
@@ -0,0 +1,658 @@
...
+  [v3](https://arxiv.org/abs/2104.02057)): uses a momentum-encoder as well,
+  whose weights are an exponential moving average of the target encoder
+- [SwAV](https://arxiv.org/abs/2006.09882): uses clustering instead of pairwise
+  comparison
+- [BarlowTwins](https://arxiv.org/abs/2103.03230): uses a cross
+  correlation-based objective instead of pairwise comparison
+
+Keras implementations of **MoCo** and **BarlowTwins** can be found in [this
+repository](https://github.com/beresandras/contrastive-classification-keras),
+with a Colab Notebook included as well.
```
8bitmp3, line 642: Suggestion:

```suggestion
which includes a Colab notebook.
```

examples/vision/semisupervised_simclr.py
```diff
@@ -0,0 +1,658 @@
...
+
+There is also a new line of works, which optimize a similar objective, however
```
8bitmp3, line 644: Suggestion:

```suggestion
There is also a new line of works, which optimize a similar objective, but
```

examples/vision/semisupervised_simclr.py
```diff
@@ -0,0 +1,658 @@
...
+without the use of any negatives:
+
+- [BYOL](https://arxiv.org/abs/2006.07733): momentum-encoder + no negatives
+- [SimSiam](https://arxiv.org/abs/2011.10566)
+  ([Keras example](https://keras.io/examples/vision/simsiam/)):
+  no momentum-encoder + no negatives
+
+In my experience these methods are more brittle (they can collapse to a constant
```
8bitmp3, line 652: Suggestion:

```suggestion
In my experience, these methods are more brittle (they can collapse to a constant
```

examples/vision/semisupervised_simclr.py
```diff
@@ -0,0 +1,658 @@
...
+representation, I could not get them to work using this encoder architecture).
+They are generally more dependent on the
+[model](https://generallyintelligent.ai/understanding-self-supervised-contrastive-learning.html)
+[architecture](https://arxiv.org/abs/2010.10241), however they improve
+performance at smaller batch sizes.
```
8bitmp3, line 657: Suggestion:

```suggestion
Even though they are generally more dependent on the
[model](https://generallyintelligent.ai/understanding-self-supervised-contrastive-learning.html)
[architecture](https://arxiv.org/abs/2010.10241), they can improve
performance at smaller batch sizes.
```

8bitmp3 (commented):
Great example. I left some suggestions, hope this helps @beresandras

examples/vision/semisupervised_simclr.py
beresandras: What did you mean to change here?

8bitmp3: Fixed it - just a small comma 😄

fchollet (approved):
It looks great, thanks for the update! I pushed various copyedits. You can now add the autogenerated files 👍 

Thanks @8bitmp3 for the help with the reviews!

beresandras:
I have added the autogenerated files. Note that I did make minor changes to the code so that the comparison with the baseline is stronger.

s-mrb (commented):
https://github.com/keras-team/keras-io/blob/1cfbd0ef056e8c4596a9e498d259725edf019114/examples/vision/semisupervised_simclr.py#L77 

spelling!

s-mrb:
also you forgot to update num_epochs, 20 in recent commit not 60.

https://github.com/keras-team/keras-io/blob/1cfbd0ef056e8c4596a9e498d259725edf019114/examples/vision/semisupervised_simclr.py#L267

beresandras:
Nice catch @s-mrb ! This actually required multiple changes in the text... Do you have any more suggestions? After that I will update the autogenerated files as well (Can I very carefully make the corresponding edits in them as well? Or should I rerun everything?).

fchollet:
> Can I very carefully make the corresponding edits in them as well? Or should I rerun everything?

Depends on the number of lines you need to change. Replicated edits in 3 files are tedious. For more than 10 lines, I'd recommend rerunning.

s-mrb:
@beresandras please double check this, i myself am not dead sure but I think there should not be `one`.

> image is more similar to its own versions with augmentations than to the ...

s-mrb:
Everything else seems perfect!

s-mrb (approved):

beresandras:
> please double check this, i myself am not dead sure but I think there should not be `one`.

Here "one" refers to the "representation" of a differently augmented version of the same image, I think it is correct that way.

s-mrb:
So there is one different version?

beresandras:
We augment each image in the batch twice, differently, and compare their representations in the loss.

s-mrb:
perfect then!

beresandras:
Okay. I have propagated the changes to the autogenerated files (edited, 7 changes each). I have previewed the site and the notebook, everything seemed fine to me.

s-mrb:
@beresandras please use this as link to your profile,

`[András Béres](https://www.linkedin.com/in/andras-beres-789190210)`
shows as
> [András Béres](https://www.linkedin.com/in/andras-beres-789190210)

The one used in recent commit is:
`[Andr\u00e1s B\u00e9res](www.linkedin.com/in/andras-beres-789190210)`
showed as
> [Andr\u00e1s B\u00e9res](www.linkedin.com/in/andras-beres-789190210)

s-mrb:
Also click the link, it won't work because it is appending to current link of page,
add https:// to not append to current address

beresandras:
Thanks! I have fixed that as well.

fchollet (approved):
LGTM, thank you. I gather this is ready to merge?

beresandras:
Yes, I think it's ready!

fchollet:

PR #474, opened 2021-05-21
code example contribution

jacoverster:


jacoverster:
@googlebot I signed it!

examples/vision/sample_size_estimate.py
```diff
@@ -0,0 +1,601 @@
...
+
+"""
+# Define augmentation function
+
+We'll use the RandAugment augmentation as shown in a recent Keras code example:
+[RandAugment for Image Classification for Improved
+Robustness](https://https://keras.io/examples/vision/randaugment/).
+"""
+
+rand_aug = iaa.RandAugment(n=3, m=7)
```
fchollet, line 292: Prefer using image augmentation layers instead, see e.g. https://keras.io/examples/vision/image_classification_from_scratch/#using-image-data-augmentation

examples/vision/sample_size_estimate.py
```diff
@@ -0,0 +1,601 @@
...
+            num_parallel_calls=AUTO,
+        )
+        .map(
+            lambda img, lab: (img, tf.one_hot(lab, num_classes)),
+            num_parallel_calls=AUTO,
+        )
+        .batch(batch_size)  # First batch images before augmenting
+        .map(
+            lambda img, lab: (
+                tf.py_function(rand_augment, [img], [tf.float32])[0],
```
fchollet, line 360: With image augmentation layers, you won't need a py_function and your pipeline will be more performant.

examples/vision/sample_size_estimate.py
```diff
@@ -0,0 +1,601 @@
...
+#             train_data, percentage)
+#         accuracy = train_model(train_ds, val_ds, test_ds)
+#         print(f"Accuracy: {accuracy}")
+#         sample_accuracy.append(accuracy)
+
+#     overall_accuracy.append(sample_accuracy)
+#     sample_sizes.append(nr_of_samples)
+#     # smallest_class_sizes.append(class_len)
+
+#     # Clear session and reset random seeds
```
fchollet, line 490: Is this really necesary?

examples/vision/sample_size_estimate.py
```diff
@@ -0,0 +1,601 @@
...
+- We train the model 5 times from scratch at each split and record the accuracy values.
+
+Note that this trains 20 models and will take some time. Make sure you have a GPU runtime
+active.
+
+To keep this example lightweight, sample data from a previous training run is provided
+after the commented out code below.
+"""
+
+# # Train all the sample models and calculate accuracy
```
fchollet, line 466: Rather than commenting an entire code block, put it in a function then point out to readers that the data you hard-code is supposed to be the output of the function.

fchollet (commented):
Thanks for the PR! Various comments. I do believe this is a very interesting concept to demonstrate as an example on keras.io!

examples/vision/sample_size_estimate.py
```diff
@@ -0,0 +1,601 @@
+"""
```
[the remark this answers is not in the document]
jacoverster: Thanks for the feedback @fchollet, I'm working on the changes this week. Quick question, do you prefer the changes made here in the comments as reviews or should I make changes and add a new PR?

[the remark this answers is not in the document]
fchollet: You can just commit to your fork/branch to update the PR. No new to create a new PR.

examples/vision/sample_size_estimate.py
```diff
@@ -0,0 +1,545 @@
...
+
+# Define variables, learning rate and number of epochs for fitting with TF
+a = tf.Variable(0.6)  # Use smart guesses for the weights to speed up fitting
+b = tf.Variable(0.05)
+learning_rate = 0.01
+training_epochs = 2000
+
+# Fit the exponential function to the data
+for epoch in range(training_epochs):
+    with tf.GradientTape() as g:
```
fchollet, line 477: `as tape`

examples/vision/sample_size_estimate.py
```diff
@@ -0,0 +1,545 @@
...
+        block_name: Str, layer name for example block_name = 'block4'.
+                    Checks if supplied string is in the layer name.
+        verbose: Int, 0 means silent, 1 prints out layers trainability status.
+
+    Returns:
+        Keras model with all layers after (and including) the specified
+        block_name to trainable, excluding BatchNormalization layers.
+    """
+
+    # Set the whole model trainable and
```
fchollet, line 219: and?

examples/vision/sample_size_estimate.py
```diff
@@ -0,0 +1,545 @@
...
+    model.trainable = True
+    set_trainable = False
```
fchollet, line 221: It appears you don't need this variable

examples/vision/sample_size_estimate.py
```diff
@@ -0,0 +1,545 @@
...
+# Define seed and fixed variables
+seed = 42
+tf.random.set_seed(seed)
+np.random.seed(seed)
+AUTO = tf.data.AUTOTUNE
+
+"""
+# Load TensorFlow dataset
+
+We'll be using the [TF Flowers dataset](https://bit.ly/34FQxWc).
```
fchollet, line 55: Use the direct link rather than bitly, https://www.tensorflow.org/datasets/catalog/tf_flowers

fchollet (commented):
Thanks for the update! This is great. The main thing that seems strange right now is the workarounds that deal with the fact that tf.data Datasets aren't indexable. Let's switch to NumPy arrays.

examples/vision/sample_size_estimate.py
```diff
@@ -0,0 +1,545 @@
...
+# Define augmentation function
+
+Define image augmentation using keras preprocessing layers.
+"""
+
+image_augmentation = keras.Sequential(
+    [
+        preprocessing.RandomFlip("horizontal"),
+        preprocessing.RandomRotation(0.1),
+        preprocessing.RandomZoom(height_factor=(-0.1, -0), fill_mode="constant"),
```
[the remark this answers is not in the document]
jacoverster: Thanks for the feedback, I was not aware of this. Set to default.

examples/vision/sample_size_estimate.py
```diff
@@ -0,0 +1,545 @@
...
+        )
+        .map(
+            lambda img, lab: (img, tf.one_hot(lab, num_classes)),
+            num_parallel_calls=AUTO,
+        )
+        .cache()
+        .batch(batch_size)
+        .prefetch(AUTO)
+    )
+    return train_ds, val_ds, num_subsamples
```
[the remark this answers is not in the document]
jacoverster: The num_subsamples output is used to capture the split sample_sizes when training iteratively. To create the x-values for our plot later. This will fall away when moving to a numpy based approach.

examples/vision/sample_size_estimate.py
```diff
@@ -0,0 +1,545 @@
...
+    )  # Resize images
+    .map(
+        lambda img, lab: (img, tf.one_hot(lab, num_classes)), num_parallel_calls=AUTO
+    )  # Convert to one-hot labels
+    .cache()
+    .batch(batch_size)
+    .prefetch(AUTO)
+)
+
+# Extract true labels from the test set (to be used for model evaluation later)
```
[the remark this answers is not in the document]
jacoverster: Agreed, this seems like a much better approach. Working with tf.data.dataset in this case is very slow. Thanks for the suggestion.

examples/vision/sample_size_estimate.py
```diff
@@ -0,0 +1,545 @@
...
+        y_pred = exp_func(x, a, b)
+        cost_function = squared_error(y_pred, mean_acc)
+    # Get gradients and compute adjusted weights
+    gradients = g.gradient(cost_function, [a, b])
+    a.assign_sub(gradients[0] * learning_rate)
+    b.assign_sub(gradients[1] * learning_rate)
+print(f"Curve fit weights: a = {a.numpy()} and b = {b.numpy()}.")
+
+# We can now estimate the accuracy for the whole training set using exp_func
+max_acc = exp_func(num_train_samples, a, b)
```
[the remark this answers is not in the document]
jacoverster: @fchollet are you referring to the whole curve fit and prediction from the curve section?

examples/vision/sample_size_estimate.py
jacoverster: It is required as the function sets it to True when it reaches the designated layer. It should keep setting all subsequent layers to trainable even if the layer name does not contain the block_name.

jacoverster: I do however, notice that I don't need to set model.trainable = True first, I can drop this and avoid setting layers to "not trainable" again.

examples/vision/sample_size_estimate.py
```diff
@@ -0,0 +1,545 @@
...
+
+To iteratively train a model over several subsample sets we need to create some functions
+for splitting the dataset and training the model.
+
+"""
+
+
+def create_subsets(dataset, fraction):
+    """Creates a subset from a dataset based on a specified fraction. It is
+        divided into training (90%) and validation (10%) sets.
```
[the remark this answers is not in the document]
jacoverster: No longer needed, I completely dropped create_subsets after using NumPy. I now just use validation_split=0.1 in model.fit.

examples/vision/sample_size_estimate.py
```diff
@@ -0,0 +1,545 @@
...
+(train_data, test_data), ds_info = tfds.load(
+    dataset_name,
+    split=["train[:90%]", "train[90%:]"],
+    shuffle_files=True,
+    as_supervised=True,
+    with_info=True,
+)
+
+# Extract number of classes and list of class names
+num_train_samples = train_data.cardinality().numpy()
```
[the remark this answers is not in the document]
jacoverster: Noted, I no longer use this now that I've got NumPy arrays.

examples/vision/sample_size_estimate.py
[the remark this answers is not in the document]
jacoverster: I did a quick bench test on the tf.data performance vs NumPy and script run-time went down from 0:32:09 to 0:18:52 (incl. iterative model training). Wowzer!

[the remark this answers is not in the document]
jacoverster: Wowzer! I just ran a quick test to measure performance of the tf.data script vs NumPy script, runtime came down from 32min to about 18min (incl. iterative training). Nice!

examples/vision/sample_size_estimate.py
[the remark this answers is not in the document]
fchollet: Yes

examples/vision/sample_size_estimate.py
```diff
@@ -0,0 +1,479 @@
...
+        metrics=[keras.metrics.AUC(name="auc"), "acc"],
+        optimizer=keras.optimizers.Adam(learning_rate=1e-4),
+        patience=5,
+        epochs=total_epochs,
+    )
+
+    # Calculate model accuracy on the test set
+    y_pred = model.predict(img_test)
+    y_pred_indices = np.argmax(y_pred, axis=1)
+    acc = np.sum(np.equal(y_pred_indices, np.argmax(label_test, axis=1))) / len(
```
fchollet, line 304: Why not use `model.evaluate()`?

fchollet (commented):
Thanks for the update. Looking good!

I made some copyedits, please pull them first.

examples/vision/sample_size_estimate.py
jacoverster: Ah, yes thanks for the suggestion. Much more elegant.

jacoverster: Thanks for the copyedits, much appreciated. Looking really good now. I've made a few minor edits and changed in model.evaluate().

examples/vision/sample_size_estimate.py
```diff
@@ -0,0 +1,474 @@
+"""
+Title: Estimating required sample size for model training
+Author: [JacoVerster](https://twitter.com/JacoVerster)
+Date created: 2021/05/20
+Last modified: 2021/06/05
+Description: Estimate the model accuracy for a specific training set size.
```
fchollet, line 6: This is the reverse of the title... how about "Modeling the relationship between training set size and model accuracy."

fchollet (commented):
Everything is looking good! 👍 

Just one last comment. You can add the generated files.

examples/vision/sample_size_estimate.py
jacoverster: I was also contemplating this, good suggestion thanks.

jacoverster:
All files added and one final complete review and proofread. Thanks for all the help getting this example up to standard, such a privilege working with you on this.

fchollet:
My pleasure! Thank you for the great contribution!

fchollet (approved):

PR #500, opened 2021-06-15
Use dictionaries for both `loss` and `loss_weights` consistently

ronshapiro:


guides/functional_api.py
```diff
@@ -390,7 +390,10 @@ def get_model():
         "priority": keras.losses.BinaryCrossentropy(from_logits=True),
         "department": keras.losses.CategoricalCrossentropy(from_logits=True),
     },
-    loss_weights=[1.0, 0.2],
+    loss_weights={
+        "priority": 1.0,
+        "department": 0.2,
+    },
```
MarkDaoust, line 396: I was wondering how it lines up the outputs to the losses given that the outputs are defined as: `outputs=[priority_pred, department_pred]`

But now I see that it's using the layer names. 

Should there be an example using `outputs={'priority': priority_pred, 'department': department_pred}`? I find that form is always clearer for the user of a multiple output model.

MarkDaoust (approved):

guides/functional_api.py
ronshapiro: Updated the description text so it's clearer where these names come from.

I don't know what's better, specifying the names in the layer or in the model.fit() call. I leave that to you.

MarkDaoust: Regardless of that Idea, I think this is a good change. Let's get it merged.

MarkDaoust (changes requested):
One last think, I should have mentioned on the first pass:

Could you run the script to update the generated files, and add those to the PR as well?

I believe the command is (from the `scripts/` directory):

```
python autogen.py add_guide functional_api.py
```

fchollet (commented):
LGTM, please replicate the fix in the md and ipynb files.

ronshapiro:
Done.

MarkDaoust (approved):

MarkDaoust:
Thanks.

PR #503, opened 2021-06-15
fixing bugs for kt doc render

haifeng-jin:
* renaming `templates/api/keras-tuner` to `templates/api/keras_tuner`, otherwise the descriptions to the api ref won't render.
* adding back `templates/guides/keras_tuner/index.md`, which caused the current master test to fail. It may be ignored by `.gitignore`. No further action needed unless we modify this file in the future. If we modify it, we need to `add -f ...`.


fchollet (approved):
LGTM, thanks

PR #506, opened 2021-06-17
Example on Image Compression using Keras added

sayannath:
One of the applications of autoencoders is image compression. So adding an example for image compression using `mnist` dataset. For reference here is the [Colab Notebook](https://colab.research.google.com/drive/1pAhaaX__YV6BxLAxFqKL5OmQTl57Kmor?usp=sharing)


fchollet (commented):
Thanks for the PR!

I think image compression is a great topic, and deep learning methods have shown to have some potential there. However, there is a lot more to it than simply training an autoencoder with a small number of intermediate units.

Here are my recommendations for a great code walkthrough about image compression:

- Use a dataset of real images, because you need to be able to visually evaluate the perceptual lossiness of your compression scheme. MNIST doesn't let you do that.
- The example should demonstrate gains on test images (not training images) over a non-learned baseline algorithm like JPEG. This is very doable for a dataset of specialized images (like, say, flowers or pets or landscapes).

Here are two papers you could draw inspiration from:

- https://arxiv.org/abs/1608.05148
- https://arxiv.org/abs/1708.00838

sayannath:
> Thanks for the PR!
> 
> I think image compression is a great topic, and deep learning methods have shown to have some potential there. However, there is a lot more to it than simply training an autoencoder with a small number of intermediate units.
> 
> Here are my recommendations for a great code walkthrough about image compression:
> 
> * Use a dataset of real images, because you need to be able to visually evaluate the perceptual lossiness of your compression scheme. MNIST doesn't let you do that.
> * The example should demonstrate gains on test images (not training images) over a non-learned baseline algorithm like JPEG. This is very doable for a dataset of specialized images (like, say, flowers or pets or landscapes).
> 
> Here are two papers you could draw inspiration from:
> 
> * https://arxiv.org/abs/1608.05148
> * https://arxiv.org/abs/1708.00838

Will surely try this!

sayannath:
Hi @fchollet! Reproducing this paper - [An End-to-End Compression Framework Based on Convolutional Neural Networks](https://arxiv.org/abs/1708.00838). Will notify you after I finish. Thanks for guiding.

PR #511, opened 2021-06-18
Add prerequisites installs

ZackPashkin:
Hi! I've found your example very interesting! But I encountered some missing dependencies in a notebook, I just added the necessary prerequisites so anyone can run your notebook smoothly! Thank you!

### DIFF
--- a/examples/rl/ipynb/deep_q_network_breakout.ipynb
+++ b/examples/rl/ipynb/deep_q_network_breakout.ipynb
@@ -1,357 +1,483 @@
 {
- "cells": [
-  {
-   "cell_type": "markdown",
-   "metadata": {
-    "colab_type": "text"
-   },
-   "source": [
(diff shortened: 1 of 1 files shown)


ZackPashkin:
@googlebot I signed it!

fchollet:
Thanks for the PR. If any dependency isn't included as part of the default Colab runtime, we should simply specify in the introduction, "This example requires [package], which can be installed via [command]". Please only edit the `.py` file at first, and only edit the other 2 files after approval.

ZackPashkin:
> Thanks for the PR. If any dependency isn't included as part of the default Colab runtime, we should simply specify in the introduction, "This example requires [package], which can be installed via [command]". Please only edit the `.py` file at first, and only edit the other 2 files after approval.

Thank you for reply! Fixed https://github.com/keras-team/keras-io/pull/519#issue-675915974

PR #515, opened 2021-06-20
Delete extraneous conv layer from example

jkhales:
The second Conv1D layer is strange and does make a significant difference in the accuracy of model. Seems like a typo?

Results with only a single Conv1D layer:

Screen Shot 2021-06-20 at 6 12 39 AM


jkhales:
> @googlebot I signed it!

fchollet (commented):
A model that only max pools over windows of 7 tokens would not be an actual sequence model. Having 2 conv layers demonstrate how conv layers can be stacked to create a sequence model.

PR #527, opened 2021-06-27
add wgan-graphs.py tutorial

akensert:
A generative model for small graphs (specifically small molecular graphs).


akensert:
@googlebot I signed it!

PR #535, opened 2021-06-29
Add densenet and cspnet example

Varchita-Beena:
The example is about DenseNet, CSPNet their varaints on Cifar10 dataset


Varchita-Beena:
> @googlebot I signed it!

@googlebot I signed it!

examples/vision/densenet_cspnet_variants.py
```diff
@@ -0,0 +1,396 @@
...
+"""
+"""shell
+densenet_cspnet_variants.ipynb
+
+Automatically generated by Colaboratory.
+
+Original file is located at
+    https://colab.research.google.com/drive/1ejnjgRzfo60FWXMp8_HyzQJ_5kBLIdwd
+
+##DenseNet
```
fchollet, line 16: The example should explain the point of these models -- why should a reader be interested in them? How should they use them? How do they compare to alternatives?

The example will also need to be polished much further, as it currently has many issues. Please check out README instructions for how to develop and preview code examples.

examples/vision/densenet_cspnet_variants.py
```diff
@@ -0,0 +1,396 @@
...
+    BatchNormalization,
+    Reshape,
+    Flatten,
+)
+from tensorflow.keras.optimizers import Adam
+from tensorflow.keras.losses import CategoricalCrossentropy
+from tensorflow.keras import backend as K
+from tensorflow.keras.utils import to_categorical
+import matplotlib.pyplot as plt
+from sklearn.model_selection import train_test_split as split
```
xybrs, line 99: sklearn is unnecessary dependency in this example, we can use split from numpy

pcoet:
Hi @Varchita-Beena, thanks for this pull request. Would you like to update the example, based on the feedback? Otherwise we'll close the request.

Varchita-Beena:
Yes I would like to update the example based on the feedback. I will complete it soon.
Thanks. 
 
  On Mon, Aug 14, 2023 at 8:30 AM, David ***@***.***> wrote:   


Hi @Varchita-Beena, thanks for this pull request. Would you like to update the example, based on the feedback? Otherwise we'll close the request.

PR #549, opened 2021-07-16
Fixing the comment line where the sequence length about the dataset was incorrect. 

jorgpg5:
The FordA sequences length value is 500 instead of 176. This situation can be seen in the jupyter notebook and the .py file.

According to the paper and the code, the sequences length value is 500 instead of 176. I know that this is a minor detail, but it might be important to have that value correct for learning purposes.

image

Page 92 [Source](https://www.cs.ucr.edu/~eamonn/time_series_data_2018/BriefingDocument2018.pdf)
image

[Source](http://www.j-wichard.de/publications/FordPaper.pdf)
image

### DIFF
--- a/examples/timeseries/ipynb/timeseries_classification_from_scratch.ipynb
+++ b/examples/timeseries/ipynb/timeseries_classification_from_scratch.ipynb
@@ -10,7 +10,7 @@
     "\n",
     "**Author:** [hfawaz](https://github.com/hfawaz/)<br>\n",
     "**Date created:** 2020/07/21<br>\n",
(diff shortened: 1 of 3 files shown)


jorgpg5:
@googlebot I signed it!

fchollet (changes requested):
Thanks for the PR. Please also modify the `.md` file.

jorgpg5:
Thanks, @fchollet. Sorry about that. I totally forgot about the .md file. Now it's done. 

Thank you!

fchollet (approved):

PR #559, opened 2021-07-23
DOC: Correct text_dataset and image_dataset examples

TPXP:
The functions are in their respective preprocessing packages


TPXP:
@googlebot I signed it!

fchollet (commented):
This is incorrect, the proper location is in `tf.keras.preprocessing`.

TPXP:
My bad, sorry for the confusion. I was import keras directly, not from tensorflow.

PR #574, opened 2021-08-05
Corrected the order of Normalization layer in Transformer example

juanantoniobellido:
According to the paper "Attention is all you need"The normalization layer goes after MultiheadAttenttionand after feed-forward layer


juanantoniobellido:
@googlebot I signed it!

fchollet (approved):
LGTM, thanks for the PR! Please also update the generated files (`.md` and `.ipynb`).

PR #579, opened 2021-08-06
Update preprocessing layer guide to list entire layer symbols

mattdangerw:
This will be helpful generally for clearing up confusion with
the symbol locations for the 2.6 release, as well as allowing
these symbols to cross link with layer documentation pages on
tensorflow.org.


fchollet (approved):
LGTM, thanks for the PR!

PR #585, opened 2021-08-09
Use absolute link to TensorBoard callback in Training & evaluation with the built-in methods guide

8bitmp3:
A proposed solution to fix a 404 on tensorflow.org. Because of case-sensitivity in URLs, the link on the same guide -  https://www.tensorflow.org/guide/keras/train_and_evaluate - points to https://www.tensorflow.org/api_docs/python/tf/keras/callbacks/tensorboard/ which is a "404" (requires `...TensorBoard/`)

### DIFF
--- a/guides/ipynb/training_with_built_in_methods.ipynb
+++ b/guides/ipynb/training_with_built_in_methods.ipynb
@@ -1835,7 +1835,7 @@
    },
    "source": [
     "For more information, see the\n",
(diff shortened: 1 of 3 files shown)


fchollet (approved):
LGTM, please also update the md and ipynb files.

8bitmp3:
Done.

fchollet (approved):

PR #591, opened 2021-08-13
Fix FAQ (```endshell breaks things)

ahasselbring:
didn't insert a space into "whenrunning" since #560 already does this.

This is what the page currently looks like:

grafik


ahasselbring:
@googlebot I signed it!

fchollet (approved):
Thanks for the PR.

PR #599, opened 2021-08-20
Use self.num_iterations instead of global variable

floscha:
This will not change the behavior of the script but I believe the class was intended to use `self.num_iterations` in this place since the class sets `self.num_iterations` in `__init__()` but doesn't use it anywhere.


fchollet (commented):
Thanks for the PR. Looks good to me. Please also apply the fix to the ipynb and md files.

floscha:
Done.

fchollet (approved):

PR #603, opened 2021-08-22
Add inference time of models

BbChip0103:
Regarding [Keras issue #13619](https://github.com/keras-team/keras/issues/15004)

Hi, thank you for your great work.

# What is problem?
In the [Keras application document](https://keras.io/api/applications/), It describes model's size, accuracy, etc.
When we use the model for our real world project, inference time is most important feature like 'accuracy'.
Of course, inference time generally tend to follow number of parameters. But these are not perfectly proportional.
So, checking the inference time of models will be useful when we search model for our project.

# Method
As the inference time depends on the spec of machine, we can use fixed and general environment for the checking 
(I used google-colab env. Please see [test code](https://colab.research.google.com/drive/1utZlKq02tlB20eU9v23PfQMB8WHg7u32?usp=sharing))

# Result
image

### Previous document
image
image
### Updated document
image
image
  
Thank you.


BbChip0103:
@googlebot I signed it!

fchollet:
Thanks for the PR! This is indeed very useful information to add. I would recommend:

- Define it as "Time per inference step (in ms)"
- Make sure you're optimizing inference speed by using a prefetched `tf.data.Dataset`
- Compute it for both CPU-only and single-GPU, because many people do inference on CPU

In particular, you're currently using `result = %timeit -n 10 -r 30 -o model.predict(x)`. I recommend doing instead:

```python
input_shape = tuple(model.input.shape[1:])
batch_size = 32
n = 20
first_batch = np.random.random((batch_size,) + input_shape).astype('float32')
full_array = np.random.random((batch_size * n,) + input_shape).astype('float32')
ds = tf.data.Dataset.from_tensor_slices(full_array).batch(batch_size).prefetch(4)

### Warmup model
model.predict(first_batch)

result = %timeit -o model.predict(ds)
result /= n
```

(*untested code, may need touch ups)

BbChip0103:
@fchollet Thank you for your comment.
I changed my code and added CPU benchmark based on your example.
([GPU test code](https://colab.research.google.com/drive/1IVbsAjAdW3Udh5a6lxYr_ILkyH4eCYHY?usp=sharing). [CPU test code](https://colab.research.google.com/drive/1K7jYa82907kT_SG__FqMewE0UbATS7xH?usp=sharing))

And changed document is like below.
image
...
image

fchollet (commented):
Awesome, thank you for the contribution!

fchollet (approved):

PR #605, opened 2021-08-24
image_dataset_from_directory function exists in .preprocessing not in .image

amshrbo:
- Fixing typo in importing `image_dataset_from_directory` function
- This function doesn't exist in the `.image` module, but in the `.preprocessing`
- And the same function used in this [article](https://keras.io/api/preprocessing/image/#image_dataset_from_directory-function) properly


amshrbo:
@googlebot I signed it!

fchollet (approved):

PR #608, opened 2021-08-25
Update docstring links to keras.io for keras guides

mattdangerw:
We will still have a number of links pointing to tensorflow.org,
but this links keras.io guides from keras.io symbol documentation
where possible.


fchollet (approved):
LGTM, thanks!

PR #613, opened 2021-08-27
Update FAQ page

rchao:
Update the answer to the question: What if I need to customize what `fit()` does?


fchollet (approved):
LGTM, thanks!

rchao:
Thanks for the review!

PR #635, opened 2021-09-13
Self-supervised contrastive learning with NNCLR

Rishit-dagli:
In this example we will be implementing NNCLR as proposed in the paper [With a LittleHelp from My Friends: Nearest-Neighbor Contrastive Learning of Visual Representations](https://arxiv.org/abs/2104.14548), by Google Research and DeepMind which:
- learns  self-supervised representations that go beyond single instance positives which allows to learn better features that are invariant to different viewpoints, deformations, and even intra-class variations
- Clustering based methods offer a great approach to go beyond single instance positives, but assuming the entire cluster  to be positives could hurt performance due to early over-generalization. Instead NNCLR uses nearest neighbors in the learned representation space as positives
- increases the performance of existing contrastive learning methods like [SimCLR](https://arxiv.org/abs/2002.05709)([Keras Example](https://keras.io/examples/vision/semisupervised_simclr))
- reduces the reliance of self-supervised methods on data augmentation strategies

---

This can be tried out in a Colab Notebook as well: [Open In Colab](https://drive.google.com/file/d/1hEINnjReMGY0TDINlWlOIwTnfP0NM72c/view?usp=sharing).

### DIFF
--- a/examples/vision/nnclr.py
+++ b/examples/vision/nnclr.py
@@ -562,7 +562,6 @@ trained for more epochs for better performance.
 The following two metrics can be used for monitoring the pretraining performance which we
 also log (taken from [this Keras
 example](https://keras.io/examples/vision/semisupervised_simclr/#selfsupervised-model-for-contrastive-pretraining)):
-example](https://keras.io/examples/vision/semisupervised_simclr/#selfsupervised-model-for-contrastive-pretraining)):
 
 - Contrastive accuracy: Self-supervised metric, the ratio of cases in which the
 representation of an image is more similar to its differently augmented version's one,
--- a/examples/vision/nnclr.py
+++ b/examples/vision/nnclr.py
@@ -58,7 +58,7 @@ the positive pairs using nearest-neighbours. A support set is used as memory dur
 training; similar to a queue (i.e. first-in-first-out) like in
 [MoCo](https://arxiv.org/abs/1911.05722).
 
-This example requires TensorFlow 2.4 or higher as well as `tensorflow_datasets` which can
+This example requires TensorFlow 2.6 or higher as well as `tensorflow_datasets` which can
 be installed with this command:
 """
 
--- a/examples/vision/nnclr.py
+++ b/examples/vision/nnclr.py
@@ -305,8 +305,8 @@ def augmenter(brightness, contrast, saturation, hue, name, scale):
     return keras.Sequential(
         [
             layers.Input(shape=input_shape),
-            layers.experimental.preprocessing.Rescaling(1 / 255),
-            layers.experimental.preprocessing.RandomFlip("horizontal"),
+            layers.Rescaling(1 / 255),
+            layers.RandomFlip("horizontal"),
             RandomResizedCrop(scale=scale, ratio=(3 / 4, 4 / 3)),
             RandomColorJitter(
                 brightness=brightness, contrast=contrast, saturation=saturation, hue=hue
--- a/examples/vision/nnclr.py
+++ b/examples/vision/nnclr.py
@@ -18,21 +18,19 @@ expensive annotations like image classification.
 ### Contrastive Learning
 
 A broad category of self-supervised learning techniques are those that use contrastive
-losses, which have been used in a wide range of computer vision applications like [image
-similarity](https://www.jmlr.org/papers/v11/chechik10a.html), [Dimensionality reduction
-(DrLIM)](http://yann.lecun.com/exdb/publis/pdf/hadsell-chopra-lecun-06.pdf) and [Face
-verification/
-identification](https://openaccess.thecvf.com/content_cvpr_2015/html/Schroff_FaceNet_A_Uni
-fied_2015_CVPR_paper.html). These methods learn a latent space that draws positive
-samples together while pushing apart negative samples.
+losses, which have been used in a wide range of computer vision applications like
+[image similarity](https://www.jmlr.org/papers/v11/chechik10a.html)
+, [Dimensionality reduction (DrLIM)](http://yann.lecun.com/exdb/publis/pdf/hadsell-chopra-lecun-06.pdf)
+and [Face verification/identification](https://openaccess.thecvf.com/content_cvpr_2015/html/Schroff_FaceNet_A_Unified_2015_CVPR_paper.html)
+. These methods learn a latent space that draws positive samples together while
+pushing apart negative samples.
 
 ### NNCLR
 
-In this example we will be implementing NNCLR as proposed in the paper [With a Little
-Help from My Friends:
-Nearest-Neighbor Contrastive Learning of Visual
-Representations](https://arxiv.org/abs/2104.14548), by Google Research and DeepMind
-which:
+In this example we will be implementing NNCLR as proposed in the paper
+[With a Little Help from My Friends: Nearest-Neighbor Contrastive Learning of Visual Representations](https://arxiv.org/abs/2104.14548)
+, by Google Research and DeepMind which:
+
 - learns  self-supervised representations that go beyond single instance positives which
 allows to learn better features that are invariant to different viewpoints, deformations,
 and even intra-class variations
@@ -41,8 +39,7 @@ but assuming the entire cluster  to be positives could hurt performance due to e
 over-generalization. Instead NNCLR uses nearest neighbors in the learned representation
 space as positives
 - increases the performance of existing contrastive learning methods like
-[SimCLR](https://arxiv.org/abs/2002.05709)([Keras
-Example](https://keras.io/examples/vision/semisupervised_simclr))
+[SimCLR](https://arxiv.org/abs/2002.05709)([Keras Example](https://keras.io/examples/vision/semisupervised_simclr))
 - reduces the reliance of self-supervised methods on data augmentation strategies
 
 Here is a great visualization by the paper authors showing how NNCLR builds on ideas from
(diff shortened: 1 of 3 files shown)


Rishit-dagli:
The black CI fails for this PR but when I try to reformat with black it shows me no changes would be made:
image

fchollet:
Thanks for the PR!

> The black CI fails for this PR but when I try to reformat with black it shows me no changes would be made:

This has to be due to a discrepancy in black versions.

Rishit-dagli:
Ah, this is really interesting since I had specifically run:

```sh
pip install black==19.10b0
```

It turns out GitHub Codespaces installs black 21.8b0 for you, I resorted to using black as a package (`python -m black *` instead of `black *`) which seems to use the right version in Codespaces. It seems to work well now.

examples/vision/nnclr.py
```diff
@@ -0,0 +1,642 @@
...
+experimented on). We here use 10,000 to show a working example.
+"""
+
+AUTOTUNE = tf.data.experimental.AUTOTUNE
+shuffle_buffer = 5000
+# The below two values are taken from https://www.tensorflow.org/datasets/catalog/stl10
+labelled_train_images = 5000
+unlabelled_images = 100000
+
+temperature = (0.1,)
```
fchollet, line 96: This should probably be a float, not a tuple

examples/vision/nnclr.py
```diff
@@ -0,0 +1,642 @@
...
+"""
+## Augmentations
+
+Other self-supervised techniques like [SimCLR](https://arxiv.org/abs/2002.05709),
+[BYOL](https://arxiv.org/abs/2006.07733), [SwAV](https://arxiv.org/abs/2006.09882) etc.
+rely heavily on a well designed data augmentation pipeline to get the best performance.
+However, NNCLR is _less_ dependent on complex augmentations as nearest-neighbors already
+provide richness in sample variations. A few common techniques often included
+augmentation pipeliness are:
+- Random resized crops
```
fchollet, line 172: Add line break before list

examples/vision/nnclr.py
```diff
@@ -0,0 +1,642 @@
...
+Self supervised learning is particularly very helpful when you do not have access to very
+limited labeled training data but you can manage to build a large corpus of unlabeled
+data as shown by previous methods like [SEER](https://arxiv.org/abs/2103.01988),
+[SimCLR](https://arxiv.org/abs/2002.05709), [SwAV](https://arxiv.org/abs/2006.09882) and
+more.
+
+You should also take a look at the blogs for these papers which neatly show that it is
+possible to achieve good results with few class labels by first pretraining on a large
+unlabeled dataset and then fine-tuning on a smaller labeled dataset:
+- [Advancing Self-Supervised and Semi-Supervised Learning with
```
fchollet, line 628: Add line break before list

examples/vision/nnclr.py
```diff
@@ -0,0 +1,642 @@
...
+SimCLR](https://ai.googleblog.com/2020/04/advancing-self-supervised-and-semi.html)
```
fchollet, line 629: Make sure that all links are on a single line (here and below)

examples/vision/nnclr.py
```diff
@@ -0,0 +1,642 @@
...
+identification](https://openaccess.thecvf.com/content_cvpr_2015/html/Schroff_FaceNet_A_Uni
+fied_2015_CVPR_paper.html). These methods learn a latent space that draws positive
+samples together while pushing apart negative samples.
+
+### NNCLR
+
+In this example we will be implementing NNCLR as proposed in the paper [With a Little
+Help from My Friends:
+Nearest-Neighbor Contrastive Learning of Visual
+Representations](https://arxiv.org/abs/2104.14548), by Google Research and DeepMind
```
fchollet, line 34: Here too

examples/vision/nnclr.py
```diff
@@ -0,0 +1,642 @@
...
+which:
+- learns  self-supervised representations that go beyond single instance positives which
```
fchollet, line 36: Add line break before list

fchollet (commented):
Thanks for the PR! It looks great 👍

examples/vision/nnclr.py
```diff
@@ -0,0 +1,642 @@
...
+
+"""
+### Color augmentation
+
+We will now do Random color jitters augmentation where we randomly change the brightness,
+contrast and saturation of an image.
+"""
+
+
+class RandomColorJitter(layers.Layer):
```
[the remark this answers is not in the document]
Rishit-dagli: Got it, a quick question do you think it would make sense to probably just add 1 transformation (like only random brightness)?

examples/vision/nnclr.py
Rishit-dagli: Fixed this as well

examples/vision/nnclr.py
Rishit-dagli: Fixed this

examples/vision/nnclr.py
Rishit-dagli: Got it

examples/vision/nnclr.py
Rishit-dagli: Fixed this

examples/vision/nnclr.py
Rishit-dagli: Thanks for spotting this, fixed it

examples/vision/nnclr.py
Rishit-dagli: Yes, in fact, a tuple would not even work, right? The brackets seem to have been added by black due to a trailing comma. Fixed this.

examples/vision/nnclr.py
```diff
@@ -0,0 +1,642 @@
...
+will later use this function as both the contrastive augmenter and classification
+augmenter.
+"""
+
+
+def augmenter(brightness, contrast, saturation, hue, name, scale):
+    return keras.Sequential(
+        [
+            layers.Input(shape=input_shape),
+            layers.experimental.preprocessing.Rescaling(1 / 255),
```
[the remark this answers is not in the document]
Rishit-dagli: Ah, yes I was experimenting on Kaggle (TensorFlow 2.4) so forgot to change this

Rishit-dagli:
@fchollet Thanks for your insightful reviews, I have implemented all but one of them for which I had a follow-up question.

examples/vision/nnclr.py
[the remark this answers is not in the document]
fchollet: Yes, I think it won't really change the results. Random brightness only uses a single random coefficient for all color channels, so it isn't the same thing as one unique coefficient per channel, but the impact of the change should be minimal. So you could just do that.

fchollet (commented):
Besides the last comment, everything is look great! I pushed some final copyedits.

examples/vision/nnclr.py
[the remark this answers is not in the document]
Rishit-dagli: Thanks, got it!

Rishit-dagli:
> Besides the last comment, everything is look great! I pushed some final copyedits.

Thanks, updated as you suggested.

fchollet (approved):
Thank you! Please add the generated files.

Rishit-dagli:
Thanks @fchollet , Added the genrated files

fchollet:

beresandras:
Dear @Rishit-dagli !

I have strong concerns with this code example, as I will explain below.

I find recent self-supervised representation learning methods fascinating, therefore I follow related literature closely, I am the author of the [Keras SimCLR code example](https://keras.io/examples/vision/semisupervised_simclr/), and I also have a [repository](https://github.com/beresandras/contrastive-classification-keras) where I have implemented 5 self-supervised algorithms, from which the latest, NNCLR, [was added 2 weeks ago](https://github.com/beresandras/contrastive-classification-keras/commit/fc345f9fdc58d2fa943ff16b2ecdf4e801bf2522).

For these reasons I was excited to dive into this new code example based on the NNCLR algorithm. To my shock, the code seemed oddly familiar, it almost seemed exactly the same as the corresponding parts of my NNCLR implementation. Upon closer inspection, I could not find any other differences between the two, except the following:
- all comments have been removed (a single comment has been added instead)
- the RandomColorJitter custom preprocessing layer was reduced to RandomBrightness during the pull request discussion, though initially it was the same as well
- transposed versions of the similarity matrices have been added to the loss, which required minor naming changes
- the network is finetuned after pretraining

Everything else, data pipeline, preprocessing, encoder architecture, metrics, loss, hyperparameters, and naming is the same, while the author does not refer to the repository in any way. The author created the pull request 5-6 days after I have added my NNCLR implementation.

There are also notable implementation differences with the research paper, that follow my implementation but are not mentioned in the text of the example:
- straight-through gradient estimation in the nearest neighbour operation instead of stop-gradient: this is an implementation trick I found, which boosts performance at this scale, but is not used in the paper
- missing predictor head: I decided against the usage of a predictor head for simplicity, because the ablations showed that it only adds a minor performance increase
- correlation accuracy: this is an unsupervised metric I experimented with, it is based on the BarlowTwins loss and is not used anywhere else. Its explanation is completely missing from the main text.

My code was released under the MIT license, which requires the preservation of the copyright notice. The author did not fulfill that.

Though I find the behaviour of the author highly unethical, I also believe that the collection of Keras code examples is a unique and powerful resource for learning, and I would not want their number to be reduced. Therefore I propose the following resolution to this issue:
- the following note should be added to the top: "The code of this example is heavily based on [this repository](https://github.com/beresandras/contrastive-classification-keras) by András Béres."
- adding back the code comments and mentioning important differences with the research paper would be beneficial as well

@fchollet and other members of the Keras community, I would love to hear your thoughts about this matter.

Rishit-dagli:
Hi @beresandras ,
Thanks for pointing this out.

I respectfully disagree that this is a copy of your repo, I have made quite a few changes taking the starter code from your really great SimCLR example.

However, this is a sincere mistake from my end forgetting to credit the repo while crediting your Keras Example. This is the first and the only time I have forgotten to give due credit and I will update it. Thanks once again for pointing this out.

PR #645, opened 2021-09-24
Update about.md

jvishnuvardhan:
I think the Python versions supported are different from what was listed on TF webpage.

https://www.tensorflow.org/install/source#tested_build_configurations

Based on TF webpage, I updated python versions that are supported by keras as 3.6 to 3.9


jvishnuvardhan:
@qlzh727 Please take a look at this PR. There is a discrepancy between TF webpage and Keras webpage on the Python versions that are supported. Please take a look and let me know what you think. Thanks!

fchollet (approved):

PR #646, opened 2021-09-25
Large-scale multi-label text classification

soumik12345:
In this example, we will build a multi-label text classifier to predict the subject areas of arXiv papers from their abstract bodies. This type of classifier can be useful for conference submission portals like [OpenReview](https://openreview.net/). Given a paper abstract, the portal could provide suggestions on which areas the underlying paper would best belong to.

Please find corresponding Colab Notebook [here](https://colab.research.google.com/drive/11ijMQ3c03I2LvWjdOVk5VnWhKbhOj1ss?usp=sharing).


sayakpaul:
Also, this PR follows [this call for contribution specs](https://github.com/keras-team/keras-io/blob/master/call_for_contributions.md#large-scale-multi-label-text-classification).

sayakpaul:
@googlebot I consent.

examples/nlp/multi_label_classification.py
```diff
@@ -0,0 +1,382 @@
...
+"""
+## Vocabulary size for vectorization
+
+Before we feed the data to our model we need to represent them as numbers. For that
+purpose, we will use the
+[`TextVectorization` layer](https://keras.io/api/layers/preprocessing_layers/text/text_vectorization).
+It can operate as a part of your main model so that the model is excluded from the core
+preprocessing logic. This greatly reduces the chances of training and serving skew.
+
+We first calculate the number of unique words present in the abstracts.
```
fchollet, line 240: If you want to index all words, you don't need to precompute that number. You can just pass `max_tokens=None`. However it's usually a good idea to limit `max_tokens` to ~20k for better efficiency.

examples/nlp/multi_label_classification.py
```diff
@@ -0,0 +1,382 @@
...
+            layers.Dense(512, activation="relu"),
+            layers.Dense(256, activation="relu"),
+            layers.Dense(len(mlb.classes_), activation="sigmoid"),
+        ]
+    )
+    return shallow_mlp_model
+
+
+"""
+Without the CPU placement, we run into:
```
fchollet, line 290: This is a known issue that we're currently fixing, so you can remove this comment (it will soon be outdated)

examples/nlp/multi_label_classification.py
```diff
@@ -0,0 +1,382 @@
...
+# `TextVectorization` needs to be adapted as per the vocabulary from our
+# training set.
+with tf.device("/CPU:0"):
+    text_vectorizer.adapt(train_dataset.map(lambda text, label: text))
+
+
+def make_model():
+    shallow_mlp_model = keras.Sequential(
+        [
+            text_vectorizer,
```
fchollet, line 280: Prefer `map()`ing the layer in the tf.data pipeline

fchollet (commented):
Thanks for the PR! This is a much needed topic :)

examples/nlp/multi_label_classification.py
```diff
@@ -0,0 +1,382 @@
...
+print(f"Number of rows in test set: {len(test_df)}")
+
+"""
+## Multi-label binarization
+
+Now we preprocess our labels using
+[`MultiLabelBinarizer`](http://scikit-learn.org/stable/modules/generated/sklearn.preprocessing.MultiLabelBinarizer.html). 
+"""
+
+mlb = MultiLabelBinarizer()
```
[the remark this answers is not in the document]
sayakpaul: Sure. So, here's what we tried:

```python
max_seqlen = 150
batch_size = 128
padding_token = "<pad>"
auto = tf.data.AUTOTUNE


def unify_text_length(text):
    # Split the given abstract and calculate its length.
    word_splits = tf.strings.split(text, sep=" ")
    sequence_length = tf.shape(word_splits)[0]
    
    # Calculate the padding amount.
    padding_amount = max_seqlen - sequence_length
    
    # Check if we need to pad or truncate.
    if padding_amount > 0:
        unified_text = tf.pad([text], [[0, padding_amount]], constant_values="<pad>")
        unified_text = tf.strings.reduce_join(unified_text, separator="")
    else:
        unified_text = tf.strings.reduce_join(word_splits[:max_seqlen], separator=" ")
    
    # The expansion is needed for subsequent vectorization.
    return tf.expand_dims(unified_text, -1)


def preprocess_label(label):
    return binarizer([label])


def preprocess_text_and_label(text, label):
    unified_text = unify_text_length(text)
    preprocessed_label = preprocess_label(label)


def make_dataset(dataframe, is_train=True):
    dataset = tf.data.Dataset.from_tensor_slices(
        (dataframe["summaries"].values, dataframe["terms"].values)
    )
    dataset = dataset.shuffle(batch_size * 10) if is_train else dataset
    dataset = dataset.map(preprocess_text_and_label, num_parallel_calls=auto).cache()
    return dataset.batch(batch_size).prefetch(auto)
```

where the `binarizer` is defined as follows:

```python
train_term_set = set()
for terms in dataframe["terms"]:
    for i in range(len(terms)):
        train_term_set.add(terms[i])

train_term_set = tf.convert_to_tensor(np.array(list(train_term_set)))

binarizer = layers.StringLookup(output_mode="multi_hot")
binarizer.adapt(train_term_set)
```

Now when we call `train_dataset = make_dataset(train_df, is_train=True)` we are met with:

```
ValueError: Failed to convert a NumPy array to a Tensor (Unsupported object type list).
```

This is why we opted for `MultiLabelBinarizer`. Happy to consider any better method you may suggest.

examples/nlp/multi_label_classification.py
sayakpaul: Oh, we wanted to let the model directly take raw strings hence this choice. We understand that it's totally possible to export a separate model including the `TextVectorization` layer on top of the trained model but when we wanted to show the users a different way. Is that okay?

examples/nlp/multi_label_classification.py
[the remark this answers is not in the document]
sayakpaul: Here's another approach we tried and it works:

```python
max_seqlen = 150
batch_size = 128
padding_token = "<pad>"
auto = tf.data.AUTOTUNE


def unify_text_length(text, label):
    # Split the given abstract and calculate its length.
    word_splits = tf.strings.split(text, sep=" ")
    sequence_length = tf.shape(word_splits)[0]
    
    # Calculate the padding amount.
    padding_amount = max_seqlen - sequence_length
    
    # Check if we need to pad or truncate.
    if padding_amount > 0:
        unified_text = tf.pad([text], [[0, padding_amount]], constant_values="<pad>")
        unified_text = tf.strings.reduce_join(unified_text, separator="")
    else:
        unified_text = tf.strings.reduce_join(word_splits[:max_seqlen], separator=" ")
    
    # The expansion is needed for subsequent vectorization.
    return tf.expand_dims(unified_text, -1), label


def preprocess_label(labels):
    preprocessed_labels = []
    for label in labels:
        preprocessed_label = binarizer([label]).numpy()
        preprocessed_labels.append(preprocessed_label)
    return np.array(preprocessed_labels)


def make_dataset(dataframe, is_train=True):
    preprocessed_labels = preprocess_label(dataframe["terms"].values)
    dataset = tf.data.Dataset.from_tensor_slices(
        (dataframe["summaries"].values, preprocessed_labels)
    )
    dataset = dataset.shuffle(batch_size * 10) if is_train else dataset
    dataset = dataset.map(unify_text_length, num_parallel_calls=auto).cache()
    return dataset.batch(batch_size).prefetch(auto)
```

But now, we are having to iterate through the entire dataset first to process the labels with the `binarizer` and then do the rest.

examples/nlp/multi_label_classification.py
fchollet: It's preferable to always train with a model where `TextVectorization` is in the data pipeline, to enable async preprocessing. Otherwise you can create a performance bottleneck on GPU/TPU

sayakpaul: Alright. 

When `map()`ing the `TextVectorization` layer what would be the preferred way? Should we do something like the following?

```
text_vectorizer.adapt(train_dataset.map(lambda text, label: text))
train_dataset = train_dataset.map(lambda text, label: (text_vectorizer(text), label))
validation_dataset = ...
test_dataset = ...
```

Note that `train_dataset` is already batched.

fchollet: Yes, this works, but add `num_parallel_calls`. Preprocessing layers work on both batched and unbatched data.

examples/nlp/multi_label_classification.py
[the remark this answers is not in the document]
sayakpaul: @fchollet a gentle ping for the above comment for your suggestion.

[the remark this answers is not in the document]
fchollet: The crux of the issue is converting the variable-length pandas column to a Tensor. @mattdangerw created this colab with a fix: https://colab.sandbox.google.com/drive/1VfqrPxm-aFBy6nqzFWSklqUW7naA7DpQ?resourcekey=0-FTknJ6xd6qhbstVVD0iaBA#scrollTo=1vgxbdwGf07E

The key part is this: you need to cast the variable-length pandas column to a ragged tensor.

```python
terms = tf.ragged.constant(train_df["terms"].values)
```

Since you also need to convert a multi-hot encoded output back to a list of labels: you'll need to write your own function for this, e.g.

```python
terms = tf.ragged.constant(train_df["terms"].values)
lookup = tf.keras.layers.StringLookup(output_mode='multi_hot')
lookup.adapt(terms)
vocab = lookup.get_vocabulary()

def invert_multi_hot(encoded_labels):
  """Reverse a single multi-hot encoded label to a tuple of vocab terms."""
  hot_indices = np.argwhere(encoded_labels == 1.0)[..., 0]
  return np.take(vocab, hot_indices)
```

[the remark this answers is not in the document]
sayakpaul: @fchollet @mattdangerw could you provide access to the Colab mentioned in the comment above?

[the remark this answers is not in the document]
mattdangerw: Hey! Sorry yeah not sure if I can share that version, but try this link.
https://colab.research.google.com/gist/mattdangerw/c2086b110a94918df1362f74697efba2/multi_label_trainer_tfidf-with-stringlookup.ipynb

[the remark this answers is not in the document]
sayakpaul: @mattdangerw a small nit was to actually use the `lookup` while creating the datasets (referring to the `make_dataset()` utility):

```
label_binarized = lookup(labels).numpy()
```

in place of 

```
label_binarized = mlb(labels)
```

But on the other hand, if you'd like to push your changes to the PR please do so and we will be more than happy to have you as a co-author. In fact, I think that should be the case since your changes and contributions to the example are quite non-trivial in my eyes.

sayakpaul:
@fchollet addressed the remaining items. 

We have provided @mattdangerw a separate acknowledgment for his generous help but as mentioned [here](https://github.com/keras-team/keras-io/pull/646#discussion_r717170740), we are happy to add him as a co-author. 

Also, you'd notice that we have provided the vocabulary size explicitly to the text vectorizer (and made an additional note on specifying `max_tokens=None`) since this resulted in a drastic performance improvement. 

Let us know if there's anything more you'd like us to address.

examples/nlp/multi_label_classification.py
```diff
@@ -0,0 +1,431 @@
...
+    else:
+        unified_text = tf.strings.reduce_join(word_splits[:max_seqlen], separator=" ")
+
+    # The expansion is needed for subsequent vectorization.
+    return tf.expand_dims(unified_text, -1), label
+
+
+def make_dataset(dataframe, is_train=True):
+    labels = tf.ragged.constant(dataframe["terms"].values)
+    label_binarized = lookup(labels).numpy()
```
mattdangerw, line 218: Do you actually need the numpy() call here? I think from_tensor_slices should handle the tensor output directly just fine.

examples/nlp/multi_label_classification.py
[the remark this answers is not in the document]
mattdangerw: Oops, yeah must have not saved with actually using the lookup there.

Just took a look at the code changes looks good to me, one nit I found. Thanks very much for updating and thanks for the acknowledgement on the end! No need for authorship, this was a small change to a big example.

Thanks!

examples/nlp/multi_label_classification.py
sayakpaul: Yeah, it should. But I find it a good practice to use uniform data types while calling `from_tensor_slices()`.

fchollet (approved):
Thanks for the updates! Everything is looking great. I pushed some small copyedits.

You can add the generated files 👍

sayakpaul:
Thank you @fchollet and @mattdangerw for all the help on this one. We learned a lot for sure :)

I just pushed the final files. Additionally, we wanted you to know, with this example, we are also releasing this Kaggle Dataset (that you might have noticed while reviewing the example): https://www.kaggle.com/spsayakpaul/arxiv-paper-abstracts. We hope it'd serve as a good benchmark for multi-label classification leading to interesting applications. 

We are currently working on an Apache Beam pipeline to collect an even bigger dataset and our progress along those lines have been steady so far.

fchollet:
> we are also releasing this Kaggle Dataset

Saw that in the example -- that's awesome! I'm sure it will be useful to the community. And best of luck with the larger dataset collection!

sayakpaul:
@fchollet we just updated the dataset: https://www.kaggle.com/spsayakpaul/arxiv-paper-abstracts. We also open-sourced our notebook used for collecting the dataset: https://github.com/soumik12345/multi-label-text-classification/blob/master/beam_arxiv_scrape.ipynb. We used Apache Beam and Dataflow.

PR #666, opened 2021-10-15
adding image_classification_with_convmixer.py

czy00000:
This example implements the ConvMixer model for image classification, and demonstrates it on the CIFAR-10 dataset.
The ConvMixer model is an extremely simple model that is similar in spirit to the Vit.

### DIFF
--- a/examples/vision/image_classification_with_convmixer.py
+++ b/examples/vision/image_classification_with_convmixer.py
@@ -1,0 +1,221 @@
+"""
+Title: Image classification with ConvMixer
+Author: [ZhiYong Chang](https://github.com/czy00000)
+Date created: 2021/10/14
+Last modified: 2021/10/14
+Description: Implementing the ConvMixer model for image classification.
+"""
+
+"""
+## Introduction
+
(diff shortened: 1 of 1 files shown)


fchollet (commented):
Thanks for the PR!

This example overlaps with another recent PR: #663. What do you think we should do? Possible outcomes are:

- If this example provides something specific (whether it's explanations, visualizations, features, etc) that differs from the other one and improves on it, we could merge the two examples into one that takes the best parts from each.
- Otherwise, we may simply merge the other example and close this one, since the other PR was a few days earlier.

Let me know!

czy00000:
@fchollet  Thank you for your reply! I just implemented the ConvMixer model. #663 gave some visual results, which should be better than my example.So you can close this PR, that is ok ：)

sayakpaul:
@czy00000 I am sorry if I jeopardized your chance to contribute. I understand that feeling of excitement when you implement something and want to share that with the broader community. I empathize with you. 

If, in the future, you want to contribute an example and think it might clash with mine you can always reach out to me here: spsayakpaul@gmail.com. I will also try to do the same.

PR #671, opened 2021-10-19
Add: Keras reproducibility guide in the call

sayakpaul:
@fchollet I have some preliminary work done in this: https://bit.ly/34V5ZNz. I wanted to extend it and include a call for contribution for this particular guide. 

I have some pointers in mind:

* Fixing seeds
* Ensuring development infrastructure unification
* [tf-determinism](https://github.com/NVIDIA/framework-determinism) for using deterministic CUDA kernels to trade away performance

With these, it's possible to attain at least some amount of reproducibility. But I am of course open to any suggestions you may have. I think it's time we had a guide on this topic.

### DIFF
--- a/call_for_contributions.md
+++ b/call_for_contributions.md
@@ -84,5 +84,9 @@ mixing numerical, categorical, and text features, doing feature engineering with
 
 ---
 
+## Keras reproducibility recipes
+
+---
+
 
 


sayakpaul:
Thank you @fchollet.

> A TF API for enabling fully-deterministic computation at the cost of a performance degradation, tf.config.experimental.enable_op_determinism()

Sounds really good. We might want to even include a friendly warning message on the performance degradation part.

PR #673, opened 2021-10-19
Add image classification with EANet.py

czy00000:
The EANet introduces a novel attention mechanism named external attention, based on two external, small, learnable, and shared memories, which can be implemented easily by simply using two cascaded linear layers and two normalization layers; it conveniently replaces self-attention in existing popular architectures.


innat:
@czy00000 Are you sure you double-check the architecture of EANet? [Here](https://github.com/MenghaoGuo/EANet/blob/d7ae481b2b68858de6c9e4826eb27130f5d8bdf8/model_torch.py#L184) is the official implementation. The implementation in the code example may have some issues. Could you please look into that?

PR #674, opened 2021-10-20
An example on MobileViT

sayakpaul:
This PR adds the [MobileViT model](https://arxiv.org/pdf/2110.02178.pdf), a mobile-friendly general-purpose architecture for various image recognition tasks such as classification, detection, and segmentation. 

Colab Notebook: https://drive.google.com/file/d/1_s1O8787UX4OKopj9DxvO9z67pAtMuMk/view?usp=sharing (originally run on a GCP VM machine to speed up training time).

### DIFF
--- a/examples/vision/ipynb/mobilevit.ipynb
+++ b/examples/vision/ipynb/mobilevit.ipynb
@@ -1,0 +1,539 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {
+    "colab_type": "text"
+   },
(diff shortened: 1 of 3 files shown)


sayannath:
@sayakpaul I guess it would be great if we make inferences of the test set on both save_model as well as the tflite model to get the accuracy as a benchmark.

Any thoughts on this? I would like to help if that can be done 🚀

sayakpaul:
@fchollet thanks for the review and I have addressed your comments.

PR #690, opened 2021-10-31
Improve image_captioning training speed by augmenting on GPU 

fedeallocati:
Further performance improvements can be achieved on this example by moving the image augmentations from the input pipeline to the GPU. 

The code changes are pretty minimal, since the example already has a custom `train_step` function.

Again I've checked for consistent improvements both locally and using Colab.

Before change (Colab GPU instance)
image

After change (Colab GPU instance)
image_captioning augment in gpu

Please let me know if there are any changes to be made!


fchollet (approved):
LGTM. That's surprising, I would not have expected such a large difference (especially since `num_parallel_calls` is correctly set).

fedeallocati:
From my experience, on this example the GPU is underutilized, so moving the preprocessing from CPU to the GPU is a win on both sides.

Thanks for the review!

PR #697, opened 2021-11-08
Added example of Barlow Twins

dewball345:
This is a Keras implementation of Barlow Twins(contrastive SSL with redundancy reduction techniques) using the CIFAR-10 dataset

* Demonstrates implementation of Barlow Twins 
* Contains a concise and clear explanation about the topic for people who don't know much about self-supervised learning

This method can reach about 63-64% validation accuracy with linear evaluation.

[COLAB LINK](https://colab.research.google.com/drive/12AALKqGSiW4JizTUOIcZhfwwcmnAa3h0?usp=sharing)


dewball345:
@haifeng-jin Just wanted to know if you could review this PR. I totally understand if you are busy, though.

haifeng-jin:
@dewball345 I will review it asap. Thanks

innat:
@dewball345 is it possible to share a colab link?

dewball345:
> @dewball345 I will review it asap. Thanks

Sure, thanks!

dewball345:
> @dewball345 is it possible to share a colab link?

Sure, https://colab.research.google.com/drive/12AALKqGSiW4JizTUOIcZhfwwcmnAa3h0?usp=sharing

examples/vision/barlow_twins.py
```diff
@@ -0,0 +1,990 @@
...
+tf.config.optimizer.set_jit(True)  # XLA optimization for faster performance
+
+"""
+# Load the CIFAR-10 dataset
+"""
+
+[
+    (train_features, train_labels),
+    (test_features, test_labels),
+] = tf.keras.datasets.cifar10.load_data()
```
innat, line 110: @dewball345 
This example contains many augmentation techniques. And so, I think we should use a more visually high-quality data set to realize the augmentation. For example, the output result of line 529 of the CIFAR dataset is hard to realize. Instead, I think, it would be better to use dataset like [oxford_flowers102](https://www.tensorflow.org/datasets/catalog/oxford_flowers102)

examples/vision/barlow_twins.py
```diff
@@ -0,0 +1,990 @@
...
+"""
+# Barlow Twins' Model Architecture
+The model consists of an
+*   Encoder network, which is a resnet-34
+*   Projector network, which creates the model embeddings.
+   * This consists of an MLP with 3 dense-batchnorm-relu layers
+"""
+
+
+class ResNet34:
```
innat, line 728: @dewball345 
I think it may not be the usual practice to define the model class this way (L728-L797). We should either use model subclassing API or functional API with a conventional modeling approach. So, here we may not need to define any model-class and simply state only the model functions.

examples/vision/barlow_twins.py
```diff
@@ -0,0 +1,990 @@
...
+
+    Returns:
+      returns a barlow twins model
+    """
+
+    # number of dense neurons in the projector
+    n_dense_neurons = 5000
+
+    # encoder network
+    resnet = ResNet34()()
```
innat, line 814: typo.

examples/vision/barlow_twins.py
```diff
@@ -0,0 +1,990 @@
...
+                for j in range(block_layers[i] - 1):
+                    x = self.identity_block(x, filter_size)
+        # Step 4 End Dense Network
+        x = tf.keras.layers.AveragePooling2D((2, 2), padding="same")(x)
+        x = tf.keras.layers.Flatten()(x)
+        model = tf.keras.models.Model(inputs=x_input, outputs=x, name="ResNet34")
+        return model
+
+
+def build_twin() -> tf.keras.Model:
```
innat, line 800: The model-class (`ResNet34`) and `build_twin` function can be done at a time. It'll optimize the code.

examples/vision/barlow_twins.py
dewball345: A problem with oxford_flowers102 is that the image sizes are not constant, also the dataset is much bigger I believe, so training time will be much larger. Also, it looks to me that CIFAR-10 is a pretty commonly used dataset for other keras.io examples(so consistency reasons as well).

dewball345: Thanks for the comments. A problem with oxford_flowers102 is that the image sizes are not consistent. Also, looks like the dataset has very few samples. The CIFAR-10 dataset is a little easier for me to install, and I think it is the baseline dataset for keras.io examples

examples/vision/barlow_twins.py
dewball345: Thanks @innat, will inherit the tf.keras.Model class. I believe it is better for it to be in a class because of the complexity of the model

examples/vision/barlow_twins.py
dewball345: I will try this out. One concern is that it combines the resnet architecture with the main model(which makes it a little overwhelming to be honest). I thought it would be good if i could separate them because of replacement(it is easy to replace resnet34 with resnet50 or resnet20 if they are separated). What do you think?

innat: Youo may also consider building the model within the `BarlowModel` class where we can import `ResNet50` from `keras.applications`.  Something like [this](https://colab.research.google.com/drive/1Wc56C-_4nftQV6YMDj-SdlbRgUnFKvin?usp=sharing).

examples/vision/barlow_twins.py
```diff
@@ -0,0 +1,990 @@
...
+        if i == 0:
+            x = dense(last_layer)
+        else:
+            x = dense(x)
+        x = tf.keras.layers.BatchNormalization(name=f"projector_bn_{i}")(x)
+        x = tf.keras.layers.ReLU(name=f"projector_relu_{i}")(x)
+
+    # final dense layer and a small reshaping for compatible output
+    x = tf.keras.layers.Dense(n_dense_neurons, name=f"projector_dense_{n_layers}")(x)
+    x = tf.keras.layers.Reshape((n_dense_neurons,))(x)
```
innat, line 830: This `Reshape` layer here may not be needed. The feature shape before it is already in `(None, 5000)`.

examples/vision/barlow_twins.py
dewball345: Hello, thanks for the reply. That was my original solution, though I found that a resnet34 gives significant boosts in accuracy(59% to 64%) and also considerable decrease in training time(2 hours to 1 hr 30 minutes)

examples/vision/barlow_twins.py
dewball345: Okay, i will check. I believe it was (None, 5000, 1) previously but I will check again.

examples/vision/barlow_twins.py
dewball345: Let me know though if there is a keras implementation of Resnet34

innat: Yes, there are actually, [here](https://github.com/qubvel/classification_models#architectures), unofficial but trusted. ([#15269](https://github.com/keras-team/keras/issues/15269))

dewball345: hmm, i see. looks like keras-contrib is deprecated, and tensorflow-addons is recommended(and i actually use it). Is your model implementation there?

examples/vision/barlow_twins.py
```diff
@@ -0,0 +1,990 @@
...
+*   *RandomToGrayscale*: randomly applies grayscale to image 20% of the time
+*   *RandomColorJitter*: randomly applies color jitter 80% of the time
+*   *RandomFlip*: randomly flips image horizontally 50% of the time
+*   *RandomResizedCrop*: randomly crops an image to a random size then resizes. This happens 100% of the time
+*   *RandomSolarize*: randomly applies solarization to an image 20% of the time
+*   *RandomBlur*: randomly blurs an image 20% of the time
+"""
+
+
+class Augmentation:
```
fchollet, line 135: Please structure these layers as preprocessing layers. You can check out the implementation of e.g. `layers.RandomContrast` for an example.

examples/vision/barlow_twins.py
```diff
@@ -0,0 +1,990 @@
...
+  * Off diagonals are squared.
+
+After this the two parts are summed together.
+
+
+
+"""
+
+
+class BarlowLoss(tf.keras.losses.Loss):
```
fchollet, line 573: Throughout the example, prefer using relative paths such as `keras.losses.Loss` or `layers.Dense`

examples/vision/barlow_twins.py
```diff
@@ -0,0 +1,990 @@
...
+Description: A keras implemenation of Barlow Twins(constrastive SSL with redundancy reduction).
+"""
+"""
+# Barlow Twins
+
+**Author:** Abhiraam Eranti<br>
+**Date created:** 11/6/21<br>
+**Last modified:** 11/6/21<br>
+**Description:** A keras implemenation of Barlow Twins(constrastive SSL with redundancy reduction).
+"""
```
haifeng-jin, line 15: duplicated header

examples/vision/barlow_twins.py
```diff
@@ -0,0 +1,990 @@
...
+
+"""
+Paper: [Barlow Twins: Self-Supervised Learning via Redundancy Reduction](https://arxiv.org/abs/2103.03230)
+
+Original Implementation: [facebookresearch/barlowtwins](https://github.com/facebookresearch/barlowtwins)
+
+Notebook by: Abhiraam Eranti
```
haifeng-jin, line 22: Author info is mentioned in the header.

examples/vision/barlow_twins.py
```diff
@@ -0,0 +1,990 @@
...
...m unlabeled data, and is often used when the data is corrupted or if there is very little of it. A practical use for SSL is to create intermediate embeddings that are learned from the data. These embeddings are based on the dataset itself, with similar images having similar embeddings, and vice versa. They are then attached to the rest of the model, which uses those embeddings as information and effectively learns and makes predictions properly. These embeddings, ideally, should contain as much information and insight about the data as possible, so that the model can make better predictions. However, a common problem that arises is that the model creates embeddings that are redundant. For example, if two images are similar, the model will create embeddings that effectively are just a string of 1’s, or some other value that contains repeating bits of information. This is no better than a one-hot encoding or just having one bit as the model’s representations; it defeats the purpose of the embeddings, as they do not learn as much about the dataset as possible. For other approaches, the solution to the problem was to carefully configure the model such that it tries not to be redundant.
```
haifeng-jin, line 24: Please reformat the long lines to shorter ones.
No more than 80 characters per line would be ideal.

examples/vision/barlow_twins.py
```diff
@@ -0,0 +1,990 @@
...
+*   Invariant
+*   Not redundant, and carry as much info about the dataset
+
+Also, it is simpler than other methods.
+
+This notebook was able to train a barlow twins model and reach approximately 64% validation accuracy
+"""
+
+"""
+<img src="https://i.imgur.com/G6LnEPT.png">
```
haifeng-jin, line 44: Is it possible to change it to markdown?

examples/vision/barlow_twins.py
```diff
@@ -0,0 +1,990 @@
...
+
+One disadvantage of Barlow Twins is that it is heavily dependent on augmentation, suffering major performance decreases in accuracy without them.
+
+TL, DR: Barlow twins creates representations that are
+*   Invariant
+*   Not redundant, and carry as much info about the dataset
+
+Also, it is simpler than other methods.
+
+This notebook was able to train a barlow twins model and reach approximately 64% validation accuracy
```
haifeng-jin, line 40: add a "." at the end of the line.

examples/vision/barlow_twins.py
```diff
@@ -0,0 +1,990 @@
...
+Taken from the original paper: [Barlow Twins: Self-Supervised Learning via Redundancy Reduction](https://arxiv.org/abs/2103.03230)
+"""
+
+"""
+# High-Level Theory
+
+"""
+
+"""
+The model takes two versions of the same image(with different augmentations) as input. Then it takes a prediction of each of them, creating representations. They are then used to make a cross-correlation matrix
```
haifeng-jin, line 57: add "."

examples/vision/barlow_twins.py
```diff
@@ -0,0 +1,990 @@
...
+(pred_1.T @ pred_2) / batch_size
+```
+
+The cross-correlation matrix measures the correlation between the output neurons in the two representations made by the model predictions of the two augmented versions of data. Ideally, a cross-correlation matrix should look like an identity matrix if the two images are the same.
+
+When this happens, it means that the representations:
+1.   Are invariant. The diagonal shows the correlation between each representation's neurons and its corresponding augmented one. Because the two versions come from the same image, the diagonal of the matrix should show that there is a strong correlation between them. If the images are different, there shouldn't really be a diagonal.
+2.   Do not show signs of redundancy. If the neurons show correlation with a non-diagonal neuron, it means that it is not correctly identifying similarities between the two augmented images. This means that it is redundant.
+
+Here is a good way of understanding in pseudocode(information from the original paper)
```
haifeng-jin, line 70: add ":" or "." at the end of the line.

examples/vision/barlow_twins.py
```diff
@@ -0,0 +1,990 @@
...
+  j is the index of the second representation's neuron
+```
+"""
+
+"""
+# Setup
+"""
+
+"""shell
+ pip install tensorflow-addons
```
haifeng-jin, line 88: remove the leading space

examples/vision/barlow_twins.py
```diff
@@ -0,0 +1,990 @@
...
+# Load the CIFAR-10 dataset
+"""
+
+[
+    (train_features, train_labels),
+    (test_features, test_labels),
+] = tf.keras.datasets.cifar10.load_data()
+
+"""
+#Necessary Hyperparameters
```
haifeng-jin, line 113: add space after "#"

examples/vision/barlow_twins.py
```diff
@@ -0,0 +1,990 @@
...
+
+
+bt_augmentor = RandomAugmentor(IMAGE_SIZE)
+
+"""
+# Dataset Creation Utility
+
+A class that creates the barlow twins' dataset.
+
+The dataset consists of two copies of each image, with each copy receiving different augmentations
```
haifeng-jin, line 417: add "." at the end of the line.

examples/vision/barlow_twins.py
```diff
@@ -0,0 +1,990 @@
...
+        a1 = self.augmented_version(ds)
+        a2 = self.augmented_version(ds)
+
+        return tf.data.Dataset.zip((a1, a2)).with_options(self.options)
+
+
+augment_versions = BTDatasetCreator(bt_augmentor)(train_features)
+
+"""
+View examples of dataset
```
haifeng-jin, line 495: add "."

examples/vision/barlow_twins.py
```diff
@@ -0,0 +1,990 @@
...
+
+plot_values(next(sample_augment_versions))
+
+"""
+# Pseudocode of loss and model
+ The following sections follow the original author's pseudocode containing both model and loss functions(see diagram below). Also contains a reference of variables used.
+"""
+
+"""
+<img src="https://i.imgur.com/Tlrootj.png">
```
haifeng-jin, line 537: change to markdown.

examples/vision/barlow_twins.py
```diff
@@ -0,0 +1,990 @@
...
+c: cross correlation matrix
+c_diff: diagonal portion of loss(invariance term)
+off_diag: off-diagonal portion of loss(redundancy reduction term)
+```
+"""
+
+"""
+# BarlowLoss: barlow twins model's loss function
+
+Barlow Twins uses the cross correlation matrix for its loss. There are two parts to the loss function
```
haifeng-jin, line 559: add ":" at the end of the line.

examples/vision/barlow_twins.py
```diff
@@ -0,0 +1,990 @@
...
+        loss = self.cross_corr_matrix_loss(c)
+        return loss
+
+
+"""
+# Barlow Twins' Model Architecture
+The model consists of an
+*   Encoder network, which is a resnet-34
+*   Projector network, which creates the model embeddings.
+   * This consists of an MLP with 3 dense-batchnorm-relu layers
```
haifeng-jin, line 724: make the "." at the end of the line consistent.

examples/vision/barlow_twins.py
```diff
@@ -0,0 +1,990 @@
...
+    x = tf.keras.layers.Reshape((n_dense_neurons,))(x)
+
+    model = tf.keras.Model(resnet.input, x)
+    return model
+
+
+"""
+# Training Loop Model
+
+See pseudocode for reference
```
haifeng-jin, line 839: add "."

examples/vision/barlow_twins.py
```diff
@@ -0,0 +1,990 @@
...
+
+        self.optimizer.apply_gradients(zip(grads_model, self.model.trainable_variables))
+        self.loss_tracker.update_state(loss)
+
+        return {"loss": self.loss_tracker.result()}
+
+
+"""
+# Model Training
+* Used the LAMB optimizer, instead of ADAM or SGD
```
haifeng-jin, line 905: add "."

examples/vision/barlow_twins.py
```diff
@@ -0,0 +1,990 @@
...
+        """
+        self.options = tf.data.Options()
+        self.options.threading.max_intra_op_parallelism = 1
+        self.seed = seed
+        self.augmentor = augmentor
+
+    def augmented_version(self, ds: list) -> tf.data.Dataset:
+        """augmented_version function.
+
+        Creates an augmented version of the dataset
```
haifeng-jin, line 453: descriptions should have "." at the end of the sentences. Please apply to all.

examples/vision/barlow_twins.py
```diff
@@ -0,0 +1,990 @@
...
+
+        Arguments:
+            ds: A list of images(CIFAR-10 brought from
+            tf.keras.load_data is a list)
```
haifeng-jin, line 457: This line (new line of the description of one argument) should have one more indent. Please apply to all.

examples/vision/barlow_twins.py
```diff
@@ -0,0 +1,990 @@
...
+
+        Returns:
+          returns an tf.data.Dataset with 1 version of augmented images
```
haifeng-jin, line 460: The indent here should be consistent with the Arguments. Also add "." by the end of the line. Capitalize the first letter of the sentence. Please apply to all.

haifeng-jin (commented):
Thanks for the PR!

examples/vision/barlow_twins.py
dewball345: hi, thanks for the comments. Just wanted to let you know that not all the augmentations in the paper are supported as preprocessing layers(RandomColorJitter, RandomBlur, RandomSolarize, RandomResizedCrop). BTW RandomResizedCrop is a bit different then just cropping+resizing, and there is a significant decrease in accuracy when just crop+resize. 

Barlow twins is heavily dependent on augmentation, so I definitely need all of them and can't just remove ones that i can't use. But i will try making custom layers and seeing how that goes

dewball345: BTW RandomResizedCrop is a bit different then just cropping+resizing, and there is a significant decrease in accuracy when just crop+resize, so i cant just use layers.RandomCrop and layers.Resize.

examples/vision/barlow_twins.py
dewball345: Hi, currently i'm having a little problem when converting the resnet to tf.keras model. 

I got the error 
```

    in user code:
    
        File "<ipython-input-11-fdae70ad456f>", line 48, in call  *
            x = keras.layers.Conv2D(64, kernel_size=7, strides=2, padding='same')(x)
        File "/usr/local/lib/python3.7/dist-packages/keras/utils/traceback_utils.py", line 67, in error_handler  **
    
        ValueError: tf.function only supports singleton tf.Variables created on the first call. Make sure the tf.Variable is only created once or created outside tf.function. See https://www.tensorflow.org/guide/function#creating_tfvariables for more information.
    
    
    Call arguments received:
      • inputs=tf.Tensor(shape=(512, 32, 32, 3), dtype=float32)
```

examples/vision/barlow_twins.py
dewball345: Yes, you are correct. I did not need the reshape layer.

dewball345:
@fchollet @haifeng-jin @innat Hello, I have made most of the updates that have been requested. However, I included a few questions for some of the suggestions, but I think I have fixed most of my issues. Please let me know if there is anything left for me to do.

examples/vision/barlow_twins.py
```diff
@@ -0,0 +1,1046 @@
...
+
+        if self.random_execute(0.8):
+            x = tf.image.random_brightness(x, 0.8)
+            x = tf.image.random_contrast(x, 0.4, 1.6)
+            x = tf.image.random_saturation(x, 0.4, 1.6)
+            x = tf.image.random_hue(x, 0.2)
+        return x
+
+
+class RandomFlip(Augmentation):
```
innat, line 266: keras provides some augmentation layers, [here](https://keras.io/api/layers/preprocessing_layers/image_augmentation/), sample flipping is one of them. I think we can use that instead. 

```python
tf.keras.layers.RandomFlip(
    mode="horizontal_and_vertical", seed=None, **kwargs
)
```

examples/vision/barlow_twins.py
```diff
@@ -0,0 +1,1046 @@
...
+            Returns a flipped version of the image 50% of the time
+              and the original image 50% of the time.
+        """
+
+        if self.random_execute(0.5):
+            x = tf.image.random_flip_left_right(x)
+        return x
+
+
+class RandomResizedCrop(Augmentation):
```
innat, line 296: And maybe, same as here too with [this ](https://keras.io/api/layers/preprocessing_layers/image_preprocessing/resizing/) and [this](https://keras.io/api/layers/preprocessing_layers/image_augmentation/random_crop/). 

```python

tf.keras.layers.RandomCrop(height, width, seed=None, **kwargs)
tf.keras.layers.Resizing(
    height, width, interpolation="bilinear", 
    crop_to_aspect_ratio=False, **kwargs
)
```

examples/vision/barlow_twins.py
```diff
@@ -0,0 +1,1046 @@
...
+              and the original image 80% of the time.
+        """
+
+        if self.random_execute(0.2):
+            s = np.random.random()
+            return tfa.image.gaussian_filter2d(image=x, sigma=s)
+        return x
+
+
+class RandomAugmentor:
```
innat, line 390: How about if we use keras layer instead. 

```python

class RandomAugmentor(keras.layers.Layer):
     def __init__():
          init all_augmentation_layers.

     def call(x):
          calling sequentially.
```

examples/vision/barlow_twins.py
```diff
@@ -0,0 +1,1046 @@
...
+"""shell
+!pip install tensorflow-addons
+"""
+
+import os
+
+os.environ["TF_GPU_THREAD_MODE"] = "gpu_private"  # for faster performance
+
+import tensorflow as tf  # framework
+import tensorflow.keras as keras  # for tf.keras
```
innat, line 122: I think it's better to import as 

```python

from tensorflow import keras

```

examples/vision/barlow_twins.py
```diff
@@ -0,0 +1,1046 @@
...
+# Setup
+"""
+
+"""shell
+!pip install tensorflow-addons
+"""
+
+import os
+
+os.environ["TF_GPU_THREAD_MODE"] = "gpu_private"  # for faster performance
```
innat, line 119: How does it improve performance?

examples/vision/barlow_twins.py
```diff
@@ -0,0 +1,1046 @@
...
+
+import tensorflow as tf  # framework
+import tensorflow.keras as keras  # for tf.keras
+import tensorflow_addons as tfa  # LAMB optimizer and gaussian_blur_2d function
+import numpy as np  # np.random.random
+import matplotlib.pyplot as plt  # graphs
+import datetime  # tensorboard logs naming
+
+tf.config.optimizer.set_jit(True)  # XLA optimization for faster performance
```
innat, line 128: All of my experiment until now with `jit = True` wasn't satisfactory. I'm curious to know, have you found it useful generally? I'll do some experiments with it.

examples/vision/barlow_twins.py
```diff
@@ -0,0 +1,1046 @@
...
+    saturation, and hue to an image. There is a 80% chance that an
+    image will be randomly color-jittered.
+
+    Methods:
+        __call__: method that color-jitters an image 80% of
+          the time.
+    """
+
+    @tf.function
+    def __call__(self, x: tf.Tensor) -> tf.Tensor:
```
innat, line 238: would it be `__call__` or only `call` ?

examples/vision/barlow_twins.py
```diff
@@ -0,0 +1,1046 @@
...
+    RandomFlip class. Randomly flips image horizontally. There is a 50%
+    chance that an image will be randomly flipped.
+
+    Methods:
+        __call__: Method that flips an image 50% of
+          the time.
+    """
+
+    @tf.function
+    def __call__(self, x: tf.Tensor) -> tf.Tensor:
```
innat, line 278: same, would it be `__call__` or only `call`?

examples/vision/barlow_twins.py
```diff
@@ -0,0 +1,1046 @@
...
+
+class RandomToGrayscale(Augmentation):
+    """RandomToGrayscale class.
+
+    RandomToGrayscale class. Randomly makes an image
+    grayscaled based on the random_execute method. There
+    is a 20% chance that an image will be grayscaled.
+
+    Methods:
+        __call__: method that grayscales an image 20% of
```
innat, line 202: `call` method.

examples/vision/barlow_twins.py
```diff
@@ -0,0 +1,1046 @@
...
+          the time.
+    """
+
+    @tf.function
+    def call(self, x: tf.Tensor) -> tf.Tensor:
+        """__call__ function.
```
innat, line 208: `call` method.

examples/vision/barlow_twins.py
```diff
@@ -0,0 +1,1046 @@
...
+
+class RandomResizedCrop(Augmentation):
+    """RandomResizedCrop class.
+
+    RandomResizedCrop class. Randomly crop an image to a random size,
+    then resize the image back to the original size.
+
+    Methods:
+        __call__: Method that does random resize crop to the image.
+    """
```
innat, line 304: `call` method.

examples/vision/barlow_twins.py
```diff
@@ -0,0 +1,1046 @@
...
+    Solarization is when pixels accidentally flip to an inverted state.
+
+    Methods:
+        __call__: Method that does random solarization 20% of the time.
+    """
+
+    @tf.function
+    def call(self, x: tf.Tensor) -> tf.Tensor:
+        """__call__ function.
+
```
innat, line 344: `call` method.

examples/vision/barlow_twins.py
```diff
@@ -0,0 +1,1046 @@
...
+    RandomBlur class. Randomly blurs an image.
+
+    Methods:
+        __call__: Method that does random blur 20% of the time.
+    """
+
+    @tf.function
+    def call(self, x: tf.Tensor) -> tf.Tensor:
+        """__call__ function.
+
```
innat, line 373: `call` method. (doc)

examples/vision/barlow_twins.py
innat: @dewball345 
Can you give a separate reproducible file?

examples/vision/barlow_twins.py
```diff
@@ -0,0 +1,1046 @@
...
+10% we get from random guessing.
+"""
+
+# Approx: 64% accuracy with this barlow twins model.
+
+xy_ds = (
+    tf.data.Dataset.from_tensor_slices((train_features, train_labels))
+    .shuffle(1000)
+    .batch(BATCH_SIZE, drop_remainder=True)
+    .prefetch(tf.data.experimental.AUTOTUNE)
```
innat, line 993: We can use 

``` python
tf.data.AUTOTUNE
```

examples/vision/barlow_twins.py
```diff
@@ -0,0 +1,1046 @@
...
+)
+
+test_ds = (
+    tf.data.Dataset.from_tensor_slices((test_features, test_labels))
+    .shuffle(1000)
+    .batch(BATCH_SIZE, drop_remainder=True)
+    .prefetch(tf.data.experimental.AUTOTUNE)
```
innat, line 1000: same. 

``` python
tf.data.AUTOTUNE
```

examples/vision/barlow_twins.py
dewball345: Maybe keras model would make more sense because the layers are being chained?

examples/vision/barlow_twins.py
dewball345: I believe it had some performance improvement saving a few seconds each epoch but I'll formally check.

examples/vision/barlow_twins.py
dewball345: It saved approx 30 seconds on the first epoch and 1-2 seconds on every other one. Approx 5-10 minutes saved in training time.

examples/vision/barlow_twins.py
dewball345: Saves approx 4 seconds per epoch. So saves 10-15 minutes of training time

examples/vision/barlow_twins.py
dewball345: thanks for the comments, I just did this

examples/vision/barlow_twins.py
dewball345: Thanks, I was able to use the Resizing layer but somehow couldn't use RandomCrop. There was an error about dtype mismatch which I couldn't solve.

examples/vision/barlow_twins.py
dewball345: Converted it into a keras.Model class

dewball345:
hello @innat, thank you for your comments. I had tried to use preprocessing layers instead of tf.image in the backend. However, when I did the training, the performance was incredibly slow.

with tf.image, the training time per epoch was only 29 seconds
but when replacing preprocessing layers for only random flip and random resize, the training per epoch rose to almost 250 seconds per epoch.

I believe something may be wrong with my code, so I will share a colab

[link](https://colab.research.google.com/drive/1x_-yWu-uWY6gGN28rGJ_sYmSWpJUv0NL?usp=sharing)

beresandras:
Hi @dewball345, I would like to add my two cents on the usage of image augmentation layers. I believe they are intended to be used as part of your model, on batched images, on GPU. You are currently doing the augmentations as part of your data pipeline, per image, on CPU. In that case I think you should stick to tf.image operations.

I am the author of the [SimCLR code example](https://keras.io/examples/vision/semisupervised_simclr/), and I took the other route, check it out if you are interested. I modified your Colab Notebook minimally (copying over the same augmentation layers from the SimCLR example), to show exactly what I mean. Check it out [here](https://colab.research.google.com/drive/10nWF7kmeFasXILmVcvnt8VH5sU60eKRE?usp=sharing). In my Colab session, the epoch time was around 110s both with tf.image ops and with the tf.keras layers. Keep in mind however, that using the current keras augmentation layers, you wont be able to replicate the pipeline of the paper as precisely as you do now.

dewball345:
Hi @beresandras , thanks for your insight about this. I will check out your notebook modification soon (keeps crashing on the device I am typing on). By the way, I believe the images are normalized but it occurs in the loss function.

dewball345:
@beresandras  Okay, i viewed your example, thank you for helping me on this. While it is an improvement from the 225 seconds, it still takes way too much time to train. 109 seconds per epoch is still more than 3x slower than 29 seconds. @innat I think i will structure the layers as preprocessing layers and use the subclassing API but all of them will have to be powered by tf.image

innat:
@dewball345 
> with tf.image, the training time per epoch was only 29 seconds
but when replacing preprocessing layers for only random flip and random resize, the training per epoch rose to almost 250 seconds per epoch.

When we use [`tf.image.random_flip_left_right`](https://www.tensorflow.org/api_docs/python/tf/image/random_flip_left_right), it flips the image horizontally by default, and when we use [`keras.layers.RandomFlip`](https://www.tensorflow.org/api_docs/python/tf/keras/layers/RandomFlip), it flips the image horizontally and vertically, both.  Also in the backend, the layer class is usually using the [`tf.image.*` modules](https://github.com/keras-team/keras/blob/2c48a3b38b6b6139be2da501982fd2f61d7d48fe/keras/layers/preprocessing/image_preprocessing.py#L384); same as [resizing layer](https://github.com/keras-team/keras/blob/2c48a3b38b6b6139be2da501982fd2f61d7d48fe/keras/layers/preprocessing/image_preprocessing.py#L102), However, I didn't benchmark the performance speed but `tf.image.*` should be computationally faster than layer for less checkpoint (`if-else` like in layer call).

dewball345:
Hi, i used flip left right because that was how the paper did it. However, i previously did flip top bottom and the performance was about the same as well

dewball345:
> @dewball345
> 
> > with tf.image, the training time per epoch was only 29 seconds
> > but when replacing preprocessing layers for only random flip and random resize, the training per epoch rose to almost 250 seconds per epoch.
> 
> When we use [`tf.image.random_flip_left_right`](https://www.tensorflow.org/api_docs/python/tf/image/random_flip_left_right), it flips the image horizontally by default, and when we use [`keras.layers.RandomFlip`](https://www.tensorflow.org/api_docs/python/tf/keras/layers/RandomFlip), it flips the image horizontally and vertically, both. Also in the backend, the layer class is usually using the [`tf.image.*` modules](https://github.com/keras-team/keras/blob/2c48a3b38b6b6139be2da501982fd2f61d7d48fe/keras/layers/preprocessing/image_preprocessing.py#L384); same as [resizing layer](https://github.com/keras-team/keras/blob/2c48a3b38b6b6139be2da501982fd2f61d7d48fe/keras/layers/preprocessing/image_preprocessing.py#L102), However, I didn't benchmark the performance speed but `tf.image.*` should be computationally faster than layer for less checkpoint (`if-else` like in layer call).

You know, this is my two cents of knowledge but I assume that part of what makes it slower is the stateless random augmentations; I previously used the newer stateless_random functions but they were very slow, so i switched back to the older ones.

fchollet:
> You know, this is my two cents of knowledge but I assume one part of what makes it slower is the stateless random augmentations; I previously used the newer stateless_random functions but they were very slow, so i switched back to the older ones.

Do you more detailed info about this?

dewball345:
@fchollet 

https://colab.research.google.com/drive/1aJuooY93eHw4v9lW20RcLCOrFgt1JrlY?usp=sharing

this is an old barlow twins example that I made. When switching to stateless augmentations, the time per epoch rises to 58 seconds while with the regular versions the time per epoch is just 27s.

I might have done something wrong, so bear with me.

examples/vision/barlow_twins.py
```diff
@@ -0,0 +1,1046 @@
+"""
+Title: Barlow Twins
+Author: [Abhiraam Eranti](https://github.com/dewball345)
+Date created: 11/4/21
+Last modified: 11/17/21
+Description: A keras implemenation of Barlow Twins(constrastive SSL with redundancy reduction).
```
haifeng-jin, line 6: Add a space before "(".

examples/vision/barlow_twins.py
```diff
@@ -0,0 +1,1046 @@
...
+"""
+"""
+Paper:
+ [Barlow Twins: Self-Supervised Learning via Redundancy Reduction](https://arxiv.org/abs/2103.03230)
+
+Original Implementation:
+ [facebookresearch/barlowtwins](https://github.com/facebookresearch/barlowtwins)
+
+Self-supervised learning is a relatively novel technique in which a model
```
haifeng-jin, line 15: Change "Self-supervised learning" to "Self-supervised learning  (SSL)". As SSL is used on line 17, we should mention it before using.

examples/vision/barlow_twins.py
```diff
@@ -0,0 +1,1046 @@
...
+embeddings that effectively are just a string of 1’s, or some other value that
+contains repeating bits of information. This is no better than a one-hot
+encoding or just having one bit as the model’s representations; it defeats the
+purpose of the embeddings, as they do not learn as much about the dataset as
+possible. For other approaches, the solution to the problem was to carefully
+configure the model such that it tries not to be redundant.
+
+
+Barlow twins is a new approach to this problem; While other solutions mainly
+tackle the first goal of invariance(similar images have similar embeddings),
```
haifeng-jin, line 35: space before "(".

examples/vision/barlow_twins.py
```diff
@@ -0,0 +1,1046 @@
...
+if there is very little of it. A practical use for SSL is to create
+intermediate embeddings that are learned from the data. These embeddings are
+based on the dataset itself, with similar images having similar embeddings, and
+vice versa. They are then attached to the rest of the model, which uses those
+embeddings as information and effectively learns and makes predictions properly.
+These embeddings, ideally, should contain as much information and insight about
+the data as possible, so that the model can make better predictions. However,
+a common problem that arises is that the model creates embeddings that are
+redundant. For example, if two images are similar, the model will create
+embeddings that effectively are just a string of 1’s, or some other value that
```
haifeng-jin, line 26: Seems we don't need the word "effectively" on this line?

examples/vision/barlow_twins.py
```diff
@@ -0,0 +1,1046 @@
...
+augmentation, suffering major performance decreases in accuracy without them.
+
+TL, DR: Barlow twins creates representations that are:
+*   Invariant.
+*   Not redundant, and carry as much info about the dataset.
+
+Also, it is simpler than other methods.
+
+This notebook was able to train a barlow twins model and reach approximately
+64% validation accuracy.
```
haifeng-jin, line 54: 64% on which dataset?

examples/vision/barlow_twins.py
haifeng-jin: I think this is not addressed yet. we should `from tensorflow import keras`.

examples/vision/barlow_twins.py
haifeng-jin: Seems you are still working on replacing it, right?

examples/vision/barlow_twins.py
```diff
@@ -0,0 +1,1046 @@
...
+
+        z_a_norm, z_b_norm = self.normalize(z_a), self.normalize(z_b)
+        c = self.cross_corr_matrix(z_a_norm, z_b_norm)
+        loss = self.cross_corr_matrix_loss(c)
+        return loss
+
+
+"""
+# Barlow Twins' Model Architecture
+The model consists of an:
```
haifeng-jin, line 764: Remove "an".

examples/vision/barlow_twins.py
```diff
@@ -0,0 +1,1046 @@
...
+   * This consists of an MLP with 3 dense-batchnorm-relu layers.
+"""
+
+"""
+Resnet encoder network implementation:
+"""
+
+
+class ResNet34:
+    """Resnet34 class.
```
haifeng-jin, line 776: A blank line follows this line.

examples/vision/barlow_twins.py
```diff
@@ -0,0 +1,1046 @@
...
+        return model
+
+
+"""
+Projector network:
+"""
+
+
+def build_twin() -> keras.Model:
+    """build_twin method
```
haifeng-jin, line 852: Add a period at the end of the line, and insert a blank line after this line.

examples/vision/barlow_twins.py
```diff
@@ -0,0 +1,1046 @@
...
+    Builds a barlow twins model consisting of an encoder(resnet-34)
+    and a projector, which generates embeddings for the images
```
haifeng-jin, line 854: Add a period at the end of the line, and insert a blank line after this line.

examples/vision/barlow_twins.py
```diff
@@ -0,0 +1,1046 @@
...
+    Returns:
+      Returns a barlow twins model
```
haifeng-jin, line 856: 4 spaces as the indent.

examples/vision/barlow_twins.py
```diff
@@ -0,0 +1,1046 @@
...
+        return [self.loss_tracker]
+
+    def train_step(self, batch: tf.Tensor) -> tf.Tensor:
+        """train_step method.
+
+        Do one train step. Make model predictions, find loss, pass loss to
+        optimizer, and make optimizer apply gradients.
+
+        Arguments:
+          batch: One batch of data to be given to the loss function.
```
haifeng-jin, line 928: 4 spaces as the indent.

examples/vision/barlow_twins.py
```diff
@@ -0,0 +1,1046 @@
...
+
+        Returns:
+          Returns a dictionary with the loss metric.
```
haifeng-jin, line 931: 4 spaces as the indent.

haifeng-jin (commented):
Thanks for addressing the previous comments! It is much better now.

examples/vision/barlow_twins.py
dewball345: yes, i have done this. Sorry it wasn't resolved.

examples/vision/barlow_twins.py
dewball345: I think I won't be able to replace it due to performance issues

dewball345:
Okay, added the fixes. Please review if anything else needs to be changed.

examples/vision/barlow_twins.py
```diff
@@ -0,0 +1,1101 @@
...
+* [Paper](https://arxiv.org/abs/2103.03230)
+* [Original Pytorch Implementation](https://github.com/facebookresearch/barlowtwins)
+* [Sayak Paul's
+Implementation](https://colab.research.google.com/github/sayakpaul/Barlow-Twins-TF/blob/main/Barlow_Twins.ipynb#scrollTo=GlWepkM8_prl).
+Implementation](https://colab.research.google.com/github/sayakpaul/Barlow-Twins-TF/blob/main/Barlow_Twins.ipynb#scrollTo=GlWepkM8_prl).
+* Thanks to Sayak Paul for his implementation. It helped me with debugging and
+comparisons of accuracy, loss.
+* [resnet34
+implementation](https://www.analyticsvidhya.com/blog/2021/08/how-to-code-your-resnet-from-scratch-in-tensorflow/#h2_2)
+implementation](https://www.analyticsvidhya.com/blog/2021/08/how-to-code-your-resnet-from-scratch-in-tensorflow/#h2_2)
```
haifeng-jin, line 1095: This line seems duplicated.

examples/vision/barlow_twins.py
```diff
@@ -0,0 +1,1101 @@
...
+data but a lot of unlabeled data.
+* You do barlow twins training on the unlabeled data, and then you do secondary training
+with the labeled data.
+
+# Helpful links
+* [Paper](https://arxiv.org/abs/2103.03230)
+* [Original Pytorch Implementation](https://github.com/facebookresearch/barlowtwins)
+* [Sayak Paul's
+Implementation](https://colab.research.google.com/github/sayakpaul/Barlow-Twins-TF/blob/main/Barlow_Twins.ipynb#scrollTo=GlWepkM8_prl).
+Implementation](https://colab.research.google.com/github/sayakpaul/Barlow-Twins-TF/blob/main/Barlow_Twins.ipynb#scrollTo=GlWepkM8_prl).
```
haifeng-jin, line 1090: This line seems duplicated.

examples/vision/barlow_twins.py
```diff
@@ -0,0 +1,1101 @@
...
+when it wasn't even trained with the labels. It can be used when you have little labeled
+data but a lot of unlabeled data.
+* You do barlow twins training on the unlabeled data, and then you do secondary training
+with the labeled data.
+
+# Helpful links
+* [Paper](https://arxiv.org/abs/2103.03230)
+* [Original Pytorch Implementation](https://github.com/facebookresearch/barlowtwins)
+* [Sayak Paul's
+Implementation](https://colab.research.google.com/github/sayakpaul/Barlow-Twins-TF/blob/main/Barlow_Twins.ipynb#scrollTo=GlWepkM8_prl).
```
haifeng-jin, line 1089: Keep the markdown hyperlinks on the same line. Put these 2 lines into the same line.

examples/vision/barlow_twins.py
```diff
@@ -0,0 +1,1101 @@
...
+Implementation](https://colab.research.google.com/github/sayakpaul/Barlow-Twins-TF/blob/main/Barlow_Twins.ipynb#scrollTo=GlWepkM8_prl).
+* Thanks to Sayak Paul for his implementation. It helped me with debugging and
+comparisons of accuracy, loss.
+* [resnet34
+implementation](https://www.analyticsvidhya.com/blog/2021/08/how-to-code-your-resnet-from-scratch-in-tensorflow/#h2_2)
```
haifeng-jin, line 1094: Keep the markdown hyperlinks on the same line. Put these 2 lines into the same line.

examples/vision/barlow_twins.py
```diff
@@ -0,0 +1,1101 @@
...
+Resnet encoder network implementation:
+"""
+
+
+class ResNet34:
+    """Resnet34 class.
+
+    Responsible for the Resnet 34 architecture.
+Modified from
+https://www.analyticsvidhya.com/blog/2021/08/how-to-code-your-resnet-from-scratch-in-tensorflow/#h2_2.
```
haifeng-jin, line 824: Indent needed.

examples/vision/barlow_twins.py
```diff
@@ -0,0 +1,1101 @@
...
+https://www.analyticsvidhya.com/blog/2021/08/how-to-code-your-resnet-from-scratch-in-tensorflow/#h2_2.
```
haifeng-jin, line 825: This line seems duplicated.

examples/vision/barlow_twins.py
```diff
@@ -0,0 +1,1101 @@
...
+Title: Barlow Twins
+Author: [Abhiraam Eranti](https://github.com/dewball345)
+Date created: 11/4/21
+Last modified: 11/17/21
+Description: A keras implemenation of Barlow Twins(constrastive SSL with redundancy reduction).
+"""
+"""
+Paper:
+[Barlow Twins: Self-Supervised Learning via Redundancy
+Reduction](https://arxiv.org/abs/2103.03230)
```
haifeng-jin, line 11: Keep the markdown hyperlinks on the same line. Put these 2 lines into the same line.

examples/vision/barlow_twins.py
```diff
@@ -0,0 +1,1101 @@
...
+
+
+
+
+
+"""
+
+"""
+Taken from the original paper: [Barlow Twins: Self-Supervised Learning via Redundancy
+Reduction](https://arxiv.org/abs/2103.03230)
```
haifeng-jin, line 70: Keep the markdown hyperlinks on the same line. line break right before  "[".

examples/vision/barlow_twins.py
haifeng-jin: This has not been addressed yet.

haifeng-jin (commented):
Thanks for the updates.

dewball345:
Fixed!

examples/vision/barlow_twins.py
haifeng-jin: This is not fixed yet.

examples/vision/barlow_twins.py
```diff
@@ -0,0 +1,1095 @@
...
+"""
+
+"""
+# High-Level Theory
+
+
+"""
+
+"""
+The model takes two versions of the same image(with different augmentations) as
```
haifeng-jin, line 79: insert space before "(".

examples/vision/barlow_twins.py
```diff
@@ -0,0 +1,1095 @@
...
+1.   Are invariant. The diagonal shows the correlation between each
+representation's neurons and its corresponding augmented one. Because the two
+versions come from the same image, the diagonal of the matrix should show that
+there is a strong correlation between them. If the images are different, there
+shouldn't really be a diagonal.
+2.   Do not show signs of redundancy. If the neurons show correlation with a
+non-diagonal neuron, it means that it is not correctly identifying similarities
+between the two augmented images. This means that it is redundant.
+
+Here is a good way of understanding in pseudocode(information from the original
```
haifeng-jin, line 103: insert space before "(".

examples/vision/barlow_twins.py
```diff
@@ -0,0 +1,1095 @@
...
+os.environ["TF_GPU_THREAD_MODE"] = "gpu_private"
+
+import tensorflow as tf  # framework
+from tensorflow import keras  # for tf.keras
+import tensorflow_addons as tfa  # LAMB optimizer and gaussian_blur_2d function
+import numpy as np  # np.random.random
+import matplotlib.pyplot as plt  # graphs
+import datetime  # tensorboard logs naming
+
+# XLA optimization for faster performance(up to 10-15 minutes total time saved)
```
haifeng-jin, line 138: insert space before "(".

examples/vision/barlow_twins.py
```diff
@@ -0,0 +1,1095 @@
...
+        self.seed = seed
+        self.augmentor = augmentor
+
+    def augmented_version(self, ds: list) -> tf.data.Dataset:
+        """augmented_version function.
+
+        Creates an augmented version of the dataset.
+
+        Arguments:
+            ds: A list of images(CIFAR-10 brought from
```
haifeng-jin, line 534: insert space before "(".

examples/vision/barlow_twins.py
```diff
@@ -0,0 +1,1095 @@
...
+
+    def __call__(self, ds: list) -> tf.data.Dataset:
+        """__call__ function.
+
+        Creates the barlow twins dataset. This dataset
+        contains two copies of the same images, except they
+        are augmented differently.
+
+        Arguments:
+            ds: A list of images(CIFAR-10 brought from
```
haifeng-jin, line 558: insert space before "(".

examples/vision/barlow_twins.py
```diff
@@ -0,0 +1,1095 @@
...
+
+    plt.show()
+
+
+plot_values(next(sample_augment_versions))
+
+"""
+# Pseudocode of loss and model
+The following sections follow the original author's pseudocode containing both model and
+loss functions(see diagram below). Also contains a reference of variables used.
```
haifeng-jin, line 613: insert space before "(".

examples/vision/barlow_twins.py
```diff
@@ -0,0 +1,1095 @@
...
+![pseudocode](https://i.imgur.com/Tlrootj.png)
+"""
+
+"""
+Reference:
+
+```
+y_a: first augmented version of original image.
+y_b: second augmented version of original image.
+z_a: model representation(embeddings) of y_a.
```
haifeng-jin, line 626: insert space before "(".

examples/vision/barlow_twins.py
```diff
@@ -0,0 +1,1095 @@
...
+z_b: model representation(embeddings) of y_b.
```
haifeng-jin, line 627: insert space before "(".

examples/vision/barlow_twins.py
```diff
@@ -0,0 +1,1095 @@
...
+z_a_norm: normalized z_a.
+z_b_norm: normalized z_b.
+c: cross correlation matrix.
+c_diff: diagonal portion of loss(invariance term).
```
haifeng-jin, line 631: insert space before "(".

examples/vision/barlow_twins.py
```diff
@@ -0,0 +1,1095 @@
...
+off_diag: off-diagonal portion of loss(redundancy reduction term).
```
haifeng-jin, line 632: insert space before "(".

examples/vision/barlow_twins.py
```diff
@@ -0,0 +1,1095 @@
...
+```
+"""
+
+"""
+# BarlowLoss: barlow twins model's loss function
+
+Barlow Twins uses the cross correlation matrix for its loss. There are two parts to the
+loss function:
+
+*   ***The invariance term***(diagonal). This part is used to make the diagonals of the
```
haifeng-jin, line 642: insert space before "(".

examples/vision/barlow_twins.py
```diff
@@ -0,0 +1,1095 @@
...
+matrix into 1s. When this is the case, the matrix shows that the images are
+correlated(same).
```
haifeng-jin, line 644: insert space before "(".

examples/vision/barlow_twins.py
```diff
@@ -0,0 +1,1095 @@
...
+# BarlowLoss: barlow twins model's loss function
+
+Barlow Twins uses the cross correlation matrix for its loss. There are two parts to the
+loss function:
+
+*   ***The invariance term***(diagonal). This part is used to make the diagonals of the
+matrix into 1s. When this is the case, the matrix shows that the images are
+correlated(same). 
+  * The loss function subtracts 1 from the diagonal and squares the values.
+*   ***The redundancy reduction term***(off-diagonal). Here, the barlow twins loss
```
haifeng-jin, line 646: insert space before "(".

examples/vision/barlow_twins.py
```diff
@@ -0,0 +1,1095 @@
...
+
+class BarlowLoss(keras.losses.Loss):
+    """BarlowLoss class.
+
+    BarlowLoss class. Creates a loss function based on the cross-correlation
+    matrix.
+
+    Attributes:
+        batch_size: the batch size of the dataset
+        lambda_amt: the value for lambda(used in cross_corr_matrix_loss)
```
haifeng-jin, line 667: insert space before "(".

examples/vision/barlow_twins.py
```diff
@@ -0,0 +1,1095 @@
...
+
+        Gets the loss based on the cross correlation matrix.
+        We want the diagonals to be 1's and everything else to be
+        zeros to show that the two augmented images are similar.
+
+        Loss function procedure:
+        take the diagonal of the cross-correlation matrix, subtract by 1,
+        and square that value so no negatives.
+
+        Take the off-diagonal of the cc-matrix(see get_off_diag()),
```
haifeng-jin, line 722: insert space before "(".

examples/vision/barlow_twins.py
```diff
@@ -0,0 +1,1095 @@
...
+        square those values to get rid of negatives and increase the value,
+        and multiply it by a lambda to weight it such that it is of equal
+        value to the optimizer as the diagonal(there are more values off-diag
```
haifeng-jin, line 725: insert space before "(".

examples/vision/barlow_twins.py
```diff
@@ -0,0 +1,1095 @@
...
+        Arguments:
+            c: A tf.tensor that represents the cross correlation
+              matrix
+
+        Returns:
+            Returns a tf.tensor which represents the cross correlation
+            matrix with its diagonals as zeros.
+        """
+
+        # subtracts diagonals by one and squares them(first part)
```
haifeng-jin, line 739: insert space before "(".

examples/vision/barlow_twins.py
```diff
@@ -0,0 +1,1095 @@
...
+            Returns a cross correlation matrix.
+        """
+        return (tf.transpose(z_a_norm) @ z_b_norm) / self.batch_size
+
+    def call(self, z_a: tf.Tensor, z_b: tf.Tensor) -> tf.Tensor:
+        """call method.
+
+        Makes the cross-correlation loss. Uses the CreateCrossCorr
+        class to make the cross corr matrix, then finds the loss and
+        returns it(see cross_corr_matrix_loss()).
```
haifeng-jin, line 789: insert space before "(".

examples/vision/barlow_twins.py
```diff
@@ -0,0 +1,1095 @@
...
+
+"""
+Projector network:
+"""
+
+
+def build_twin() -> keras.Model:
+    """build_twin method.
+
+    Builds a barlow twins model consisting of an encoder(resnet-34)
```
haifeng-jin, line 899: insert space before "(".

examples/vision/barlow_twins.py
haifeng-jin: This is not fixed yet.

examples/vision/barlow_twins.py
```diff
@@ -0,0 +1,1095 @@
...
+model.fit(xy_ds, epochs=35, validation_data=test_ds)
+
+"""
+# Conclusion
+*   Barlow Twins is a simple and concise method for contrastive and self-supervised
+learning.
+*   With this resnet-34 model architecture, we were able to reach 62-64% validation
+accuracy.
+
+# Use-Cases of Barlow-Twins(and contrastive learning in General)
```
haifeng-jin, line 1076: insert space before "(".

haifeng-jin (commented):
Thanks for the updates.

dewball345:
@haifeng-jin sorry for the redundant mistakes. I think I have fixed the issues addressed.

examples/vision/barlow_twins.py
haifeng-jin: This is not fixed yet.

examples/vision/barlow_twins.py
```diff
@@ -0,0 +1,1095 @@
...
+
+
+def build_twin() -> keras.Model:
+    """build_twin method.
+
+    Builds a barlow twins model consisting of an encoder (resnet-34)
+    and a projector, which generates embeddings for the images
+
+    Returns:
+        returns a barlow twins model
```
haifeng-jin, line 903: Add a period at the end of the line.

haifeng-jin (commented):
Thanks for the update. Everything else looks good to me. Just need to add the two periods.

dewball345:
Fixed these issues

haifeng-jin:
@dewball345 Seems it is failing the black style check. Would you try to run black locally and fix the issue? Thanks.

dewball345:
I have reformatted the file with black.

dewball345:
Looks like black works and all checks have passed. Would you like me to add the markdown files and notebook?

haifeng-jin:
@dewball345 Yes, please generate the md and ipynb files. Thanks.

dewball345:
I added all the files.

examples/vision/barlow_twins.py
innat: @dewball345 cc @haifeng-jin 
Is it addressed?

dewball345: I will send a reproducible file later, but for now, I think it's okay the way the resnet structure is. However, I do agree it would be better if the resnet was defined as a model and not as a model factory. 

Could you elaborate on what you meant by original paperwork? Do you mean the original paper on resnets?

innat: yes, the residual network paper.

haifeng-jin (approved):

dewball345:
@innat will likely add that as an update along with some other changes that
I want to add. Thank you so much for your help :)

dewball345:
@haifeng-jin Thank you so much for approving :)

examples/vision/barlow_twins.py
```diff
@@ -0,0 +1,1094 @@
...
+redundant. For example, if two images are similar, the model will create
+embeddings that are just a string of 1’s, or some other value that
+contains repeating bits of information. This is no better than a one-hot
+encoding or just having one bit as the model’s representations; it defeats the
+purpose of the embeddings, as they do not learn as much about the dataset as
+possible. For other approaches, the solution to the problem was to carefully
+configure the model such that it tries not to be redundant.
+
+
+Barlow twins is a new approach to this problem; While other solutions mainly
```
fchollet, line 34: Capitalize the method name

Also fix typos, e.g. no capital after `;`

examples/vision/barlow_twins.py
```diff
@@ -0,0 +1,1094 @@
...
+```
+(pred_1.T @ pred_2) / batch_size
+```
+
+The cross-correlation matrix measures the correlation between the output
+neurons in the two representations made by the model predictions of the two
+augmented versions of data. Ideally, a cross-correlation matrix should look
+like an identity matrix if the two images are the same.
+
+When this happens, it means that the representations:
```
fchollet, line 93: Add line break before list

examples/vision/barlow_twins.py
```diff
@@ -0,0 +1,1094 @@
...
+
+where:
+  c is the cross-correlation matrix
+  i is the index of one representation's neuron
+  j is the index of the second representation's neuron
+```
+"""
+
+"""
+# Setup
```
fchollet, line 118: `##`

examples/vision/barlow_twins.py
```diff
@@ -0,0 +1,1094 @@
...
+
+"""shell
+!pip install tensorflow-addons
+"""
+
+import os
+
+# slightly faster improvements, on the first epoch 30 second decrease and a 1-2 second
+# decrease in epoch time. Overall saves approx. 5 min of training time
+os.environ["TF_GPU_THREAD_MODE"] = "gpu_private"
```
fchollet, line 129: This deserves further explanations

examples/vision/barlow_twins.py
```diff
@@ -0,0 +1,1094 @@
...
+[
+    (train_features, train_labels),
+    (test_features, test_labels),
+] = keras.datasets.cifar10.load_data()
+
+train_features = train_features / 255.0
+test_features = test_features / 255.0
+
+"""
+# Necessary Hyperparameters
```
fchollet, line 154: `## Hyperparameters`

examples/vision/barlow_twins.py
```diff
@@ -0,0 +1,1094 @@
...
+# Width and height of image
+IMAGE_SIZE = 32
+
+"""
+# Augmentation Utilities
+The Barlow twins algorithm is heavily reliant on
+Augmentation. One unique feature of the method is that sometimes, augmentations
+probabilistically occur.
+ 
+**Augmentations**
```
fchollet, line 168: Fix list formatting. Please use `python autogen.py make` and `python autogen.py serve` to preview what the example looks like after being rendered to a HTML page

examples/vision/barlow_twins.py
```diff
@@ -0,0 +1,1094 @@
...
+    barlow twins' dataset.
+
+    Methods:
+        __init__: gets instance variables.
+        __call__: creates barlow dataset.
+        create_half: creates 1 half of the dataset.
+    """
+
+    def __init__(self, augmentor: RandomAugmentor, seed: int = 1024):
+        """__init__ function.
```
fchollet, line 512: These line descriptions provide no value. Yes, this is the `__init__` method, the reader of the code can see that.

Constructor arguments should be documented in the class docstring above: this docstring can be removed

examples/vision/barlow_twins.py
```diff
@@ -0,0 +1,1094 @@
...
+              augmented versions.
+            augmentor: augmentor used for augmentation.
+        """
+        self.options = tf.data.Options()
+        self.options.threading.max_intra_op_parallelism = 1
+        self.seed = seed
+        self.augmentor = augmentor
+
+    def augmented_version(self, ds: list) -> tf.data.Dataset:
+        """augmented_version function.
```
fchollet, line 529: Fix

examples/vision/barlow_twins.py
```diff
@@ -0,0 +1,1094 @@
...
+            tf.data.Dataset.from_tensor_slices(ds)
+            .shuffle(1000, seed=self.seed)
+            .map(self.augmentor, num_parallel_calls=tf.data.AUTOTUNE)
+            .batch(BATCH_SIZE, drop_remainder=True)
+            .prefetch(tf.data.AUTOTUNE)
+            .with_options(self.options)
+        )
+
+    def __call__(self, ds: list) -> tf.data.Dataset:
+        """__call__ function.
```
fchollet, line 551: Fix

examples/vision/barlow_twins.py
```diff
@@ -0,0 +1,1094 @@
...
+
+"""
+View examples of dataset.
+"""
+
+sample_augment_versions = iter(augment_versions)
+
+
+def plot_values(batch: tuple):
+    """plot_values function.
```
fchollet, line 580: Fix

examples/vision/barlow_twins.py
```diff
@@ -0,0 +1,1094 @@
...
+        x = self.random_solarize(x)
+
+        x = tf.clip_by_value(x, 0, 1)
+        return x
+
+
+bt_augmentor = RandomAugmentor(IMAGE_SIZE)
+
+"""
+# Dataset Creation Utility
```
fchollet, line 490: `## Data loading`

fchollet (changes requested):
I did another pass. The code looks great! The text copy (and docstrings) need some more work. Please address my comments (note that I haven't pointed out every single issue, so try to fix each category of issue globally, e.g. list formatting issues, `##`, etc.

You cam remove the autogenerated files for the time being.

ebursztein:
Looking at the code, it seems you could simplify it by using TF similarity barlow losses and the augmenters from it. They are currently in nightly and should be hitting prod this week -- see:  https://github.com/tensorflow/similarity/blob/development/tensorflow_similarity/losses/barlow.py

Same for the augmenters, you could just reuse the one in tf similarity.

dewball345:
I could add that as an update(as there are a series of other major
modifications I would like to make soon) but would like to initially
release the original version first. Thank you so much for mentioning this
library, I really wish I could have seen this before developing ;).

examples/vision/barlow_twins.py
dewball345: Hi, I'm getting an error `AttributeError: 'NoneType' object has no attribute 'keras_tuner'` even when I just clone the repository.

dewball345:
@fchollet I fixed the changes. Thank you for your patience, and sorry for the wait.

examples/vision/barlow_twins.py
fchollet: Perhaps you need to install KerasTuner (`pip install keras-tuner`). @haifeng-jin what do you think about this issue?

fchollet (approved):
Awesome, thanks for the update! LGTM. Are the md/ipynb files up to date?

dewball345:
No, will generate the files quickly today

examples/vision/barlow_twins.py
dewball345: I figured out the issue... I forgot to install packages in requirements.txt

dewball345:
@fchollet I have updated the source files.

fchollet:
@dewball345 thank you for the great contribution!

dewball345:
@fchollet is it normal that I cannot see my example on keras.io?

Issue #714, opened 2021-11-19
"Node Classification with Graph Neural Networks" example stopped working. 


fmerizzi:
**executing the code gnn_citations.py in colab produces an error in the training part, reported below.** 

`InvalidArgumentError                      Traceback (most recent call last)
<ipython-input-26-024a436bf839> in <module>()
      1 x_train = train_data.paper_id.to_numpy()
----> 2 history = run_experiment(gnn_model, x_train, y_train)

2 frames
/usr/local/lib/python3.7/dist-packages/tensorflow/python/eager/execute.py in quick_execute(op_name, num_outputs, inputs, attrs, ctx, name)
     57     ctx.ensure_initialized()
     58     tensors = pywrap_tfe.TFE_Py_Execute(ctx._handle, device_name, op_name,
---> 59                                         inputs, attrs, num_outputs)
     60   except core._NotOkStatusException as e:
     61     if name is not None:

InvalidArgumentError:  assertion failed: [Condition x == y did not hold element-wise:] [x (sparse_categorical_crossentropy/SparseSoftmaxCrossEntropyWithLogits/Shape_1:0) = ] [256 1] [y (sparse_categorical_crossentropy/SparseSoftmaxCrossEntropyWithLogits/strided_slice:0) = ] [256]
	 [[node sparse_categorical_crossentropy/SparseSoftmaxCrossEntropyWithLogits/assert_equal_1/Assert/Assert
 (defined at /usr/local/lib/python3.7/dist-packages/keras/backend.py:5114)
]] [Op:__inference_train_function_20561]

Errors may have originated from an input operation.
Input Source operations connected to node sparse_categorical_crossentropy/SparseSoftmaxCrossEntropyWithLogits/assert_equal_1/Assert/Assert:
In[0] sparse_categorical_crossentropy/SparseSoftmaxCrossEntropyWithLogits/assert_equal_1/All:	
In[1] sparse_categorical_crossentropy/SparseSoftmaxCrossEntropyWithLogits/assert_equal_1/Assert/Assert/data_0:	
In[2] sparse_categorical_crossentropy/SparseSoftmaxCrossEntropyWithLogits/assert_equal_1/Assert/Assert/data_1:	
In[3] sparse_categorical_crossentropy/SparseSoftmaxCrossEntropyWithLogits/Shape_1:	
In[4] sparse_categorical_crossentropy/SparseSoftmaxCrossEntropyWithLogits/assert_equal_1/Assert/Assert/data_3:	
[...]

PR #717, opened 2021-11-20
Fixing issue #714 by changing call() of class GNNNodeClassifier()

JanMarcelKezmann:
Fixing bug in Issue #714 by changing single line in call() function of the GNNNodeClassifier() class.


fchollet (approved):
LGTM, thanks for the PR

PR #722, opened 2021-11-25
Update call_for_contributions.md to include MoE

sayakpaul:
@ksalama what are your thoughts on adding an example on the call added in this PR?


fchollet (approved):
Excellent!

PR #744, opened 2021-12-16

RenuPatelGoogle:
Updated 'tf.keras.preprocessing.image.array_to_img' to 'keras.utils.array_to_img' in 'End-to-end example: a GAN training loop from scratch' and ran the program successfully after changes.


fchollet (approved):
Thanks for the PR!

PR #752, opened 2021-12-26
add: an example on GauGAN

sayakpaul:
This example implements the [GauGAN](https://arxiv.org/abs/1903.07291) model for conditional realistic image generation. Following the success of **GauGAN2** we believe this example will be helpful for the community dive deeper into generative deep learning. 

A Colab Notebook for the example is available here: https://colab.research.google.com/drive/18Ar6z6N78nECcm0ftfnJQhYj6jdmxhiy?usp=sharing. 

Cc: @soumik12345


fchollet:
Thank you for the PR! Love the topic. I'm currently travelling, but I'll be able to review in 2-3 days.

sayakpaul:
No worries, @fchollet. We wanted to end the year with a fun example but we definitely understand your situation.

innat:
@sayakpaul 
I think `def build_generator` has an `Input` layer class that didn't import (in the above colab and also in the PR files.)

```python
def build_generator(mask_shape, latent_dim=256):
    latent = Input(shape=(latent_dim))
    mask = Input(shape=mask_shape)

````

**Comment by soumik12345** [comment]
@innat Thanks for pointing out the bug. I have fixed it.

File: examples/generative/gaugan.py [hunk 1]
```diff
@@ -0,0 +1,792 @@
...
+N_CLASS = 12
+AUTOTUNE = tf.data.AUTOTUNE
+
+
+def load(image_files, batch_size, is_train=True):
+    def _random_crop(
+        image,
+        segmentation_map,
+        labels,
+        crop_size=tf.convert_to_tensor((IMG_HEIGHT, IMG_WIDTH)),
```
fchollet, line 127: Avoid calling a function in a signature definition, so move the `convert_to_tensor` inside the function body

examples/generative/gaugan.py
```diff
@@ -0,0 +1,792 @@
...
+    plt.imshow((real_image + 1) / 2)
+    plt.show()
+
+"""
+Note that in the rest of this example, we have used a couple of figures from the
+[original GauGAN paper](https://arxiv.org/abs/1903.07291) for convenience. 
+"""
+
+"""
+# Custom layers
```
fchollet, line 202: This should be `##`

examples/generative/gaugan.py
```diff
@@ -0,0 +1,792 @@
...
+convolved to produce the modulation parameters `γ` and `β`. Unlike prior conditional
+normalization methods, `γ` and `β` are not vectors, but tensors with spatial dimensions.
+The produced `γ` and `β` are multiplied and added to the normalized activation
+element-wise. As the modulation parameters are adaptive to the input segmentation mask,
+the proposed SPADE is better suited for semantic image synthesis.
+"""
+
+
+class SPADE(layers.Layer):
+    def __init__(self, filters, epsilon=1e-5):
```
fchollet, line 235: In general in custom layers, accept `**kwargs` in the signature and pass it to `super().__init__(**kwargs)`, which enables layer naming, etc.

examples/generative/gaugan.py
```diff
@@ -0,0 +1,792 @@
...
+        super(SPADE, self).__init__()
```
fchollet, line 236: Python 3 style: just `super().__init__`

examples/generative/gaugan.py
```diff
@@ -0,0 +1,792 @@
...
+        beta = self.conv_beta(x)
+        mean, var = tf.nn.moments(input_tensor, axes=(0, 1, 2), keepdims=True)
+        std = tf.sqrt(var + self.epsilon)
+        normalized = (input_tensor - mean) / std
+        output = gamma * normalized + beta
+        return output
+
+
+class ResBlock(layers.Layer):
+    def __init__(self, filters):
```
fchollet, line 258: Likewise for the other layers

examples/generative/gaugan.py
```diff
@@ -0,0 +1,792 @@
...
+            channels,
+            kernels,
+            strides=strides,
+            padding="same",
+            use_bias=False,
+            kernel_initializer=keras.initializers.GlorotNormal(),
+        )
+    )
+    if apply_norm:
+        block.add(tfa.layers.InstanceNormalization())
```
fchollet, line 329: (Note: we'll add this one to core Keras soon)

examples/generative/gaugan.py
```diff
@@ -0,0 +1,792 @@
...
+def generator_loss(y):
+    return -tf.reduce_mean(y)
+
+
+def kl_divergence_loss(mean, variance):
+    return -0.5 * tf.reduce_sum(1 + variance - tf.square(mean) - tf.exp(variance))
+
+
+class FeatureMatchingLoss(keras.losses.Loss):
+    def __init__(self, *args, **kwargs):
```
fchollet, line 446: `*args` should not be needed, no?

examples/generative/gaugan.py
```diff
@@ -0,0 +1,792 @@
...
+        real_features = self.vgg_model(y_true)
+        fake_features = self.vgg_model(y_pred)
+        loss = 0
+        for i in range(len(real_features)):
+            loss += self.weights[i] * self.mae(real_features[i], fake_features[i])
+        return loss
+
+
+class DiscriminatorLoss(keras.losses.Loss):
+    def __init__(self, *args, **kwargs):
```
fchollet, line 485: Likwise

examples/generative/gaugan.py
```diff
@@ -0,0 +1,792 @@
...
+    def __init__(
+        self,
+        image_size,
+        n_classes,
+        batch_size,
+        latent_dim,
+        feature_loss_coeff=10,
+        vgg_feature_loss_coeff=0.1,
+        kl_divergence_loss_coeff=0.1,
+        *args,
```
fchollet, line 551: `*args` not needed

examples/generative/gaugan.py
```diff
@@ -0,0 +1,792 @@
...
+Finally, we put everything together inside a subclassed model (from `tf.keras.Model`)
+overriding its `train_step()` method. 
+"""
+
+
+class GauGAN(keras.Model):
+    def __init__(
+        self,
+        image_size,
+        n_classes,
```
fchollet, line 545: `num_`

examples/generative/gaugan.py
```diff
@@ -0,0 +1,792 @@
...
+        generated_image = self.generator([latent_input, mask_input])
+        discriminator_output = self.discriminator([image_input, generated_image])
+        patch_size = discriminator_output[-1].shape[1]
+        combined_model = keras.Model(
+            [latent_input, mask_input, image_input],
+            [discriminator_output, generated_image],
+        )
+        return patch_size, combined_model
+
+    def compile(self, gen_lr=1e-4, disc_lr=4e-4, *args, **kwargs):
```
fchollet, line 593: `*args` not needed

examples/generative/gaugan.py
```diff
@@ -0,0 +1,792 @@
...
+        mean, variance = self.encoder(image)
+        latent_vector = self.sampler([mean, variance])
+        discriminator_loss = self.train_discriminator(
+            latent_vector, segmentation_map, image, labels
+        )
+        (generator_loss, feature_loss, vgg_loss, kl_loss) = self.train_generator(
+            latent_vector, segmentation_map, labels, image, mean, variance
+        )
+        return {
+            "discriminator_loss": discriminator_loss,
```
fchollet, line 662: This will only display the loss value for the latest batch rather than the average over the batches of the current epoch. To do the latter, use a `Mean()` metric tracker listed in `self.metrics` and `return metric.result() for metric in self.metrics` after updating the trackers above

examples/generative/gaugan.py
```diff
@@ -0,0 +1,792 @@
...
+        g_loss = generator_loss(pred)
+        kl_loss = self.kl_divergence_loss_coeff * kl_divergence_loss(mean, variance)
+        vgg_loss = self.vgg_feature_loss_coeff * self.vgg_loss(image, fake_image)
+        feature_loss = self.feature_loss_coeff * self.feature_matching_loss(
+            real_d_output, fake_d_output
+        )
+        total_generator_loss = g_loss + kl_loss + vgg_loss + feature_loss
+
+        return {
+            "discriminator_loss": total_discriminator_loss,
```
fchollet, line 700: Same

fchollet (commented):
Very cool implementation! Very clean and readable. I learned a lot about GauGAN 👍

sayakpaul:
@fchollet thank you for your valuable comments and glad that you liked the implementation. 

We have addressed the comments. 

Also, happy new year!

sayakpaul:
Additionally, we created this GIF for aiding the example:

animation

fchollet (approved):
Wonderful!

I added some copyedits. You can now generate the ipynb and md files. Thank you! 👍

sayakpaul:
@fchollet thank you. We had to increase the LOC to make the generation process work. Over to you now.

fchollet:
> I was testing the example locally, the title seems to be appearing two times:

Typically that's because you didn't use `autogen.py` to generate the files and you left a blank line at the top of your `.md` file. The keras.io page generation framework will add the title line automatically if the first line doesn't start with `#`.

> The nested bullet points seem to be broken:

Nested lists are not standard markdown unfortunately. To support nested lists, we need to include an extension that handles it. If you can figure out which one, it would go here: https://github.com/keras-team/keras-io/blob/master/scripts/autogen.py#L695

I see we already have `truly-sane-lists`, maybe it supports it but with a different syntax.

> There is a :::python identifier in the cells:

The shell code cells should not contain `!`. It gets added automatically.

sayakpaul:
@fchollet just readded the files.

fchollet (approved):
Thank you for the wonderful contribution! 👍

PR #754, opened 2021-12-28
update mpnn tutorial

akensert:
Main/notable updates:
1. `unsorted_segment_sum` now used instead of `segment_sum` to deal with edge cases (where the last node has no edges). So the implementation is now more robust.
2. Trainable variables are given names to allow saving the model (an issue brought up before)
3. Functional model simplified by putting partition+padding and average pooling inside the transformer encoder (now named `TransformerEncoderReadout`).
4. Text/math notation explaining the 'message passing' somewhat simplified for easier reading.
5. Abbreviation ("MPNN") added to title.
6. URL had changed for MoleculeNet website (referenced in the tutorial). Now updated accordingly.
7. "`atom_partition_indices`" changed to "`molecule_indicator`" (makes more sense).
8. Conclusions slightly modified.

### DIFF
--- a/examples/graph/mpnn-molecular-graphs.py
+++ b/examples/graph/mpnn-molecular-graphs.py
@@ -1,8 +1,8 @@
 """
-Title: Message-passing neural network for molecular property prediction
+Title: Message-passing neural network (MPNN) for molecular property prediction
 Author: [akensert](http://github.com/akensert)
 Date created: 2021/08/16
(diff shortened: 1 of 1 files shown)


fchollet (approved):
Thanks for the update, LGTM!

akensert:
@fchollet generated files were not included. I'm creating a new pull request with the generated files.

PR #767, opened 2022-01-09
Edits to the GauGAN example

sayakpaul:
The main change introduced in this PR is the `load()` function.


fchollet (approved):
Thanks for the PR!

PR #770, opened 2022-01-10
Video Classification with TimeDistributed Layer (w/o GPU)

SujoyKG:
New PR


fchollet (commented):
Thanks for the PR!

Is there a realistic video classification dataset (e.g. action classification) that you could use? I don't think this model architecture would perform well on a real dataset. In addition, the architecture used seems fundamentally expensive (e.g. use of conv2d instead of separable conv, use of LSTM) which contradicts the claim that it is optimized for CPU. At equal parameter count, much leaner architectures are possible.

I would recommend reworking this into a more conventional video classification example. Please start by finding an appropriate dataset, then we can discuss what architecture we should use and whether to use a pretrained base model.

SujoyKG:
> Thanks for the PR!
> 
> Is there a realistic video classification dataset (e.g. action classification) that you could use? I don't think this model architecture would perform well on a real dataset. In addition, the architecture used seems fundamentally expensive (e.g. use of conv2d instead of separable conv, use of LSTM) which contradicts the claim that it is optimized for CPU. At equal parameter count, much leaner architectures are possible.
> 
> I would recommend reworking this into a more conventional video classification example. Please start by finding an appropriate dataset, then we can discuss what architecture we should use and whether to use a pretrained base model.

Thanks for the reply. Please note what I am trying to express through this example. The dataset generated is optimized for quick checking of any Video Classification Model; the model is not optimized for CPU, as you have asked. I am more focused to the synthetic data generation rather than model building. Before feeding the actual data (real data) to the model, this generated data could get feeded to check the model at high level.

SujoyKG:
Would you like to get it added to "Quick Keras Recipes"?

SujoyKG:
Hi Team,

I really want you to go through the approach & code again. I believe it is
valuable & could be accepted.

Thanks,


-- Sent from mobile

PR #781, opened 2022-01-19
TabTransformer - Fix figure URL and keras model plotting

ksalama:
@fchollet kindly review this PR that fixes the URL for the TabTransformer figure, as well as plotting the Keras models in the makrdown file.


fchollet (approved):
Thank you, LGTM

PR #788, opened 2022-01-29
Update applications table for completeness and readability

ageron:
Full list of changes:

* Completes the _Top-1_ and _Top-5_ accuracy columns for EfficientNet models, based on [Table 2 from the paper](https://arxiv.org/pdf/1905.11946.pdf).
* Completes and updates the _Depth_ ​column to count only layers with parameters, as it is the most common convention. For example, VGG16 and VGG19 are named this way because they have 16 and 19 layers respectively, excluding parameter-free layers.
* Adds a short sentence to explain the _Depth_ column.
* Shortens the _Parameters_ column by showing millions, instead of exact count.
* Updates the _Top-1_ and _Top-5_ columns to show percentages, with 1 decimal point.
* Rounds the inference time columns to 1 decimal point instead of 2.
* Reorders the ResNet models so that v1 and v2 models are contiguous (makes the comparison easier).


fchollet (approved):
Thanks for the PR! Very helpful.

PR #789, opened 2022-01-29
Fix VAE randomness

tmiv:
The reparameterization trick did not perturb the gaussian multiplier during training. Now it does.

Introducing a new parameter 𝜖 allows us to reparameterize 𝑧 in a way that allows backprop to flow through the deterministic nodes. 𝜖 should be a source of randomness.

### DIFF
--- a/examples/generative/vae.py
+++ b/examples/generative/vae.py
@@ -19,15 +19,20 @@ from tensorflow.keras import layers
 ## Create a sampling layer
 """
 
-
 class Sampling(layers.Layer):
     """Uses (z_mean, z_log_var) to sample z, the vector encoding a digit."""
+    def __init__(self, epsilon=1.0, **kwargs):
+        super(Sampling, self).__init__(**kwargs)
+        self.epsilon = epsilon
+        
+    def get_config(self):
+        return {"epsilon": self.epsilon}
 
     def call(self, inputs):
         z_mean, z_log_var = inputs
         batch = tf.shape(z_mean)[0]
         dim = tf.shape(z_mean)[1]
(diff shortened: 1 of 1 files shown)


tmiv:
This merge request fixes the issue #790

fchollet:
With `epsilon = 1.0`, the line `random_normal(shape=(batch, dim))` is equivalent to `GaussianNoise(1.0)(tf.zeros((batch, dim)))`, except the latter is harder to read/follow. What's the purpose of this change?

For the record, `GaussianNoise(1.0)(x)` is equivalent to `x + random_normal()`, and is in fact implemented like this (except using V2-style stateless RNG instead of the old `random_normal`): https://github.com/keras-team/keras/blob/b96518a22bfd92a29811e507dec0b34248a8a3f5/keras/layers/regularization/gaussian_noise.py

tmiv:
Hello @fchollet, 
Thank you for your response. I can't yet say why the two implementations are not equivalent. As evidence that they are: In the colab for this example (https://colab.research.google.com/github/keras-team/keras-io/blob/master/examples/generative/ipynb/vae.ipynb) if you replace only the single epsilon assignment line:
image

 with 
image


You will get different outputs for the losses. Particularly the kl_losses are an order of magnitude less in the GaussianNoise implementation.

Original:
image

Gaussian Noise:
image


I will further investigate the difference as to why `epsilon = tf.keras.backend.random_normal(shape=(batch, dim))` is not equivalent to `epsilon = keras.layers.GaussianNoise(1.0)(tf.zeros((batch,dim))) `

PR #811, opened 2022-02-24
Add: A minimal implementation of ShiftViT

ariG23498:
This is a minimal implementation of [ShiftViT](https://arxiv.org/abs/2201.10801). In this paper the authors have proposed to replace the attention module of a ViT with a **NO PARAMETER** shifting operation. In this example we have tried aligning the implementation as closely as we could with the [official implementation](https://github.com/microsoft/SPACH/blob/main/models/shiftvit.py) in PyTorch. We hope this will benefit the keras community.

[Colab Notebook](https://colab.research.google.com/drive/11XA4oH8Lf7DSIbl4xMCRVd_jHOOTSvhU?usp=sharing)

CC: @ritwikraha


examples/vision/shiftvit.py
```diff
@@ -0,0 +1,871 @@
...
+            {
+                "data_augmentation": self.data_augmentation,
+                "patch_projection": self.patch_projection,
+                "stages": self.stages,
+                "global_avg_pool": self.global_avg_pool,
+            }
+        )
+        return config
+
+    def _calculate_loss(self, inputs, training=False):
```
fchollet, line 641: Use `data` rather than `inputs` (same for `test_step`)

examples/vision/shiftvit.py
```diff
@@ -0,0 +1,871 @@
...
+        self.warmup_steps = warmup_steps
+        self.total_steps = total_steps
+        self.pi = tf.constant(np.pi)
+
+    def __call__(self, step):
+        # Check whether the total number of steps is larger than the warmup
+        # steps. If not, then throw a value error.
+        if self.total_steps < self.warmup_steps:
+            raise ValueError(
+                "Total number of steps must be larger or equal to warmup steps."
```
fchollet, line 747: Printing both values will make this a more helpful error message.

fchollet (commented):
Thanks for the PR! 👍 

Please format the code using `black`.

ariG23498:
Thank you for the review @fchollet 

We have made the necessary changes!

examples/vision/shiftvit.py
```diff
@@ -0,0 +1,866 @@
...
+out[:, g * 0:g * 1, :, :-1] = x[:, g * 0:g * 1, :, 1:]  # shift left
+out[:, g * 1:g * 2, :, 1:] = x[:, g * 1:g * 2, :, :-1]  # shift right
+out[:, g * 2:g * 3, :-1, :] = x[:, g * 2:g * 3, 1:, :]  # shift up
+out[:, g * 3:g * 4, 1:, :] = x[:, g * 3:g * 4, :-1, :]  # shift down
+
+out[:, g * 4:, :, :] = x[:, g * 4:, :, :]  # no shift
+```
+
+In TensorFlow it would be infeasible for us to assign shifted channels to a tensor in the
+middle of a training process. This is why we have resorted to the following procedure:
```
fchollet: There are several things that could be done to implement just this, but concatenation works fine as well.

examples/vision/shiftvit.py
```diff
@@ -0,0 +1,866 @@
...
+        # Apply the patch merging algorithm on the feature maps
+        x = self.layer_norm(x)
+        x = self.reduction(x)
+        return x
+
+
+"""
+#### Stacked Shift Blocks
+
+Each stage will have a different number of stacking as suggested in the paper. This is a
```
fchollet: What does this sentence mean?

fchollet (commented):
Thanks for the updates! I pushed some copyedits.

examples/vision/shiftvit.py
```diff
@@ -0,0 +1,866 @@
...
+SEED = 42
+keras.utils.set_random_seed(SEED)
+
+"""
+## Hyperparameters
+
+These are the hyperparameters that we have chosen for the experiment. Please feel free to
+play around with them.
+
+_Note_: We do not use 4 stages as the input dimension is too small to support 4 stages
```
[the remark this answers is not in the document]
ariG23498: This is a mistake on my part. We were getting dimension error with 4 stages. We were under the impression that the dimensions of the feature map after the final stage was too small for the shifitng operation to work! I have tried with 4 stages and different setting, it works. I will be removing this note.

examples/vision/shiftvit.py
```diff
@@ -0,0 +1,866 @@
...
+3. MLP Layer
+"""
+
+"""
+#### MLP Layer
+
+MLP network is a Feed Forward Network. By default, the MLP is implemented by
+`keras.layers.Dense`. However, in this implementation, the data layout is
+in format of `[N, H, W, C]`, therefore we use 1x1 convolution to implement
+fully-connected MLP layers.
```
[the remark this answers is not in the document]
ariG23498: This comment was taken directly from the [official implementaiton](https://github.com/microsoft/SPACH/blob/main/models/shiftvit.py#L26-L29).

This was done to be as closely aligned as possible.

examples/vision/shiftvit.py
ariG23498: What would be the best way according to you? This would just be a great learning curve for me.

examples/vision/shiftvit.py
ariG23498: We meant that each stage has a variable number of stages ShiftViT Block. Should I change this sentence?

examples/vision/shiftvit.py
```diff
@@ -0,0 +1,866 @@
+"""
+Title: Shifting operation replacing attention in ViT
```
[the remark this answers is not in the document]
ariG23498: Options:

- A Vision Transformer without Attention
- Replacing Attention with a No Parameter layer in a ViT

What do you think about them?

examples/vision/shiftvit.py
fchollet: None of the alternative ways are super satisfactory to be honest. You could use `tensor_scatter_nd_add `, you could use `tf.experimental.numpy`, you could all the XLA op `dynamic_update_slice` directly...

examples/vision/shiftvit.py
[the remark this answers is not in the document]
fchollet: IMO using `Dense` seems preferable (more intuitive) unless there are performance benefits to using `Conv2D`. What does the performance look like?

examples/vision/shiftvit.py
fchollet: Yes, please clarify

examples/vision/shiftvit.py
ariG23498: Makes sense to me! Thanks.

examples/vision/shiftvit.py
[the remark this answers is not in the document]
ariG23498: Conv2D:
Time per epoch ~ 9sec
Test 1 accuracy ~ 75%

Dense:
Time per epoch ~ 10 sec
Test 1 accuracy ~ 74%

I also think with the `Conv2D` it is more about the optimization in the GPU. I would like to keep the `Conv2D` here mentioning that this design choice is taken from the official implementaiton. What do you think?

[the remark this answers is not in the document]
fchollet: Since there is no significant performance difference, then I recommend we go with `Dense` since it is more intuitive and readable.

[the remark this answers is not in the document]
ariG23498: Cool!

ariG23498:
@fchollet I have made changes as requested!

examples/vision/shiftvit.py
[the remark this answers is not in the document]
fchollet: "A Vision Transformer without Attention" sounds great. Please update the title and then you can add the generated files. Thank you!

ariG23498:
Thanks for the thoroguh review @fchollet 

We have added the generated files after some minor edits (removing spaces and adding a period to the desctiption).

_Note_: The code was more than 300 lines, so we modified the `tutobook.py` and generated the files. Will that be a problem?

fchollet (approved):
Awesome -- thank you for the great contribution! 👍

fchollet:
> Note: The code was more than 300 lines, so we modified the tutobook.py and generated the files. Will that be a problem?

As much as possible, this should be avoided, but it's not a blocker.

PR #817, opened 2022-02-28
Add instructions to install tfsim from pypi.

owenvallis:
Currently, running the example using Colab will fail when tfsim is not installed. This PR adds instructions for installing tfsim from pypi.


owenvallis:
Not sure why the PR pulled in the older commits (maybe because they were squashed when we merged last time). This is just a small change to make it easier to install tfsim when running the example notebook on Colab.

owenvallis:
Looks like the black formatter doesn't allow the `!pip install tensorflow_similarity` line. Is there a preferred way to handle install deps in the examples?

fchollet (commented):
Thanks for the PR! Everything looks good, please also update the `md` and `ipynb` files.

owenvallis:
Thanks! Should be good to go now.

fchollet (approved):
LGTM, thank you

PR #821, opened 2022-03-06
Update formatting of Applications page

leSamo:
- Format Top-1 Accuracy and Top-5 Accuracy as percentages in line with models above. [1]
- Format Parameters count in millions in line with models above. [1]
- Add newline in front of sentence explaining top-1 and top-5 accuracy, so the explanation is outside the table in line with sentences explaining other columns. [2]
- Add newline in front of list of components so its formatted correctly as a list. [2]
- RAM in upper-case. [2]


fchollet (approved):
LGTM, thank you

PR #822, opened 2022-03-08
added Hugging Face spaces at the end of markdown

merveenoyan:
Hello 👋
This PR addresses issue in https://github.com/keras-team/keras-io/issues/793. I added links to model and Space repositories at the end of the markdown, looks like this:
Ekran Resmi 2022-03-08 14 07 49
We don't have demos for all Keras examples, we will add even more in the next sprint.


fchollet (changes requested):
Thanks for the update!

The `.md` files (as well as the `.ipynb` files) are intended to be generated from the `.py` files, so let's update all 3 filetypes to be in sync.

Please also add a line break to make sure the lines added are kept below 80 chars.

merveenoyan:
Wrong commit message: I added `python` and notebooks (though in contribution guide it says I should commit only the script).

fchollet (commented):
The `ipynb` files have a lot of extraneous modifications at this time, can you restrict the modifications to just adding the 2 lines of text?

merveenoyan:
I added them to only the last markdown parts of notebooks (instead of adding a new cell and adding more changes) and aligned markdown & scripts accordingly.

fchollet (approved):
LGTM, thanks for applying the changes!

PR #835, opened 2022-03-18
Fix: Provide reference

ariG23498:
The reference to the warmup cosine was missing. I apologise for not getting to this earlier. Did not mean to plagarise any code. As soon as @sayakpaul noticed it and let me know, I made the changes.

Hope this is taken as my mistake and not otherwise.


sayakpaul:
Happens all the time even with seasoned practitioners. So, as long as one addresses it and makes the needed changes I think it should be taken in a good spirit. 

Onwards!

fchollet (approved):
LGTM, thank you

PR #841, opened 2022-03-25
fix kt pip install block

haifeng-jin:
The `pip install` block of the KerasTuner guides were incorrect.
This PR fixes them.


fchollet (approved):
Thanks for the PR!

Issue #801, opened 2022-02-10
InvalidArgumentError ner_transformers example


rodyoukai:
Hi,

I try to run your example [(https://github.com/keras-team/keras-io/blob/master/examples/nlp/ner_transformers.py)](https://github.com/keras-team/keras-io/blob/master/examples/nlp/ner_transformers.py) and I get this error: 

InvalidArgumentError:  StringToNumberOp could not correctly convert string: 
	 [[{{node StringToNumber_1}}]]
	 [[IteratorGetNext]] [Op:__inference_train_function_2490]

Function call stack:
train_function

I am newbie in Keras, can you help me?

SuryanarayanaY:
Hi @rodyoukai ,

The issue resolved now with the above PR . I have tested the example and its working fine without any error. Please refer attached [gist](https://colab.sandbox.google.com/gist/SuryanarayanaY/73a7d5d2e7035a4c26379ec6740935a1/kerasio_801-of-ner_transformers.ipynb).

Could you please check and close the issue ? Thanks!

PR #846, opened 2022-03-30
Update ner_transformers.py

lidija-jovanovska:
The bug was found in Issue #801. During the dataset export (lines 137-150), an extra line was added at the end of the dataset files (train and validation), where the number of tokens was 0, and the rest was an empty string. Fixed by adding an additional check to verify that the number of tokens is larger than 0 before writing to the files.


fchollet (approved):
LGTM, thanks for the PR. Please replicate the fix in the 2 generated files (`.md` and `.ipynb`).

lidija-jovanovska:
Hi @fchollet, I am happy to have contributed. The fix is now replicated in the generated files as well.

fchollet (approved):

PR #849, opened 2022-04-05
Add an example on distilling ViTs through attention

sayakpaul:
Main repository: https://github.com/sayakpaul/deit-tf/.

### DIFF
--- a/examples/vision/deit.py
+++ b/examples/vision/deit.py
@@ -2,7 +2,7 @@
 Title: Distilling Vision Transformers
 Author: [Sayak Paul](https://twitter.com/RisingSayak)
 Date created: 2022/04/05
-Last modified: 2022/04/05
+Last modified: 2022/04/08
 Description: Distillation of Vision Transformers through attention.
 """
 """
@@ -89,7 +89,13 @@ WEIGHT_DECAY = 0.0001
 # Data
 BATCH_SIZE = 256
 AUTO = tf.data.AUTOTUNE
-NB_CLASSES = 5
+NUM_CLASSES = 5
+
+"""
+You probably noticed that `DROPOUT_RATE` has been set 0.0. Dropout has been used
+in the implementation to keep it complete. For smaller models (like the one used in
+this example) don't need it but for bigger models, using dropout helps.
+"""
 
 """
 ## Load the `tf_flowers` dataset and prepare preprocessing utilities
@@ -102,7 +108,7 @@ and so on. However, to keep the example simple to work through, we'll discard th
 
 
 def preprocess_dataset(is_training=True):
-    def _pp(image, label):
+    def fn(image, label):
         if is_training:
             # Resize to a bigger spatial resolution and take the random
             # crops.
@@ -111,10 +117,10 @@ def preprocess_dataset(is_training=True):
             image = tf.image.random_flip_left_right(image)
         else:
             image = tf.image.resize(image, (RESOLUTION, RESOLUTION))
-        label = tf.one_hot(label, depth=NB_CLASSES)
+        label = tf.one_hot(label, depth=NUM_CLASSES)
         return image, label
 
-    return _pp
+    return fn
 
 
 def prepare_dataset(dataset, is_training=True):
@@ -152,7 +158,7 @@ class StochasticDepth(layers.Layer):
         super().__init__(**kwargs)
         self.drop_prob = drop_prop
 
-    def call(self, x, training=None):
+    def call(self, x, training=True):
         if training:
             keep_prob = 1 - self.drop_prob
             shape = (tf.shape(x)[0],) + (1,) * (len(tf.shape(x)) - 1)
@@ -267,9 +273,9 @@ class ViTClassifier(keras.Model):
         # Other layers.
         self.dropout = layers.Dropout(DROPOUT_RATE)
         self.layer_norm = layers.LayerNormalization(epsilon=LAYER_NORM_EPS)
-        self.head = layers.Dense(NB_CLASSES, name="classification_head",)
+        self.head = layers.Dense(NUM_CLASSES, name="classification_head",)
 
-    def call(self, inputs, training=None):
+    def call(self, inputs, training=True):
         n = tf.shape(inputs)[0]
 
         # Create patches and project the patches.
@@ -332,8 +338,8 @@ class ViTDistilled(ViTClassifier):
         )
 
         # Head layers.
-        self.head = layers.Dense(NB_CLASSES, name="classification_head",)
-        self.head_dist = layers.Dense(NB_CLASSES, name="distillation_head",)
+        self.head = layers.Dense(NUM_CLASSES, name="classification_head",)
+        self.head_dist = layers.Dense(NUM_CLASSES, name="distillation_head",)
 
     def call(self, inputs, training=True):
         n = tf.shape(inputs)[0]
@@ -421,6 +427,16 @@ class DeiT(keras.Model):
         self.student = student
         self.teacher = teacher
 
+        self.student_loss_tracker = keras.metrics.Mean(name="student_loss")
+        self.dist_loss_tracker = keras.metrics.Mean(name="distillation_loss")
+
+    @property
+    def metrics(self):
+        metrics = super().metrics
+        metrics.append(self.student_loss_tracker)
+        metrics.append(self.dist_loss_tracker)
+        return metrics
+
     def compile(
         self, optimizer, metrics, student_loss_fn, distillation_loss_fn,
     ):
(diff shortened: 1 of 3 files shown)


examples/vision/deit.py
```diff
@@ -0,0 +1,576 @@
...
+https://github.com/sayakpaul/deit-tf/releases/download/v0.1.0/bit_teacher_flowers.zip 
+unzip -q bit_teacher_flowers.zip 
+"""
+
+# teacher_model_file = keras.utils.get_file( 
+#     origin="https://github.com/sayakpaul/deit-tf/releases/download/v0.1.0/bit_teacher_flowers.zip",
+#     extract=True,
+#     cache_subdir="models"
+# )
+bit_teacher_flowers = keras.models.load_model("bit_teacher_flowers")
```
sayakpaul, line 512: I intentionally kept this block. 

If I use it to download the weights I run into:

```
FileExistsError: [Errno 17] File exists: '/root/.keras/datasets/bit_teacher_flowers'.
```

sayakpaul:
Not sure why the Black check is failing. I am on `19.10b0` version of `black` and I did run `black examples/vision/deit.py` before pushing.

examples/vision/deit.py
```diff
@@ -0,0 +1,575 @@
...
+"""
+## Training through distillation
+"""
+
+deit_tiny = ViTDistilled()
+deit_distiller = DeiT(student=deit_tiny, teacher=bit_teacher_flowers)
+
+lr_scaled = (BASE_LR / 512) * BATCH_SIZE
+deit_distiller.compile(
+    optimizer=tfa.optimizers.AdamW(weight_decay=WEIGHT_DECAY, learning_rate=lr_scaled),
```
fchollet, line 522: Fun fact: AdamW has recently been included in the API (in Keras 2.9 it's at `keras.optimizers.experimental.AdamW` and it will be at `keras.optimizers.AdamW` likely in 2.10). But let's keep TFA for now.

examples/vision/deit.py
```diff
@@ -0,0 +1,575 @@
...
+
+The authors use an array of different augmentation techniques including MixUp
+([Zhang et al.](https://arxiv.org/abs/1710.09412)),
+RandAugment ([Cubuk et al.](https://arxiv.org/abs/1909.13719)),
+and so on. However, to keep the example simple to work through, we'll discard them.
+"""
+
+
+def preprocess_dataset(is_training=True):
+    def _pp(image, label):
```
fchollet, line 105: What's `_pp`? For a generic function name you could use `fn`...

examples/vision/deit.py
```diff
@@ -0,0 +1,575 @@
...
+
+# Training
+NUM_EPOCHS = 20
+BASE_LR = 0.0005
+WEIGHT_DECAY = 0.0001
+
+# Data
+BATCH_SIZE = 256
+AUTO = tf.data.AUTOTUNE
+NB_CLASSES = 5
```
fchollet, line 92: Nit: prefer `num_` for the number prefix for general consistency (as you do elsewhere in this example).

examples/vision/deit.py
```diff
@@ -0,0 +1,575 @@
...
+NUM_PATCHES = (RESOLUTION // PATCH_SIZE) ** 2
+LAYER_NORM_EPS = 1e-6
+PROJECTION_DIM = 192
+NUM_HEADS = 3
+NUM_LAYERS = 12
+MLP_UNITS = [
+    PROJECTION_DIM * 4,
+    PROJECTION_DIM,
+]
+DROPOUT_RATE = 0.0
```
fchollet, line 81: Always a bit weird to have default option at 0 in a code example (makes you question whether it should be introduced at all). Should it be non-zero?

examples/vision/deit.py
```diff
@@ -0,0 +1,575 @@
...
+which is used in DeiT for regularization.
+"""
+
+# Referred from: github.com:rwightman/pytorch-image-models.
+class StochasticDepth(layers.Layer):
+    def __init__(self, drop_prop, **kwargs):
+        super().__init__(**kwargs)
+        self.drop_prob = drop_prop
+
+    def call(self, x, training=None):
```
fchollet, line 155: Prefer defaulting `training` to a bool, e.g. `True` (it will be passed by the parent anyway...). In the core API we only use the default `None` in order to keep compatibility with the symbolic learning phase in TF1.

examples/vision/deit.py
```diff
@@ -0,0 +1,575 @@
...
+        self.cls_token = tf.Variable(
+            initial_value=initial_value, trainable=True, name="cls"
+        )
+
+        # Other layers.
+        self.dropout = layers.Dropout(DROPOUT_RATE)
+        self.layer_norm = layers.LayerNormalization(epsilon=LAYER_NORM_EPS)
+        self.head = layers.Dense(NB_CLASSES, name="classification_head",)
+
+    def call(self, inputs, training=None):
```
fchollet, line 272: Likewise.

examples/vision/deit.py
```diff
@@ -0,0 +1,575 @@
...
+        self.optimizer.apply_gradients(zip(gradients, trainable_vars))
+
+        # Update the metrics configured in `compile()`.
+        student_predictions = (cls_predictions + dist_predictions) / 2
+        self.compiled_metrics.update_state(y, student_predictions)
+
+        # Return a dict of performance.
+        results = {m.name: m.result() for m in self.metrics}
+        results.update(
+            {"student_loss": student_loss, "distillation_loss": distillation_loss}
```
fchollet, line 464: To keep track of current-epoch averages of these losses, rather than just last-batch numbers, use `Mean` metric trackers (created in `compile()` for instance)

examples/vision/deit.py
```diff
@@ -0,0 +1,575 @@
...
+
+        # Calculate the loss.
+        student_loss = self.student_loss_fn(y, y_prediction)
+
+        # Update the metrics.
+        self.compiled_metrics.update_state(y, y_prediction)
+
+        # Return a dict of performance.
+        results = {m.name: m.result() for m in self.metrics}
+        results.update({"student_loss": student_loss})
```
fchollet, line 483: Same here.

examples/vision/deit.py
```diff
@@ -0,0 +1,575 @@
...
+
+"""shell
+wget -q https://github.com/sayakpaul/deit-tf/releases/download/v0.1.0/bit_teacher_flowers.zip
+unzip -q bit_teacher_flowers.zip
+"""
+
+# teacher_model_file = keras.utils.get_file(
+#     origin="https://github.com/sayakpaul/deit-tf/releases/download/v0.1.0/bit_teacher_flowers.zip",
+#     extract=True,
+#     cache_subdir="models"
```
fchollet, line 509: This statement seems to work for me when I try it?

It's fine to use the command line anyhow.

fchollet (commented):
Thanks for the PR! Super interesting article. I really appreciate the effort that went into clearly commenting each section of the code, too, the flow is very clear and readable.

examples/vision/deit.py
sayakpaul: It was for preprocessing :D But I changed it to `fn`.

examples/vision/deit.py
sayakpaul: Actually, I kept it to make the implementation complete. For bigger models (i.e., with more number of attention heads and transformer blocks), you might need this argument. Hence.

I will add a comment for this.

sayakpaul:
@fchollet thanks for your feedback. I have addressed them. 

I am still clueless about the `black` test failing.

kdave123:
Great article @sayakpaul , for the failing `black` test case, maybe you can try running black command on .py file again (don't make any changes after that) and push the code.

sayakpaul:
> Great article @sayakpaul , for the failing `black` test case, maybe you can try running black command on .py file again (don't make any changes after that) and push the code.

I have done that already, didn't help.

fchollet:
As per `.github/workflows/continuous_integration.yml`, we use `black==22.1.0`. Upgrading should do it.

fchollet (commented):
Thanks for the updates! I just added final copyedits -- you can now add the generated files.

sayakpaul:
@fchollet added the generated files. Ready for you to take the reigns.

sayakpaul:
@fchollet I was previewing the site locally after adding the generated files. There's a `:::python` at the top of the code blocks. Could you suggest what can be done to remove it?

image

fchollet:
It looks fine on my end. My guess is that you do not have the markdown extension that enables Python syntax highlighting (`codehilite` I think) in your local environment. See `autogen.py`:

```
html_content = markdown.markdown(
            md_content,
            extensions=["fenced_code", "tables", "codehilite", "mdx_truly_sane_lists",],
            extension_configs={"codehilite": {"guess_lang": False,},},
        )
```

fchollet (approved):
LGTM

sayakpaul:
Thank you @fchollet. This PR was really fun.

PR #850, opened 2022-04-09
Tweets Classification using Tensorflow decision forest and pre-trained embeddings

IMvision12:
Greetings, @fchollet! I'd like to add this example which shows how to use TensorFlow Decision Forest for text classification as well as how to use pre-trained embedding with TFDF. I came across a fantastic code example [Classification with tfdf](https://keras.io/examples/structured_data/classification_with_tfdf/) and want to extend this example for text classification.

Please let me know whether this is a valid example, and I'll get to work on the text copy and code.

[Ipynb file](https://github.com/IMvision12/Tweets-Classification-NLP/blob/main/Tweets_Classification_using_TFDF_%2B_pre_trained_embeddings.ipynb)


examples/nlp/Tweet-classification-using-TFDF.py
```diff
@@ -0,0 +1,130 @@
...
+
+"""shell
+! pip install kaggle
+! mkdir ~/.kaggle
+! cp kaggle.json ~/.kaggle/
+! chmod 600 ~/.kaggle/kaggle.json
+"""
+
+"""shell
+!kaggle competitions download -c nlp-getting-started
```
fchollet, line 20: Note that the `!` should not be added for shell commands.

examples/nlp/Tweet-classification-using-TFDF.py
```diff
@@ -0,0 +1,130 @@
...
+with zipfile.ZipFile("/content/nlp-getting-started.zip", "r") as zip_ref:
+    zip_ref.extractall("/content/")
+
+import pandas as pd
+import numpy as np
+import re
+
+import tensorflow as tf
+import tensorflow_decision_forests as tfdf
+from sklearn.metrics import precision_recall_fscore_support, accuracy_score
```
fchollet, line 34: Don't use metrics from sklearn, use those from Keras, e.g. `Recall`, `Precision`, `AUC`, etc.

examples/nlp/Tweet-classification-using-TFDF.py
```diff
@@ -0,0 +1,130 @@
...
+        print(f"Target: {target} (real disaster)")
+    else:
+        print(f"Target: {target} (not real disaster)")
+    print(f"Text: {text}\n")
+    print("-" * 10)
+
+from sklearn.model_selection import train_test_split
+
+# Use train_test_split to split training data into training and validation sets
+train_sentences, val_sentences, train_labels, val_labels = train_test_split(
```
fchollet, line 75: You can do this by hand, it doesn't take more code.

examples/nlp/Tweet-classification-using-TFDF.py
```diff
@@ -0,0 +1,130 @@
...
+    train_df_shuffled["text"].to_numpy(),
+    train_df_shuffled["target"].to_numpy(),
+    test_size=0.1,  # dedicate 10% of samples to validation set
+    random_state=42,
+)  # random state for reproducibility
+
+train_dataset = tf.data.Dataset.from_tensor_slices((train_sentences, train_labels))
+valid_dataset = tf.data.Dataset.from_tensor_slices((val_sentences, val_labels))
+
+train_dataset
```
fchollet, line 85: The code should be thought of as a standalone Python script: anything you want to display should use `print` / etc.

examples/nlp/Tweet-classification-using-TFDF.py
```diff
@@ -0,0 +1,130 @@
...
+    trainable=False,
+    name="USE",
+)
+
+inputs = tf.keras.layers.Input(shape=(), dtype=tf.string)
+x = sentence_encoder_layer(inputs)
+preprocessor = tf.keras.Model(inputs=inputs, outputs=x)
+
+model = tfdf.keras.GradientBoostedTreesModel(preprocessing=preprocessor, num_trees=1000)
+model.compile(metrics=["accuracy"])
```
fchollet, line 103: Can you not add your desired metrics here instead of leveraging sklearn?

examples/nlp/Tweet-classification-using-TFDF.py
```diff
@@ -0,0 +1,130 @@
...
+
+model.fit(x=train_dataset)
+
+print("Evaluation:", model.evaluate(valid_dataset))
+
+print("Evaluation :", model.evaluate(valid_dataset))
```
fchollet, line 109: Line is redundant

fchollet (commented):
Thanks for the draft PR! This is indeed an interesting and valuable example.

Adding @achoum for visibility.

achoum:
Cool example :).

*A few nits.

IMvision12:
Thanks for the feedback @fchollet @achoum, I will address the comments and improve the code as well as text copy

IMvision12:
@fchollet @achoum I have addressed the comments and updated the code, after the review, I will add the text copy

examples/nlp/Tweet-classification-using-TFDF.py
```diff
@@ -0,0 +1,133 @@
+"""
+Title: FILLME
+Author: FILLME
+Date created: FILLME
+Last modified: FILLME
+Description: FILLME
```
fchollet, line 6: Time to fill these :)

examples/nlp/Tweet-classification-using-TFDF.py
```diff
@@ -0,0 +1,133 @@
...
+
+import pandas as pd
+import numpy as np
+import tensorflow as tf
+from tensorflow import keras
+import tensorflow_hub as hub
+from tensorflow.keras import layers
+import tensorflow_decision_forests as tfdf
+from sklearn.model_selection import train_test_split
+from tensorflow.keras.metrics import Recall, Precision, Accuracy, AUC
```
fchollet, line 17: Just access these `metrics`, don't import individual objects.

examples/nlp/Tweet-classification-using-TFDF.py
```diff
@@ -0,0 +1,133 @@
...
+"""
+
+import pandas as pd
+import numpy as np
+import tensorflow as tf
+from tensorflow import keras
+import tensorflow_hub as hub
+from tensorflow.keras import layers
+import tensorflow_decision_forests as tfdf
+from sklearn.model_selection import train_test_split
```
fchollet, line 16: Don't import this, do train/test split by hand.

examples/nlp/Tweet-classification-using-TFDF.py
```diff
@@ -0,0 +1,133 @@
...
+from tensorflow.keras.metrics import Recall, Precision, Accuracy, AUC
+
+# Turn .csv files into pandas DataFrame's
+train_df = pd.read_csv(
+    "https://raw.githubusercontent.com/IMvision12/Tweets-Classification-NLP/main/train.csv"
+    "https://raw.githubusercontent.com/IMvision12/Tweets-Classification-NLP/main/train.csv"
```
fchollet, line 22: Line is redundant.

examples/nlp/Tweet-classification-using-TFDF.py
```diff
@@ -0,0 +1,133 @@
...
+)
+test_df = pd.read_csv(
+    "https://raw.githubusercontent.com/IMvision12/Tweets-Classification-NLP/main/test.csv"
+)
+train_df.head()
```
fchollet, line 27: Anything you want to print should be printed explicitly.

examples/nlp/Tweet-classification-using-TFDF.py
```diff
@@ -0,0 +1,133 @@
...
+
+train_df.shape
```
fchollet, line 29: Same.

examples/nlp/Tweet-classification-using-TFDF.py
```diff
@@ -0,0 +1,133 @@
...
+
+train_df_shuffled = train_df.sample(frac=1, random_state=42)
+train_df_shuffled.drop(["id", "keyword", "location"], axis=1, inplace=True)
+train_df_shuffled.reset_index(inplace=True, drop=True)
+train_df_shuffled.head()
+
+train_df_shuffled.info()
```
fchollet, line 36: Same.

examples/nlp/Tweet-classification-using-TFDF.py
```diff
@@ -0,0 +1,133 @@
...
+    print(f"Target : {train_df_shuffled['target'][ind]}")
+    print(f"Text : {train_df_shuffled['text'][ind]}")
+    if counter == 5:
+        break
+
+# Use train_test_split to split training data into training and validation sets
+train_sentences, val_sentences, train_labels, val_labels = train_test_split(
+    train_df_shuffled["text"].to_numpy(),
+    train_df_shuffled["target"].to_numpy(),
+    test_size=0.2,  # dedicate 10% of samples to validation set
```
fchollet, line 55: (the comment doesn't match the code. But you should remove this code and do the split by hand)

examples/nlp/Tweet-classification-using-TFDF.py
fchollet: Can't you just put `Recall(), Precision(), Accuracy(), AUC()` here?

fchollet (commented):
Thanks for the updates!

IMvision12:
@fchollet I have addressed the comments and added text copy

examples/nlp/Tweet-classification-using-TFDF.py
```diff
@@ -0,0 +1,215 @@
...
+"""
+
+# Turn .csv files into pandas DataFrame's
+train_df = pd.read_csv(
+    "https://raw.githubusercontent.com/IMvision12/Tweets-Classification-NLP/main/train.csv"
+)
+print(train_df.head())
+
+# Printing shape of the training dataset
+print(train_df.shape)
```
fchollet, line 72: When printing values, make sure to include to sentence to give context, e.g.

```
print(f"Training dataset shape: {train_df.shape}")
```

examples/nlp/Tweet-classification-using-TFDF.py
```diff
@@ -0,0 +1,215 @@
...
+# Dropping id, keyword and location columns as these columns consists of mostly nan values
+# we will be using only text and target columns
+train_df_shuffled.drop(["id", "keyword", "location"], axis=1, inplace=True)
+train_df_shuffled.reset_index(inplace=True, drop=True)
+print(train_df_shuffled.head())
+
+print(train_df_shuffled.info())
+
+# Printing total number of Disaster and non-Disaster tweets
+print(train_df_shuffled.target.value_counts())
```
fchollet, line 85: Same here

examples/nlp/Tweet-classification-using-TFDF.py
```diff
@@ -0,0 +1,215 @@
...
+
+The Preprocessor is then passed to the GradientBoostedTreesModel which will be used for
+predicting disaster tweets.
+
+
+"""
+
+inputs = layers.Input(shape=(), dtype=tf.string)
+# Passing sentences to sentence_encoder_layer
+x = sentence_encoder_layer(inputs)
```
fchollet, line 160: `outputs =`

examples/nlp/Tweet-classification-using-TFDF.py
```diff
@@ -0,0 +1,215 @@
...
+"""
+## Downloading Pre-trained Embeddings
+
+The Universal Sentence Encoder encodes text into high-dimensional vectors that can be
+used for text classification, semantic similarity, clustering and other natural language
+tasks.It is trained on a variety of data sources and a variety of tasks. The input is
+variable length English text and the output is a 512 dimensional vector.
+
+To understand better about pre-trained embeddings : [Universal Sentence
+Encoder](https://tfhub.dev/google/universal-sentence-encoder/4)
```
fchollet, line 132: Make sure markdown links fit on a single line. Follow the guide: https://github.com/keras-team/keras-io/blob/master/contributor_guide.md

examples/nlp/Tweet-classification-using-TFDF.py
```diff
@@ -0,0 +1,215 @@
...
+
+
+def create_dataset(dataframe, training=False):
+    df = dataframe.copy()
+    dataset = tf.data.Dataset.from_tensor_slices(
+        (df["text"].to_numpy(), df["target"].to_numpy())
+    )
+
+    if training:
+        dataset = dataset.repeat()
```
fchollet, line 112: Why do you need `repeat()`?

examples/nlp/Tweet-classification-using-TFDF.py
```diff
@@ -0,0 +1,215 @@
...
+)
+
+"""
+## Train the model
+"""
+
+# Compiling model
+model.compile(metrics=["Accuracy", "Recall", "Precision", "AUC"])
+# Training the model
+model.fit(x=train_ds, steps_per_epoch=200)
```
fchollet, line 178: Please clarify -- since `steps_per_epoch` is specified, but `epochs` isn't, what is the intended behavior?

fchollet (commented):
Thanks for the update!

examples/nlp/Tweet-classification-using-TFDF.py
```diff
@@ -0,0 +1,215 @@
...
+"""
+## Introduction
+
+[TensorFlow Decision Forests](https://www.tensorflow.org/decision_forests) (TF-DF)
+provide powerful models, especially with structured data,
+It is a collection of state-of-the-art algorithms of Decision Forest models that are
+compatible with Keras APIs. The module includes Random Forests, Gradient Boosted Trees,
+and CART, and can be used for regression, classification, and ranking tasks.
+
+Alternatively for getting started
```
achoum, line 17: Can you add a link to the "Using text and neural network features" official tutorial:
https://www.tensorflow.org/decision_forests/tutorials/intermediate_colab

TF-DF can consume text feature natively (see https://arxiv.org/abs/2009.09991). This is slower than using pre-trained embedding, but it is often simpler to use, give good results, and combines well with pre-trained embedding (when one is available). It is worth mentioning it in the introduction.

examples/nlp/Tweet-classification-using-TFDF.py
```diff
@@ -0,0 +1,215 @@
...
+    c. location - the location the tweet was sent from (may be blank)
+
+    d. keyword - a particular keyword from the tweet (may be blank)
+
+e. target - in train.csv only, this denotes whether a tweet is about a real disaster (1)
+or not (0)
+"""
+
+# Turn .csv files into pandas DataFrame's
+train_df = pd.read_csv(
```
achoum, line 66: Since "train_df" contains both the training and testing dataset, what about calling this variable something like "all_df"

examples/nlp/Tweet-classification-using-TFDF.py
achoum: TF-DF trains on exactly one epoch of the dataset.
See: https://www.tensorflow.org/decision_forests/migration#dataset_io

examples/nlp/Tweet-classification-using-TFDF.py
```diff
@@ -0,0 +1,215 @@
...
+    df = dataframe.copy()
+    dataset = tf.data.Dataset.from_tensor_slices(
+        (df["text"].to_numpy(), df["target"].to_numpy())
+    )
+
+    if training:
+        dataset = dataset.repeat()
+        dataset = dataset.shuffle(1000)
+
+    dataset = dataset.batch(32)
```
achoum, line 115: Unlike for neural network, the batch size has not effect on the model. However, smaller batch size make dataset reading slower. Therefore, it is a good idea to use large batch size. By default, there will be a warning when training TF-DF with a batchsize <100.

examples/nlp/Tweet-classification-using-TFDF.py
```diff
@@ -0,0 +1,215 @@
...
+train_df = train_df_shuffled.drop(test_df.index)
+
+print(f"Using {len(train_df)} samples for training and {len(test_df)} for validation")
+
+"""
+## Convert data to tf.Dataset
+"""
+
+
+def create_dataset(dataframe, training=False):
```
achoum, line 105: Optional:
What about using the function "pd_dataframe_to_tf_dataset". It is made specially to feed Pandas dataframes to 
 TF-DF  models.
https://www.tensorflow.org/decision_forests/api_docs/python/tfdf/keras/pd_dataframe_to_tf_dataset

In your case, the dataset is effectively consumed by the tf-hub embedding, but it should still work.

examples/nlp/Tweet-classification-using-TFDF.py
```diff
@@ -0,0 +1,215 @@
...
+
+To understand better about pre-trained embeddings : [Universal Sentence
+Encoder](https://tfhub.dev/google/universal-sentence-encoder/4)
+
+
+"""
+
+sentence_encoder_layer = hub.KerasLayer(
+    "https://tfhub.dev/google/universal-sentence-encoder/4",
+    input_shape=[],
```
achoum, line 139: I don't think any of those arguments (except for the tfhub id) are required here.

examples/nlp/Tweet-classification-using-TFDF.py
```diff
@@ -0,0 +1,215 @@
...
+
+
+"""
+
+inputs = layers.Input(shape=(), dtype=tf.string)
+# Passing sentences to sentence_encoder_layer
+x = sentence_encoder_layer(inputs)
+preprocessor = keras.Model(inputs=inputs, outputs=x)
+
+model = tfdf.keras.GradientBoostedTreesModel(
```
achoum, line 163: max_depth=100 makes little sense with Gradient Boosted Tree model. The model will overfit immediately.
A tree with depth 100 can contain 2^100 nodes. That many (many) more times the number of atoms in the entire universe. The depth is generally in ~3-8. With, 8 when you have large dataset with complex interactions.

num_trees=10 also make little sense. GBT generally work with 200-100 trees. <200 trees is possible but unlikely in your case.

max_vocab_count=1000. This is only used if text features are consumed natively by TF-DF. In your example, the text features are processed by a pre-trained embedding. So this parameter is not used.

What about using the default hyper-parameters, or something close to them? Generally, there are good starting points.

examples/nlp/Tweet-classification-using-TFDF.py
achoum, line 160: Nit: You can also concatenate the raw text feature to the embedding. This way, the TF-DF model consumes both the pre-trained embedding and the raw text.

examples/nlp/Tweet-classification-using-TFDF.py
```diff
@@ -0,0 +1,130 @@
...
+
+model.fit(x=train_dataset)
+
+print("Evaluation:", model.evaluate(valid_dataset))
+
+print("Evaluation :", model.evaluate(valid_dataset))
+
+
+def calculate_results(y_true, y_pred):
+    model_accuracy = accuracy_score(y_true, y_pred) * 100
```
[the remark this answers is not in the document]
IMvision12: @achoum I guess you are viewing older version of Tweets-classification-using-TFDF.py, I have removed that function in the newer version.

examples/nlp/Tweet-classification-using-TFDF.py
IMvision12: Yeah sure I will add a reference link

examples/nlp/Tweet-classification-using-TFDF.py
IMvision12: I Tried `pd_dataframe_to_tf_dataset` earlier but i was getting errors :
Capture

examples/nlp/Tweet-classification-using-TFDF.py
IMvision12: @achoum That's a great idea but how do i do that, if I cannot concatenate raw text with embeddings there will be a shape error

examples/nlp/Tweet-classification-using-TFDF.py
fchollet: Then `repeat` does nothing in this case, right?

Can you clarify what the intended behavior of `steps_per_epoch` is in this setting?

fchollet: IMO the `steps_per_epoch` argument should not be accepted and should raise an error since the concept of epochs is ill-defined here.

IMvision12:
@fchollet @achoum thanks for your feedback I have addressed them

examples/nlp/Tweet-classification-using-TFDF.py
IMvision12: okay I will use default hyper-parameters

examples/nlp/Tweet-classification-using-TFDF.py
```diff
@@ -0,0 +1,207 @@
...
+also text classification using pre-trained embeddings
+[Notebook](https://www.tensorflow.org/decision_forests/tutorials/intermediate_colab).
+
+In this example we will use Gradient Boosted Trees with pre-trained embeddings to
+classify disaster tweets.
+"""
+
+"""
+Install Tensorflow Decision Forest using following command : 
+`!pip3 install -U tensorflow_decision_forests`
```
fchollet, line 29: Just `pip install tensorflow_decision_forests`

examples/nlp/Tweet-classification-using-TFDF.py
```diff
@@ -0,0 +1,207 @@
...
+"""
+## Get the data
+
+Dataset is avaliable at [kaggle](https://www.kaggle.com/c/nlp-getting-started)
+
+dataset description :
+
+1. Files
+
+    a. train.csv - the training set
```
fchollet, line 54: Remove indent

examples/nlp/Tweet-classification-using-TFDF.py
```diff
@@ -0,0 +1,207 @@
...
+
+2. Columns
+
+    a. id - a unique identifier for each tweet
```
fchollet, line 58: Remove indent and blank lines between items

examples/nlp/Tweet-classification-using-TFDF.py
```diff
@@ -0,0 +1,207 @@
...
+
+print(f"Using {len(train_df)} samples for training and {len(test_df)} for validation")
+
+"""
+## Convert data to tf.Dataset
+"""
+
+
+def create_dataset(dataframe):
+    df = dataframe.copy()
```
fchollet, line 113: Why is the copy operation necessary?

examples/nlp/Tweet-classification-using-TFDF.py
```diff
@@ -0,0 +1,207 @@
...
+
+plt.subplot(1, 2, 2)
+plt.plot([log.num_trees for log in logs], [log.evaluation.loss for log in logs])
+plt.xlabel("Number of trees")
+plt.ylabel("Loss")
+
+plt.show()
+
+"""
+## Evaluating on test data
```
fchollet, line 201: Please add a conclusion section outlining what was learned from this example.

examples/nlp/Tweet-classification-using-TFDF.py
```diff
@@ -0,0 +1,207 @@
...
+model = tfdf.keras.GradientBoostedTreesModel(preprocessing=preprocessor)
+
+"""
+## Train the model
+"""
+
+# Compiling model
+model.compile(metrics=["Accuracy", "Recall", "Precision", "AUC"])
+# Here we do not specify epochs as, TF-DF trains exactly one epoch of the dataset
+model.fit(x=train_ds)
```
fchollet, line 170: We need to monitor validation accuracy on a validation set.

examples/nlp/Tweet-classification-using-TFDF.py
```diff
@@ -0,0 +1,207 @@
...
+plt.xlabel("Number of trees")
+plt.ylabel("Loss")
+
+plt.show()
+
+"""
+## Evaluating on test data
+"""
+
+results = model.evaluate(test_ds, return_dict=True, verbose=0)
```
fchollet, line 204: For test data, show inference (predict)

fchollet (commented):
Thanks for the updates!

examples/nlp/Tweet-classification-using-TFDF.py
IMvision12: As per the documentation, TFDF does not require a validation set [docs](https://www.tensorflow.org/decision_forests/migration#validation_dataset)

IMvision12:
@fchollet thanks for your feedback I have addressed them

fchollet:
@achoum any remaining comments?

examples/nlp/Tweet-classification-using-TFDF.py
achoum: The training of DF is "applied" on each example once and exactly once.
The ideal / recommended solution is to create a finite dataset (e.g. without repeating). When the dataset runs out-of-examples, the training steps.

For non distributed training, if the user uses a datasets with a repeat instruction, or if keras' fit is called with epoch!=1, an error (of a warning, depending on the configuration options) is raised. While not recommended, I suspect steps_per_epoch can make sense if the tf dataset does not export a concept of epoch.

For distributed training, it is currently impossible to use finite datasets with async training (see internal discussions; we have some ideas and a prototype to resolve this issue). Once resolved, the same logic as for non distributed training will be implemented. In the meantime, steps_per_epoch seems needed.

examples/nlp/Tweet-classification-using-TFDF.py
achoum: In this example, the call should be:
model.fit(train_ds)

When running the statement:
running model.fit(x=train_ds, steps_per_epoch=200)
Since "epoch" defaults to 1, the training will be done on batch_size * 200 examples (which can be more or less than the number of training examples).

examples/nlp/Tweet-classification-using-TFDF.py
```diff
@@ -0,0 +1,272 @@
...
+
+print(df_shuffled.info())
+
+# Printing total number of Disaster and non-Disaster tweets
+print(
+    f"Total Number of disaster and non-disaster tweets\n{df_shuffled.target.value_counts()}"
+)
+
+# Viewing 5 records from training data
+for ind, counter in enumerate(df_shuffled.index):
```
achoum, line 89: Nit: Replace block with:

for index, example in df[:5].iterrows():
  print(f"Example #{index}")
  print(f"\tTarget : {example['target']}")
  print(f"\tText : {example['text']}")

examples/nlp/Tweet-classification-using-TFDF.py
```diff
@@ -0,0 +1,272 @@
...
+"""
+
+sentence_encoder_layer = hub.KerasLayer(
+    "https://tfhub.dev/google/universal-sentence-encoder/4"
+)
+
+"""
+## Build a model
+
+We will create 2 models In first model raw text will be directly passed to the Gradient
```
achoum, line 142: This is reversed, right?

The "first model" (model_1) uses pre-trained embeddings, while the "second model" (model_2) consumes text natively.

examples/nlp/Tweet-classification-using-TFDF.py
```diff
@@ -0,0 +1,272 @@
...
+logs_2 = model_2.make_inspector().training_logs()
+
+print(logs_1)
+print(logs_2)
+
+"""
+The `model.summary()` function returns a variety of information about your decision trees
+model, including model type, task, input features, and feature importance.
+
+In our the inputs to the GradientBoostedTreesModel are 512 dimensional vectors so, it
```
achoum, line 193: In model 1.

examples/nlp/Tweet-classification-using-TFDF.py
```diff
@@ -0,0 +1,272 @@
...
+prints all information of those vectors.
+"""
+
+print(model_1.summary())
+
+print(model_2.summary())
+
+logs = model_1.make_inspector().training_logs()
```
achoum, line 201: This is "logs_1" defined above. No need to re-create it.

examples/nlp/Tweet-classification-using-TFDF.py
```diff
@@ -0,0 +1,272 @@
...
+plt.subplot(1, 2, 1)
+plt.plot([log.num_trees for log in logs], [log.evaluation.accuracy for log in logs])
+plt.xlabel("Number of trees")
+plt.ylabel("Accuracy")
+
+plt.subplot(1, 2, 2)
+plt.plot([log.num_trees for log in logs], [log.evaluation.loss for log in logs])
+plt.xlabel("Number of trees")
+plt.ylabel("Loss")
+
```
achoum, line 213: Nit: Factor the plotting code in a function.

examples/nlp/Tweet-classification-using-TFDF.py
```diff
@@ -0,0 +1,272 @@
...
+for name, value in results.items():
+    print(f"{name}: {value:.4f}")
+
+"""
+# Predicting on validation data
+"""
+
+counter = 0
+test_df = test_df.sample(frac=0.1)
```
achoum, line 250: No need to shuffle the test data.

examples/nlp/Tweet-classification-using-TFDF.py
```diff
@@ -0,0 +1,272 @@
...
+for index, row in test_df.iterrows():
+    text = tf.expand_dims(row["text"], axis=0)
+    preds = model_1.predict_step(text)
+    preds = tf.squeeze(tf.round(preds))
+    print(f"Text: {row['text']}")
+    print(f"Prediction: {preds}")
+    print(f"Ground Truth : {row['target']}")
+    if counter == 5:
```
achoum, line 258: Use "index" instead.

achoum:
Looks good. A few last comments.

examples/nlp/Tweet-classification-using-TFDF.py
IMvision12: Yeah

IMvision12:
@achoum  thanks for your feedback I have addressed them.

achoum:
Thanks.
LGTM.

IMvision12:
@fchollet are there any more comments?

fchollet (approved):
Looks great! I added copyedits, please pull them first.

You can now add the generated files!

IMvision12:
@fchollet I have generated .ipynb and .md files but many code cells got combined so I formated the code, could please review it.
Code

IMvision12:
@fchollet I have added autogenerated files

fchollet (approved):
LGTM, thank you! 👍

PR #856, opened 2022-04-17
Fix `transformer_asr.md`'s markdown

joshiayush:
Leave a line feed before starting a `snippet` inside `html` tags because markdown formatter doesn't identify `snippets` joined with `html` tags as a valid markdown syntax.

## Before

Screenshot from 2022-04-17 12-03-46
Screenshot from 2022-04-17 12-04-15

## After

Screenshot from 2022-04-17 12-04-39
Screenshot from 2022-04-17 12-04-57

Signed-off-by: Ayush Joshi <ayush854032@gmail.com>

### DIFF
--- a/examples/audio/md/transformer_asr.md
+++ b/examples/audio/md/transformer_asr.md
@@ -334,12 +334,13 @@ def get_data(wavs, id_to_text, maxlen=50):
 ```
 
 <div class="k-default-codeblock">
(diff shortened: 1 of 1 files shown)


joshiayush:
Sorry, now I know that it will break the [keras.io](https://keras.io/) site.

PR #861, opened 2022-04-26
Update the required Python version

haifeng-jin:
Made this change according to the TF requirements.
https://www.tensorflow.org/install/pip


fchollet (approved):
LGTM, thanks

PR #863, opened 2022-04-28
remove the name of kt

### LINKED ISSUE #862: Stop using "kt" as the canonical import for keras_tuner in guides
Both keras_cv, and keras_nlp will not recommend an abbreviated import. Both for consistency, and clear branding for keras, we should consider dropping the `import keras_tuner as kt` pattern for keras_tuner.

haifeng-jin:
resolves #862


fchollet (approved):
LGTM, thank you

PR #864, opened 2022-04-28
Stop replacing lang syntax for fenced markdown

mattdangerw:
The CodeHilite to the markdown renderer now supports the usual syntax,
and newer version appear to no longer supports the
`::: syntax`.

I'm not sure exactly when this happened, but we can just depend on
the latest version of python-markdown.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,6 @@
 pygments>=2.7.4
 jinja2
-markdown
+markdown>=3.3
 requests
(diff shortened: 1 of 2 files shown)


mattdangerw:
@fchollet is there a way to test this our on whatever machine we publish from? This fixed syntax highlighting for me, both in a local pip workflow and docker workflow.

fchollet (approved):
LGTM. I'll test it

PR #865, opened 2022-05-04
Fix vq_vae.py

louisgeisler:
There was a problem with the function get_decoder, because when it wants to get the output shape of the encoder, it directly calls the get_encoder function without parameter, so it assumes that the current latent_dim is the default latent_dim: 16. So, to address this issue, it can just pass the current latent_dim parameter when calling get_encoder function.


fchollet (approved):
Good catch! LGTM. Please also update the `.md` file and the `.ipynb` file with the same change.

louisgeisler:
Thanks!

Ok, I updated them 👍

fchollet (approved):

PR #873, opened 2022-05-13
Fix typo

tornikeo:
Add a missing word in `StringLookup` [description](https://www.tensorflow.org/guide/keras/preprocessing_layers#categorical_features_preprocessing).


fchollet (commented):
Thanks for the PR! Please also the typo in the corresponding `md` and `ipynb` files.

tornikeo:
So, in the future, I should only change the ipynb and just then just `make container-test`, No need to manually copy stuff over, right?

fchollet (approved):

PR #894, opened 2022-05-31
added getconfig in neural machine translation

rushic24:
model.save doesn't work without these changes

```
NotImplementedError: 
Layer PositionalEmbedding has arguments ['self', 'sequence_length', 'vocab_size', 'embed_dim']
in `__init__` and therefore must override `get_config()`.

Example:

class CustomLayer(keras.layers.Layer):
    def __init__(self, arg1, arg2):
        super().__init__()
        self.arg1 = arg1
        self.arg2 = arg2

    def get_config(self):
        config = super().get_config()
        config.update({
            "arg1": self.arg1,
            "arg2": self.arg2,
        })
        return config
```

### DIFF
--- a/examples/nlp/ipynb/neural_machine_translation_with_transformer.ipynb
+++ b/examples/nlp/ipynb/neural_machine_translation_with_transformer.ipynb
@@ -53,7 +53,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 0,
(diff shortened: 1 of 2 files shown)


fchollet (approved):
LGTM, thanks

rushic24:
Edited

fchollet (approved):

PR #899, opened 2022-06-01
update cut_mix_mix_up_and_rand_augment

ncmatson:
Remove sentence about decorating with `tf.function`.  `tf.Dataset` does it for you. @LukeWood


LukeWood:
I can confirm that this is the correct change to make.  Thanks Cameron!

fchollet (commented):
Thanks for the PR! Could you please replicate the fix in the `md` and `ipynb` files (or regenerate the example files)

LukeWood:
> Thanks for the PR! Could you please replicate the fix in the `md` and `ipynb` files (or regenerate the example files)

I can do this as I’ll need to edit out some warnings after generating them.

fchollet (approved):

PR #911, opened 2022-06-13
VQ-VAE example: pin tensorflow-probability to last known working version

reinvantveer:
The usage of the `sample()` method used here from tensorflow-probability doesn't work for any version after 0.13.0 for this particular example. See also: https://stackoverflow.com/questions/69620035/typeerror-dimension-value-must-be-integer-or-none-in-keras-vq-vae-example/72561898#72561898

### DIFF
--- a/examples/generative/ipynb/vq_vae.ipynb
+++ b/examples/generative/ipynb/vq_vae.ipynb
@@ -53,7 +53,7 @@
    },
    "outputs": [],
    "source": [
-    "!pip install -q tensorflow-probability"
+    "!pip install -q tensorflow-probability==0.13.0"
    ]
   },
   {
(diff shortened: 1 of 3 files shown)


reinvantveer:
> Thanks for the PR. Please also update the `md` and `ipynb` file.
Will do, hang on

> What should we use in a recent TFP version instead?

Good question I have no idea. I agree that it's better to update the example with code that works with the current version. But I'll have to take some time diving into the v.0.14.0 diff to find out, either that or just ask the devs directly. I believe pinning to 0.13.0 solves the immediate block for now.

reinvantveer:
There's an example in https://github.com/tensorflow/probability/blob/main/tensorflow_probability/examples/vq_vae.py but the implementation diverges too much from the Keras VQVAE example to make any sense. I'll open an issue and ask how the offending Categorical(...).sample() should be handled in 0.14.0+

PR #913, opened 2022-06-13
Clarify why TextVectorization works on CPU

8bitmp3:
As discussed with @fchollet @mattdangerw @MarkDaoust 

This PR introduces the following improvements:

- Clarify that the Keras `TextVectorization` layer should be used as part of a `tf.data` input pipeline (runs only on CPUs, won't utilize hardware acceleration if included inside a `Model`, affecting performance).
- Perform some minor linting.

---

- Guide: Working with preprocessing layers (https://keras.io/guides/preprocessing_layers/, https://www.tensorflow.org/guide/keras/preprocessing_layers)
- Section: Preprocessing data before the model or inside the model

### DIFF
--- a/guides/md/preprocessing_layers.md
+++ b/guides/md/preprocessing_layers.md
@@ -116,6 +116,14 @@ print("Features std: %.2f" % (normalized_data.numpy().std()))
 
 <div class="k-default-codeblock">
 ```
+2022-06-15 15:02:07.223345: W tensorflow/stream_executor/platform/default/dso_loader.cc:64] Could not load dynamic library 'libcudart.so.11.0'; dlerror: libcudart.so.11.0: cannot open shared object file: No such file or directory
+2022-06-15 15:02:07.223381: I tensorflow/stream_executor/cuda/cudart_stub.cc:29] Ignore above cudart dlerror if you do not have a GPU set up on your machine.
+2022-06-15 15:02:20.304033: W tensorflow/stream_executor/platform/default/dso_loader.cc:64] Could not load dynamic library 'libcuda.so.1'; dlerror: libcuda.so.1: cannot open shared object file: No such file or directory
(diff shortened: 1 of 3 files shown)


fchollet (commented):
LGTM, thanks for the PR. Please regenerate the `ipynb` and `md` files.

mattdangerw (approved):
Thanks!

guides/md/preprocessing_layers.md
```diff
@@ -614,6 +638,7 @@ print("Model output:", test_output)
 
 <div class="k-default-codeblock">
 ```
+WARNING:tensorflow:5 out of the last 1567 calls to <function PreprocessingLayer.make_adapt_function.<locals>.adapt_step at 0x7f80ec464a60> triggered tf.function retracing. Tracing is expensive and the excessive number of tracings could be due to (1) creating @tf.function repeatedly in a loop, (2) passing tensors with different shapes, (3) passing Python objects instead of tensors. For (1), please define your @tf.function outside of the loop. For (2), @tf.function has reduce_retracing=True option that can avoid unnecessary retracing. For (3), please refer to https://www.tensorflow.org/guide/function#controlling_retracing and https://www.tensorflow.org/api_docs/python/tf/function for  more details.
```
8bitmp3, line 641: @mattdangerw @fchollet  Note the warnings in the output after regenerating `.md`

guides/md/preprocessing_layers.md
```diff
@@ -116,6 +116,14 @@ print("Features std: %.2f" % (normalized_data.numpy().std()))
 
 <div class="k-default-codeblock">
 ```
+2022-06-15 15:02:07.223345: W tensorflow/stream_executor/platform/default/dso_loader.cc:64] Could not load dynamic library 'libcudart.so.11.0'; dlerror: libcudart.so.11.0: cannot open shared object file: No such file or directory
```
8bitmp3, line 119: @mattdangerw @fchollet Note tha new messages in the output after regerating the Markdown file

fchollet (approved):
We'll remove the warnings in post. Thanks!

PR #915, opened 2022-06-14
Fix characters set inconsistency

AminHP:
The PR will fix the problem explained in [this](https://stackoverflow.com/questions/68153448/how-to-correctly-save-and-load-a-model-with-custom-ctc-layer-keras-example/70731125#70731125) StackOverflow question.


fchollet (commented):
LGTM, please also update the `md` and `ipynb` files.

AminHP:
Updated all files, please check again.

fchollet (approved):

PR #919, opened 2022-06-16
QOL changes to autogen.py

LukeWood:
Allows users to use tab completion for example and guide names.  Should help reduce friction when creating code examples.

Now you can run:
```
python autogen.py add_guide ../guides/keras_cv/...
```

which your shell will  be able to autocomplete


fchollet (approved):
LGTM, thank you!

PR #927, opened 2022-06-21
Add Collaborative Filter HF Links

mindwrapped:
As per the ongoing 🤗 Keras sprint, have updated the Collaborative Filtering example with the HF spaces.

Feel free to let me know any changes that need to be done 👍


fchollet (approved):
Thanks for the PR, LGTM

PR #930, opened 2022-06-22
Added 🤗 model and spaces links for WGAN-GP

IMvision12:
As per the ongoing 🤗 keras sprint, have updated the WGAN-GP example


fchollet (approved):
Thanks for the PR!

PR #932, opened 2022-06-24
Code Example: Denoising diffusion implicit models

beresandras:
Added a detailed example on denoising diffusion generative models, using the Oxford Flowers dataset.

Results gif [here](https://i.imgur.com/gyc8eZX.gif)

### DIFF
--- a/examples/generative/ddim.py
+++ b/examples/generative/ddim.py
@@ -305,7 +305,8 @@ def get_network():
         )
         angular_speeds = 2.0 * math.pi * frequencies
         embeddings = tf.concat(
-            [tf.sin(angular_speeds * x), tf.cos(angular_speeds * x)], axis=3,
+            [tf.sin(angular_speeds * x), tf.cos(angular_speeds * x)],
+            axis=3,
         )
         return embeddings
 
@@ -475,7 +476,7 @@ class DiffusionModel(keras.Model):
 
     def denormalize(self, images):
         # convert the pixel values back to 0-1 range
-        images = self.normalizer.mean + (images * self.normalizer.variance ** 0.5)
+        images = self.normalizer.mean + (images * self.normalizer.variance**0.5)
         return tf.clip_by_value(images, 0.0, 1.0)
 
     def diffusion_schedule(self, diffusion_times):
@@ -500,7 +501,7 @@ class DiffusionModel(keras.Model):
             network = self.ema_network
 
         # predict noise component and calculate the image component using it
-        pred_noises = network([noisy_images, noise_rates ** 2], training=training)
+        pred_noises = network([noisy_images, noise_rates**2], training=training)
         pred_images = (noisy_images - noise_rates * pred_noises) / signal_rates
 
         return pred_noises, pred_images
@@ -616,7 +617,8 @@ class DiffusionModel(keras.Model):
     def plot_images(self, epoch=None, logs=None, num_rows=3, num_cols=6):
         # plot random generated images for visual evaluation of generation quality
         generated_images = self.generate(
-            num_images=num_rows * num_cols, diffusion_steps=plot_diffusion_steps,
+            num_images=num_rows * num_cols,
+            diffusion_steps=plot_diffusion_steps,
         )
 
         plt.figure(figsize=(num_cols * 2.0, num_rows * 2.0))
--- a/examples/generative/ddim.py
+++ b/examples/generative/ddim.py
@@ -1,0 +1,838 @@
+"""
+Title: Denoising Diffusion Implicit Models
+Author: [András Béres](https://www.linkedin.com/in/andras-beres-789190210)
+Date created: 2022/06/24
+Last modified: 2022/06/24
+Description: Generating images with denoising diffusion implicit models using the
+Oxford Flowers dataset.
+"""
+
+"""
+## Introduction
+
+### What are diffusion models?
+
+Recently, [denoising diffusion models](https://arxiv.org/abs/2006.11239), including
+[score-based generative models](https://arxiv.org/abs/1907.05600), gained popularity as a
+powerful class of generative models, that can [rival](https://arxiv.org/abs/2105.05233)
+even [generative adversarial networks (GANs)](https://arxiv.org/abs/1406.2661) in image
+synthesis quality. They tend to generate more diverse samples, while being stable to
+train as easy to scale. One of their drawbacks is however, that they are slower to sample
+from, because they require multiple forward passes for generating an image.
+
+Diffusion refers to the process of turning a structured signal (an image) into noise
+step-by-step. By simulating diffusion, we can generate noisy images from our training
+images, and can train a neural network to try to denoise them. Using the trained network
+we can simulate the opposite of diffusion, reverse diffusion, which is the process of an
+image emerging from noise.
+
+![diffusion process gif](https://i.imgur.com/rZZTgnJ.gif)
+
+My one-sentence summary: **diffusion models are trained to denoise noisy images, and can
+generate images by iteratively denoising pure noise.**
+
+### Goal of this example
+
+This code example intends to be a minimal but feature-complete (with a generation quality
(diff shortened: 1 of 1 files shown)


examples/generative/ddim.py
```diff
@@ -0,0 +1,839 @@
...
+
+        # KID is estimated per batch and is averaged across batches
+        self.kid_tracker = keras.metrics.Mean()
+
+        # a pretrained InceptionV3 is used without its classification layer
+        # transform the pixel values to the 0-255 range, then use the same
+        # preprocessing as during pretraining
+        self.encoder = keras.Sequential(
+            [
+                layers.InputLayer(input_shape=(image_size, image_size, 3)),
```
fchollet, line 194: Prefer using `keras.Input(shape=(image_size, image_size, 3))`

examples/generative/ddim.py
```diff
@@ -0,0 +1,839 @@
...
+                layers.Rescaling(255.0),
+                layers.Resizing(height=kid_image_size, width=kid_image_size),
+                layers.Lambda(keras.applications.inception_v3.preprocess_input),
+                keras.applications.InceptionV3(
```
fchollet, line 198: What motivates the choice of InceptionV3 (which is quite heavy)?

examples/generative/ddim.py
```diff
@@ -0,0 +1,839 @@
...
+* We disable the learnable center and scale parameters of the batch normalization layers,
+since the following convolution layers make them redundant.
+* We initialize the last convolution's kernel to all zeros as a good practice, making the
+network predict only zeros after initialization, which is the mean of its targets. This
+will improve behaviour at the start of training and make the mean squared error loss
+start at exactly 1.
+"""
+
+
+def get_network():
```
fchollet, line 297: Following the closure idea, it would be more consistent to give this is a capitalized name

examples/generative/ddim.py
```diff
@@ -0,0 +1,839 @@
...
+            x, skips = x
+            x = layers.UpSampling2D(size=2, interpolation="bilinear")(x)
+            for _ in range(block_depth):
+                x = layers.Concatenate()([x, skips.pop()])
+                x = ResidualBlock(width)(x)
+            return x
+
+        return apply
+
+    noisy_images = keras.Input(shape=(image_size, image_size, 3))
```
fchollet, line 353: image_size seems like it should be a constructor argument.

examples/generative/ddim.py
```diff
@@ -0,0 +1,839 @@
...
+
+
+"""
+## Training
+"""
+
+# create and compile the model
+model = DiffusionModel()
+model.compile(
+    optimizer=tfa.optimizers.AdamW(
```
fchollet, line 644: Prefer using `keras.optimizers.experimental.AdamW` (which will soon be `keras.optimizers.AdamW`, in 2.10)

examples/generative/ddim.py
```diff
@@ -0,0 +1,839 @@
...
+        plt.close()
+
+
+"""
+## Training
+"""
+
+# create and compile the model
+model = DiffusionModel()
+model.compile(
```
fchollet, line 643: Random question -- do you get better performance if you add `jit_compile=True`?

fchollet (commented):
Thank you for the contribution! We definitely needed a diffusion model example. It is beautifully written and very readable 👍

examples/generative/ddim.py
```diff
@@ -0,0 +1,839 @@
...
+
+"""
+
+
+class KID(keras.metrics.Metric):
+    def __init__(self, name="kid", **kwargs):
+        super().__init__(name=name, **kwargs)
+
+        # KID is estimated per batch and is averaged across batches
+        self.kid_tracker = keras.metrics.Mean()
```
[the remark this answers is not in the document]
beresandras: The KID class subclasses keras.metrics.Metric, and is named ("kid"). This mean tracker metric is never logged/shown explicitly. Should I still name it?

examples/generative/ddim.py
beresandras: I get an error:
error message

And a similar one for bilinear upsampling.

examples/generative/ddim.py
beresandras: A smaller network, like a pretrained MobileNet, could probably work as well. But InceptionV3 is what is commonly used in FID and KID metrics, they even have it in their names. I reduce its resolution from 299 to 75 to be lighter. I did the same in a GAN code example: https://github.com/keras-team/keras-io/blob/master/examples/generative/gan_ada.py#L221

examples/generative/ddim.py
beresandras: Fair, but this returns a model object, not a function like the closures.

examples/generative/ddim.py
```diff
@@ -0,0 +1,839 @@
...
+            )
+        )
+        angular_speeds = 2.0 * math.pi * frequencies
+        embeddings = tf.concat(
+            [tf.sin(angular_speeds * x), tf.cos(angular_speeds * x)],
+            axis=3,
+        )
+        return embeddings
+
+    def ResidualBlock(width):
```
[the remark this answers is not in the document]
beresandras: Should I put the sinusoidal embedding outside as well?

examples/generative/ddim.py
```diff
@@ -0,0 +1,839 @@
...
+    noise_variances = keras.Input(shape=(1, 1, 1))
+
+    e = layers.Lambda(sinusoidal_embedding)(noise_variances)
+    e = layers.UpSampling2D(size=image_size, interpolation="nearest")(e)
+
+    x = layers.Conv2D(widths[0], kernel_size=1)(noisy_images)
+    x = layers.Concatenate()([x, e])
+
+    skips = []
+    for width in widths[:-1]:
```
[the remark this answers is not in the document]
beresandras: Done, along with image_size and block_depth.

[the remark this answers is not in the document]
beresandras: Should embedding_dims and embedding_max_frequency be constructor arguments as well?

examples/generative/ddim.py
fchollet: Got it -- upsampling is not supported with XLA, at least not the op we use. That's too bad.

examples/generative/ddim.py
```diff
@@ -0,0 +1,839 @@
...
+        # this is computationally demanding, kid_diffusion_steps has to be small
+        images = self.denormalize(images)
+        generated_images = self.generate(
+            num_images=batch_size, diffusion_steps=kid_diffusion_steps
+        )
+        self.kid.update_state(images, generated_images)
+
+        return {m.name: m.result() for m in self.metrics}
+
+    def plot_images(self, epoch=None, logs=None, num_rows=3, num_cols=6):
```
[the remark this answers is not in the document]
fchollet: Ok, sounds good.

examples/generative/ddim.py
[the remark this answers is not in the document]
fchollet: I'd say that's good practice -- maybe just `"tracker"` or something.

examples/generative/ddim.py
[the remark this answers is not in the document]
beresandras: I decided not to put the embedding hyperparams into the constructor, because they are only needed for the Lambda layer for which it is better if they are global.

examples/generative/ddim.py
beresandras: Added note on using AdamW from tensorflow_addons for tensorflow versions below 2.9

beresandras:
Huggingface demo links added as well.

I believe I am ready, should I update anything else?

fchollet (approved):
Looks good! Please add the generated files.

fchollet:

fchollet:
Oh, I see the generated files weren't included yet. Please reopen a PR to add them. Sorry about that!

beresandras:
No problem! I added the generated files in #962

PR #940, opened 2022-06-27
Add Node2Vec Movielens HF links

bp-high:
As per the ongoing 🤗 Keras sprint, have updated the Node2Vec Movielens ( Graph representation learning with node2vec) example with the HF spaces.

Feel free to let me know any changes that need to be done 👍


fchollet (approved):
LGTM, thank you for the PR

PR #941, opened 2022-06-28
Added huggingface keras siamese contrastive v2

rushic24:
As per the ongoing hugs Keras sprint, have updated the keras siamese contrastive notebooks for hugging face models


fchollet (approved):
LGTM, thanks

PR #943, opened 2022-06-28
Update serialization_and_saving.md

synandi:
Fixed spellings of Tensorflow and definitions.


fchollet (commented):
Thanks for the PR. Please also fix the `py` and `ipynb` files.

pcoet:
@synandi Thanks for the PR. It looks like this is fixed now, so I'll close.

PR #956, opened 2022-07-01
Open Ended Text Generation Guide KerasNLP

aflah02:
This guide compares byte and unicode tokenizers for text generation
@mattdangerw Raised a PR based on our discussion, it's open for review now!
Referring to the README, I will add the generated files after approval.
Reference: https://github.com/keras-team/keras-nlp/issues/191

### DIFF
--- a/examples/nlp/character_and_byte_level_text_generation.py
+++ b/examples/nlp/character_and_byte_level_text_generation.py
@@ -1,0 +1,284 @@
+"""
+Title: Character and byte level text generation with KerasNLP
+Author: [Mohammad Aflah Khan](https://github.com/aflah02)
+Date created: 2022/07/01
+Last modified: 2022/07/01
+Description: Use KerasNLP to train an open ended text generation LSTM model.
+"""
+
+"""
+KerasNLP provides lots of building blocks for NLP (model layers, tokenizers, metrics, etc.) and
+makes it convenient to construct NLP pipelines on the fly.
+
+In this tutorial we'll use KerasNLP's `UnicodeCharacterTokenizer` and `ByteTokenizer` to train a 
+character level LSTM model for open ended text generation. This example draws inspiration from 
+[Character-level recurrent sequence-to-sequence model example](https://keras.io/examples/generative/lstm_character_level_text_generation/) 
+by [fchollet](https://twitter.com/fchollet) for it's model architecture and uses 
+[Don Quijote by Miguel de Cervantes Saavedra (Spanish Version)](https://www.gutenberg.org/ebooks/2000) as a data source
+
+This tutorial broadly covers the following:
+- Tokenization using `keras_nlp.tokenizers.UnicodeCharacterTokenizer` and `keras_nlp.tokenizers.ByteTokenizer` to obtain character level tokens.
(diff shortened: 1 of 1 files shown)


fchollet (commented):
Thanks for the PR! Some preliminary comments for now.

@mattdangerw please take a look at the example as well.

examples/nlp/character_and_byte_level_text_generation.py
```diff
@@ -0,0 +1,285 @@
...
+
+"""
+Let's see how we can detokenize the text!
+"""
+
+print(UnicodeCharacterTokenizer.detokenize(unicode_tokenized_text[:50]))
+print(ByteTokenizer.detokenize(byte_tokenized_text[:50]))
+
+"""
+We can see post tokenization the datasets now differ which is to be expected as byte tokenization and unicode
```
mattdangerw, line 87: maybe just phrase this as

Let's try tokenizing the the input string "señora mía".

print(codepoint_tokenizer("señora mía))
print(byte_tokenizer("señora mía))

This captures the tradeoff between the two tokenizer. The byte tokenizer will handle any text with only 256 output ids, at the cost of encoding to a longer sequence. The unicode tokenizer will produce shorted sequence, at the cost of a larger id space (and larger embeddings).

examples/nlp/character_and_byte_level_text_generation.py
```diff
@@ -0,0 +1,285 @@
...
+
+"""
+EXAMPLE 1
+PROMPT:   que ya su dulce esposo no vivía, rompió los aires con suspiros, hirió los cielos con quejas, maltra
+
+Model: model_unicode
+OUTPUT:   que ya su dulce esposo no vivía, rompió los aires con suspiros, hirió los cielos con quejas, maltrae. Ɏ y el como son prés como sancho de la carón
+
+Model: model_byte
+OUTPUT:   que ya su dulce esposo no vivía, rompió los aires con suspiros, hirió los cielos con quejas, maltra de su don quijoteâ, que su perreso de su cab
```
mattdangerw, line 249: We need to keep looking at where these erroneous characters are coming from `â`, maybe we need a longer dataset? Different hyperparameters? Maybe trying the transfomer will help?

Especially with greed search, we should at least be able to train to the point where we output a valid set of characters that are actually in the source text.

mattdangerw (changes requested):
Thanks! Left some initial comments

examples/nlp/character_and_byte_level_text_generation.py
```diff
@@ -0,0 +1,285 @@
...
+- Decoding using `keras_nlp.utils.greedy_search` utility
+
+This tutorial will be pretty useful and will be a good starting point for learning about KerasNLP and how to 
+incorporate it into your own NLP pipelines.
+"""
+
+"""
+## Setup
+
+Let's do our imports first
```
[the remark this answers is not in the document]
aflah02: Got it!

aflah02:
@pcoet Sorry I'm blocked due to other commitments and would be unable to finish this PR in the near future. Closing it for now then

PR #1001, opened 2022-07-26
Change `kt` to `keras_tuner` in imports

haifeng-jin:
This fix the one I missed last time.


fchollet (approved):
Thanks for the PR. LGTM.

Please do not create additional branches on the repo; use a separate fork for this purpose.

PR #1017, opened 2022-08-04
updated keras-cv version

shivalikasingh95:
updating keras-cv version to avoid docker-image check from failing


fchollet (approved):
LGTM, thanks

PR #1023, opened 2022-08-12
Corrects node size

yutsai84:
The existing implementation assumed the max of `node_indicies` is the size of the nodes. 

The assumption is not alway true. For example, if the largest node index of the nodes is not in  `node_indicies`, this assumption is wrong.

This PR addresses this by directly checking the size of nodes. That is the size of the nodes is the row number of `node_representations`.


fchollet (commented):
Thanks for the PR. The change looks good to me. Please replicate the fix in the md and ipynb files.

yutsai84:
> Thanks for the PR. The change looks good to me. Please replicate the fix in the md and ipynb files.

@fchollet . Thanks for the review. I addressed your comment by adding the same change to ipynb files as well.

fchollet (approved):
Thanks!

PR #1026, opened 2022-08-14
added print(translated)

ayyucedemirbas:
You had forgotten to print the translated text


fchollet (commented):
Thanks for the PR! The reason for not having the `print()` statement in the code is that the output would get rendered directly on the website. In our case we prefer selecting a few examples to show instead of letting the notebook randomize it when it gets rendered. I guess we could just set the random seed to avoid the randomization issue...

ayyucedemirbas:
Thank you so much for your kind and helpful reply.

PR #1050, opened 2022-09-01
Update siamese contrastive example to match label definition from paper

hertschuh:


hertschuh:
Context: keras-team/keras#16935

fchollet (approved):
LGTM

PR #1056, opened 2022-09-06
Add matplotlib in requirements as it is used in many examples.

hertschuh:
Also converted all newlines as there was a mix of Unix and Windows style newlines.


fchollet (approved):
LGTM

PR #1058, opened 2022-09-07
For better understanding of Siamese Network

stsfaroz:
included the convolutional model in a function  and included the siamese model in a function, inside the siamese function calling the conv_function, so that clearly everyone can understand the flow of the siamese model


fchollet (commented):
Thanks for the PR!

This is a style-only refactoring PR, which we usually don't merge. It also features Python style issues unfortunately.

PR #1059, opened 2022-09-08
Tf text 2.10 is out, we can stop pinning

mattdangerw:
We don't need to explicitly pull in the pre-release anymore.


fchollet (approved):
Thanks!

PR #1066, opened 2022-09-15
Fix broken link of Caltech 101 Dataset

gadagashwini:


fchollet (commented):
Thanks for the PR! Please also replicate the fix in the md and ipynb files.

PR #1089, opened 2022-09-26
Add pip install KerasCV to StableDiffusion guide

LukeWood:
needs a re-render, hold off on merging; will run on my A100 after a run of the RetinaNet finishes


LukeWood:
Bad merge

PR #1101, opened 2022-09-30
Fix broken link of tf.data.Dataset

gadagashwini:


fchollet (commented):
Thanks -- please replicate the fix in the md and ipynb files.

PR #1102, opened 2022-10-02
Update gif links for SD random walks example

ianstenbit:
Panda to plane I had to downscale to a factor of 0.5

The other two I downscaled to 0.7, as they weren't as big.


fchollet (approved):
Thank you!

PR #1126, opened 2022-11-01
Add a new example

emirhanai:
Hello,
I wanted to add a software related to Quantum6 technology. In this way, I have used Quantum algorithms for the first time in the Vision section of Keras. Additionally, the last two commits are correct. The first 3 commits were uploaded incorrectly. For this reason, I would like to delete the first 3 downloads and save the last two.
Thank you.


artificialintelligence4994:
Thank you @emirhanai . Quantum Technology is very newest and innovative. 6G is very speed and useful with AI vision!

fchollet (commented):
Thanks for the PR!

This is very interesting work, but quantum ML is very niche at this time, and isn't a good fit for keras.io. Perhaps you could instead reach out to the TFQ folks to try to see if they'd like to include this example (or something close) in their docs?

PR #1133, opened 2022-11-11
Change from categorical to binary accuracy

cumbalik:
As communicated with @sayakpaul, for multilabel classifications, categorical accuracy is, unfortunately, a not good choice. It works only for “real” one-hot-encodings, i.e., one-hot-encoding where “1” appears exactly once and the remaining entries are “0”. You can see it also in the implementation of the [categorical accuracy]( https://github.com/keras-team/keras/blob/b80dd12da9c0bc3f569eca3455e77762cf2ee8ef/keras/metrics/metrics.py#L286) the `tf.math.argmax` returns exactly one index of the first largest number.

The closest alternative to categorical accuracy is binary accuracy. Therefore, I adjusted the code so that binary accuracy could be used, and I updated the metric result.

Lastly, binary accuracy is generally not a good choice for multilabel tasks since the results can be misleading. For multilabel tasks, most of the labels in one-hot-encoding will be zero and the model will also predict mostly zeros; thus, you can easily achieve accuacies around 95%. I would suggest using metrics based on the confusion matrix for multilabel tasks.


sayakpaul:
Thanks for adding this change. 

> Lastly, binary accuracy is generally not a good choice for multilabel tasks since the results can be misleading. For multilabel tasks, most of the labels in one-hot-encoding will be zero and the model will also predict mostly zeros; thus, you can easily achieve accuacies around 95%. I would suggest using metrics based on the confusion matrix for multilabel tasks.

What if we could compute a masked accuracy for non-zero values? Happy to go with the confusion matrix too. Feel free to contribute that. We'll be more than happy to give you the acknowledgement for it.

cumbalik:
Thanks for your quick response.

A masked accuracy is an option; however, I'm not sure if there is an implementation in Keras. I know in Scikit-Learn, the "classical accuracy", if applied to multi-label datasets, behaves like masked accuracy, i.e. see the screenshot below
grafik
Source: https://www.kaggle.com/code/kmkarakaya/multi-label-model-evaluation

The fact is, if you assume that masked accuracy is an appropriate metric for your multi-label multi-classification problem. You, in the end, decide that each sample can be assigned a "new label" out of 2^(num_of_labels). Maybe an example is more clear what I mean:

Assume a book can be labeled with the following four labels: L = {l1, l2, l3, l4}. If you use a masked accuracy, you generate a "new label"  set that will contain all possible combinations out of L. So the label set will have 2^4=16 labels. 

The bottom line is that you transform your multi-label multi-classification problem into a multi-classification problem with 2^(num_of_labels) new classes. This transformation can be reasonable for some cases, but it only works if you have at least one sample for each class out of 2^(num_of_labels).

I think it is more reasonable not to use it because the labels, in our case, are more or less independent from each other.
If you agree, then I am happy to extend the script by applying the confusion matrix-based metrics.

sayakpaul:
> I think it is more reasonable not to use it because the labels, in our case, are more or less independent from each other.

Could you elaborate? 

In retrospect, I think it's reasonable not to add confusion matrix to keep the example lightweight. 
Also, if there's a no off the shelf support for something in the library, we usually implement it for the tutorial depending on the value.

cumbalik:
> > I think it is more reasonable not to use it because the labels, in our case, are more or less independent from each other.
> 
> Could you elaborate?
Yes, sure. 
I will explain using an example:

Suppose a data set that contains these label combinations:
{[cs.CV, stat.ML,cs.CL], [cs.CR, cs.CL], [eess.SP, cs.SI, math.IT, stat.AP], [cs.CR, cs.CL]], [cs.CV, stat.ML,cs.CL], [eess.SP, cs.SI, cs.CR ]}.

Now, the data will be split into train and test labels:
Train-labels: {[cs.CV, stat.ML, cs.CL], [cs.CR, cs.CL], [eess.SP, cs.SI, math.IT, stat.AP]}
Test-labels: {[cs.CR, cs.CL], [cs.CV, stat.ML, stat.AP], [eess.SP, cs.SI, cs.CR]}

As next, we train a classifier on the training dataset.

Now, we feed the classifier with the test sample with labels [cs.CR, cs.CL]. Since our train dataset contained a sample with this label, there is a good chance that the classifier predicted [cs.CR, cs.CL]. Therefore, we assume that the model predicted this label combination. So, we have a first correct prediction based on “classical” accuracy.

We feed the classifier with the test sample with label [cs.CV, stat.ML, stat.AP]. Suppose this sample is similar to the training sample with label [cs.CV, stat.ML, cs.CL]. Thus our classifier predicts the latter labels. Although the classifier predicted some labels correctly (cs.CV, stat.ML), based on the classical accuracy, it’s a wrong prediction.

Similarly to the last test label, if we assume, the classifier predicted these labels correctly [eess.SP, cs.SI], it will be anyway wrongly predicted based on the classical accuracy.

In the end, the classical accuracy on the test data is 1/3. And this can be a misleading metric, especially if the classifier predicted for each label combination [cs.CV, stat.ML, stat.AP], [eess.SP, cs.SI, cs.CR] 2 out of 3 labels correctly.

To conclude, if it is assumed that the labels are more or less independent on each other, then is more suitable to use also a metric that better reflexes the individual label prediction, e.g. like above that 2 out of 3 labels are correctly predicted.

For this reason, I suggest using, additionally, the [f1 metric](https://www.tensorflow.org/addons/api_docs/python/tfa/metrics/F1Score) to extend the interpretation of the accuracy of the trained model.

sayakpaul:
Amazing, thanks for explaining. Feel free to link the readers to your comment above from the edited text.

sayakpaul:
@fchollet @mattdangerw could you review the PR and approve?

examples/nlp/multi_label_classification.py
```diff
@@ -409,4 +417,7 @@ def plot_result(item):
 We would like to thank [Matt Watson](https://github.com/mattdangerw) for helping us
 tackle the multi-label binarization part and inverse-transforming the processed labels
 to the original form.
+
+Thanks [Cingis Kratochvil](https://github.com/cumbalik) for suggesting and extending
+this code example by the binary accuracy.
```
sayakpaul, line 422:
```suggestion
Thanks to [Cingis Kratochvil](https://github.com/cumbalik) for suggesting and extending this code example by introducing binary accuracy as the evaluation metric.
```

sayakpaul (approved):
Thanks so much.

fchollet (approved):
Thanks! Please update the generated `md` file and `ipynb` file.

cumbalik:
> Thanks! Please update the generated `md` file and `ipynb` file.

Thanks you, I've just updated the generated md, ipynb, and img files.

fchollet (approved):
Thanks!

PR #1152, opened 2022-12-10
Update "tokens" -> "token_ids"

jbischof:
Output dictionary names changed in https://github.com/keras-team/keras-nlp/pull/327, but guide never updated

TESTED=ran guide locally

### DIFF
--- a/guides/keras_nlp/transformer_pretraining.py
+++ b/guides/keras_nlp/transformer_pretraining.py
@@ -251,7 +251,7 @@ def preprocess(inputs):
     # Split the masking layer outputs into a (features, labels, and weights)
     # tuple that we can use with keras.Model.fit().
     features = {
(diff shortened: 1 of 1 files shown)


ianstenbit (approved):

MarkDaoust (changes requested):
Seems uncontroversal.

Please run:

```
cd scripts
python autogen.py add_guide/transformer_pretraining
```

You update downstream files.

mattdangerw:
@jbischof actually we should be careful here. I believe that change is only on our master branch, and we want to keep this guide in a state that will work with `!pip install keras-nlp` until we do our next release.

I opened https://github.com/keras-team/keras-nlp/issues/508 to track issue like this.

We should make sure not to actually land this PR until we are pushing out our new pypi release.

fchollet (approved):
LGTM

PR #1155, opened 2022-12-12
Revert "Update "tokens" -> "token_ids""

mattdangerw:
Reverts keras-team/keras-io#1152

We don't actually want to ship this until we release KerasNLP 0.4 to pip.


jbischof (approved):
Luckily this doesn't hurt anything because I never ran `autogen`, but agree that rollback makes sense.

MarkDaoust (approved):

PR #1159, opened 2022-12-16
Add "Getting Started" guide for KerasNLP

jbischof:


guides/keras_nlp/quick_tour.py
```diff
@@ -0,0 +1,610 @@
...
+
+
+"""
+
+"""
+# API quickstart
+
+Our highest level API is `keras_nlp.models`. For each `XX` architecture (e.g., `Bert`),
+we offer the following modules:
+* **Tokenizer**: `keras_nlp.models.XXTokenizer`
```
mattdangerw, line 58: Looks like you need a newline before the list for this to render right.

guides/keras_nlp/quick_tour.py
```diff
@@ -0,0 +1,610 @@
...
+preset weights and architectures when used out-of-the-box and are easily customizable
+when more control is needed.
+
+This library is an extension of the core `keras` API; all high level modules are `Layers`
+or `Models`. If you are familiar with `keras`, congratulations! You already understand
+most of `keras-nlp`.
+
+This guide demonstrates our modular approach using a sentiment analysis example at six
+levels of complexity:
+* Inference with a pretrained classifier
```
mattdangerw, line 38: same here need a newline

guides/keras_nlp/quick_tour.py
```diff
@@ -0,0 +1,610 @@
...
+import keras_nlp
+import tensorflow as tf
+from tensorflow import keras
+import tensorflow_datasets as tfds
+
+# Use mixed precision for optimal performance
+keras.mixed_precision.set_global_policy("mixed_float16")
+
+"""
+# KerasNLP: Modular NLP Workflows for Keras
```
mattdangerw, line 24: This title can be removed or we get a double title with line 2.

guides/keras_nlp/quick_tour.py
```diff
@@ -0,0 +1,610 @@
...
+Throughout our guide we use Professor Keras, the official Keras mascot, as a visual
+reference for the complexity of the material:
+
+![picture](https://drive.google.com/uc?id=1d14Qpmfgjf6zu4z30HBaonH8PYDHgVoU)
+
+
+"""
+
+"""
+# API quickstart
```
mattdangerw, line 54: Every heading from here and below should be one level lower. E.g. `#` -> `##`, and `##` -> `###`.

mattdangerw (commented):
Just leaving some random rendering comments I noticed when generating this.

guides/keras_nlp/quick_tour.py
```diff
@@ -0,0 +1,610 @@
...
+Date created: 2022-12-15
+Last modified: 2022-12-15
+Description: An introduction to the KerasNLP API.
+"""
+"""
+<a
+href="https://colab.research.google.com/github/jbischof/keras-io/blob/quickstart/guides/ke
+ras_nlp/keras_nlp_quick_tour.ipynb" target="_parent"><img
+src="https://colab.research.google.com/assets/colab-badge.svg" alt="Open In Colab"/></a>
+"""
```
fchollet, line 13: Remove this block, the links get autogenerated.

guides/keras_nlp/quick_tour.py
```diff
@@ -0,0 +1,610 @@
...
+We will use a running example of sentiment analysis of IMDB movie reviews. In this task,
+we use the text to predict whether the review was positive (`label = 1`) or negative
+(`label = 0`).
+
+We load the data from `tensorflow_datasets`, a collection of machine learning benchmarks
+that uses the powerful `tf.data.Dataset` format for examples.
+"""
+
+BATCH_SIZE = 16
+imdb_train, imdb_test = tfds.load(
```
fchollet, line 94: I'd encourage you to start from raw text files rather than TFDS, because it is closer to the kind of workflows that users go through. It only takes a few lines, and these few lines will better generalize to users' own datasets. https://keras.io/examples/nlp/text_classification_from_scratch/#load-the-data-imdb-movie-review-sentiment-classification

guides/keras_nlp/quick_tour.py
```diff
@@ -0,0 +1,610 @@
...
+    "imdb_reviews",
+    split=["train", "test"],
+    as_supervised=True,
+    batch_size=BATCH_SIZE,
+)
+
+# Inspect first review
+# Format is (review text tensor, label tensor)
+imdb_train.unbatch().take(1).get_single_element()
```
fchollet, line 103: Use `print()` for anything you want to display. The file should also work as a standalone Python script.

guides/keras_nlp/quick_tour.py
```diff
@@ -0,0 +1,610 @@
...
+outputs = keras_nlp.layers.TokenAndPositionEmbedding(
+    vocabulary_size=len(vocab),
+    sequence_length=packer.sequence_length,
+    embedding_dim=64,
+)(token_id_input)
+outputs = keras_nlp.layers.TransformerEncoder(
+    num_heads=2,
+    intermediate_dim=128,
+    dropout=0.1,
+)(outputs, padding_mask=padding_mask)
```
fchollet, line 577: `TokenAndPositionEmbedding` and `TransformerEncoder` aren't mask aware? I'd expect to be able to do

```python
token_id_input = keras.Input(
    shape=(None,),
    dtype="int32",
    name="token_ids",
)
outputs = keras_nlp.layers.TokenAndPositionEmbedding(
    vocabulary_size=len(vocab),
    sequence_length=packer.sequence_length,
    embedding_dim=64,
    mask_zero=True,
)(token_id_input)
outputs = keras_nlp.layers.TransformerEncoder(
    num_heads=2,
    intermediate_dim=128,
    dropout=0.1,
)(outputs)
```

(assuming input sequences have zeros at padded locations)

guides/keras_nlp/quick_tour.py
```diff
@@ -0,0 +1,610 @@
...
+    jit_compile=True,
+)
+model.fit(
+    imdb_preproc_train_ds,
+    validation_data=imdb_preproc_val_ds,
+    epochs=3,
+)
+
+"""
+While our classification accuracy is a fairly poor 0.76, the transformer architecture is
```
fchollet, line 607: With a tiny Transformer on IMDB you should be able to do close to 0.9. The (basically untuned) hyperparameters in my book (ch 11) will get you to 0.875.

guides/keras_nlp/quick_tour.py
```diff
@@ -0,0 +1,610 @@
...
+`from_preset` constructor and can be composed with custom `Layer`s. Detailed examples can
+be found at https://keras.io/guides/transfer_learning/.
+
+A **backbone** `Model` does not include automatic preprocessing but can be paired with a
+matching **preprocessor** using the same **preset** as shown in the previous workflow.
+
+In this workflow we experiment with freezing our backbone model and adding two trainable
+transfomer layers to adapt to the new input.
+
+**Note**: We can igonore the warning about gradients for the `pooled_dense` layer because
```
fchollet, line 312: ignore

guides/keras_nlp/quick_tour.py
```diff
@@ -0,0 +1,610 @@
...
+"""
+
+"""
+## Custom preprocessing
+
+In cases where custom preprocessing is required, we offer direct access to the
+`Tokenizer` class that maps raw strings to tokens. It also has a `from_preset`
+constructor to get the vocabulary matching pretraining.
+
+**Note:** `Tokenizer` does not pad sequences, so output is `tf.RaggedTensor`.
```
fchollet, line 261: its output is a

fchollet (commented):
Thanks for the PR!

guides/keras_nlp/quick_tour.py
mattdangerw: They are mast aware. We could do this and it would simplify the code a bit.

guides/keras_nlp/quick_tour.py
jbischof: Thanks, updated! I had been optimizing for runtime, not performance. Increasing `sequence_length` from 64 to 512 matches your performance.

guides/keras_nlp/quick_tour.py
mattdangerw: I made these changes locally before pushing the preview. So that is what it looks like after the change. Before looked pretty wonky.

guides/keras_nlp/quick_tour.py
mattdangerw: To me it looks kinda odd as rendered right now. Both this (which now renders as a section heading), and the top actual header read like titles for the whole guide. I would probably just rename this "Introduction", which would fit with [other guides](https://keras.io/guides/distributed_training/). Could also just ditch the second "KerasNLP", but "Introduction" would read clearer to me.

image

guides/keras_nlp/getting_started.py
```diff
@@ -0,0 +1,619 @@
+"""
```
mattdangerw, line 1: You will need to add this as the first entry in `children` here, so this populated in the guides section as a link under KerasNLP.

https://github.com/keras-team/keras-io/blob/3c5d6ec6b334ce0ab806dd874feca68fb95bfa6f/scripts/guides_master.py#L26

guides/keras_nlp/getting_started.py
```diff
@@ -0,0 +1,619 @@
...
+* Fine tuning a pretrained backbone
+* Fine tuning with user-controlled preprocessing
+* Fine tuning a custom model
+* Pretraining a backbone model
+* Build and train your own transformer from scratch
+
+Throughout our guide we use Professor Keras, the official Keras mascot, as a visual
+reference for the complexity of the material:
+
+![picture](https://drive.google.com/uc?id=1d14Qpmfgjf6zu4z30HBaonH8PYDHgVoU)
```
mattdangerw, line 43: Before we commit, we should download these in `theme/img/` and reference them here. We probably don't want everyone on keras.io hitting your google drive.

guides/keras_nlp/getting_started.py
```diff
@@ -0,0 +1,619 @@
...
+Want to implement a novel transformer architecture? The KerasNLP library offers all the
+low-level modules used to build SoTA architectures in our `models` API. This includes the
+`keras_nlp.tokenizers` API which allows you to train your own subword tokenizer using
+`WordPieceTokenizer`, `BytePairTokenizer`, or `SentencePieceTokenizer`.
+
+In this workflow we train a custom tokenizer on the IMDB data and design a backbone with
+custom transformer architecture. For simplicity we then train directly on the
+classification task. Interested in more details? We wrote an entire guide to pretraining
+and finetuning a custom transformer:
+https://keras.io/guides/keras_nlp/transformer_pretraining/
```
mattdangerw, line 523: I think this can be /guides/keras_nlp/transformer_pretraining/. Also consider not using a bare link

guides/keras_nlp/getting_started.py
```diff
@@ -0,0 +1,619 @@
...
+The lower layers of this model are then packaged as a **backbone** to be combined with
+layers relating to a new task.
+
+The KerasNLP library offers SoTA **backbones** and **tokenizers** to be trained from
+scratch without presets.
+
+In this workflow we pretrain a BERT **backbone** using our IMDB review text. We skip the
+"next sentence prediction" (NSP) loss because it adds significant complexity to the data
+processing and was dropped by later models like RoBERTa. See our e2e [BERT pretraining
+example](https://github.com/keras-team/keras-nlp/tree/4f9ebefa82af22b4f4267dfa80fa525f7a03
```
mattdangerw, line 401: I would actually recommend using the non-perma-link to https://github.com/keras-team/keras-nlp/tree/master/examples/bert (and we could update it as needed). Sending people to a frozen in spot time of our repo seems like it could be confusing.

guides/keras_nlp/getting_started.py
```diff
@@ -0,0 +1,619 @@
...
+improve performance. If we want to predict IMDB review sentiment, using IMDB data should
+perform better than Rotten Tomatoes data! And for many tasks no relevant pretrained model
+will be available (e.g., categorizing customer reviews).
+
+The workflow for fine-tuning is almost identical to above, except that we request a
+**preset** for the **backbone**-only model rather than the entire classifier. When passed
+a **backone** **preset**, a **task** `Model` will randomly initialize all task-specific
+layers in preparation for training. For all the **backbone** presets available for
+`BertClassifier`, see
+[keras.io](https://resilient-dango-43f7b8.netlify.app/api/keras_nlp/models/).
```
mattdangerw, line 187: /api/keras_nlp/models/

guides/keras_nlp/getting_started.py
```diff
@@ -0,0 +1,619 @@
...
+## Introduction
+
+
+KerasNLP is a natural language processing library that supports users through their
+entire development cycle. Our workflows are built from modular components that have SoTA
+preset weights and architectures when used out-of-the-box and are easily customizable
+when more control is needed.
+
+This library is an extension of the core Keras API; all high level modules are
+[`Layers`](https://keras.io/api/layers/) or
```
mattdangerw, line 26: /api/layers/ and /api/models/

guides/keras_nlp/getting_started.py
```diff
@@ -0,0 +1,619 @@
...
+"""
+All **tasks** have a `from_preset` method that constructs a `keras.Model` instance with
+preset preprocessing, architecture and weights. This means that we can pass raw strings
+in any format accepted by a `keras.Model` and get output specific to our task.
+
+This particular **preset** is a `"bert_tiny_uncased_en"` **backbone** fine-tuned on
+`sst2`, another movie review sentiment analysis (this time from Rotten Tomatoes). We use
+the `tiny` architecture for demo purposes, but larger models are recommended for SoTA
+performance. For all the task-specific presets available for `BertClassifier`, see
+[keras.io](https://resilient-dango-43f7b8.netlify.app/api/keras_nlp/models/).
```
mattdangerw, line 152: /api/keras_nlp/models/

mattdangerw (changes requested):
Just did a pass on some of the hyperlinks!

guides/keras_nlp/getting_started.py
jbischof: I think they should be in `guides/img/` right?

fchollet (commented):
The images appear to be pixellated (the text below them especially). Can you generate clean ones? Also, no need to add them to the git tree (in general it should be avoided since it increases the size of the repo, which we're constantly downloading), you can put them on imgur.

jbischof:
@fchollet imgur seems a little weird. What if we used our GCP bucket? I don't have access to our the keras-io bucket but here is an example upload to a test bucket ([link](https://storage.googleapis.com/keras-nlp-test/prof_keras_labeled.png)).

guides/keras_nlp/getting_started.py
```diff
@@ -0,0 +1,618 @@
...
+Last modified: 2022-12-15
+Description: An introduction to the KerasNLP API.
+"""
+
+import keras_nlp
+import tensorflow as tf
+from tensorflow import keras
+
+# Use mixed precision for optimal performance
+keras.mixed_precision.set_global_policy("mixed_float16")
```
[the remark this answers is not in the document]
jbischof: No clear opinion here. I originally tried to teach about the TF ecosystem as well since it's a selling point of our library.

fchollet (approved):
Looking good! Please add the generated files.

Please note that I've pushed minor copyedits (in particular, using absolute URLs for links, which is necessary since many folks use the tutorial interactively via the Colab link where relative URLs wouldn't work). Pull them first.

mattdangerw:
> (in particular, using absolute URLs for links, which is necessary since many folks use the tutorial interactively via the Colab link where relative URLs wouldn't work)

Oops, my bad! I asked for this. Thought relative would be better for previews, but the colab issue makes sense.

mattdangerw (approved):

innat:
What a well-written guide! Love it. 
Thanks @jbischof  👍

PR #1176, opened 2022-12-28
Add a tutorial on fine-tuning Stable Diffusion on a custom dataset

sayakpaul:
Colab Notebook: https://colab.research.google.com/drive/1BICyJVVtELRRX1EiiXGzKQlOGyMyklbx?usp=sharing

Done in collaboration with @deep-diver.


sayakpaul:
@deep-diver I guess you need to sign the CLA 'cause mine's already signed.

sayakpaul:
I guess the reason due to which the CI for Docker image is failing is unrelated to the PR.

examples/generative/finetune_stable_diffusion.py
```diff
@@ -0,0 +1,438 @@
...
+        self.vae.trainable = False
+
+    def train_step(self, inputs):
+        images = inputs["images"]
+        encoded_text = inputs["encoded_text"]
+        bsz = tf.shape(images)[0]
+
+        with tf.GradientTape() as tape:
+            # Project image into the latent space.
+            latents = self.vae(images, training=False)
```
sayakpaul, line 268: We should ideally sample from the variance and mean predicted by the VAE instead of taking its final outputs directly. We're currently experimenting with it ([code branch](https://github.com/sayakpaul/stable-diffusion-keras-ft/tree/feat/sampler)) and will provide an update as soon as we have some results. But I guess it doesn't block the reviewing. 

Cc: @deep-diver

tanzhenyu (commented):
Thanks for the PR! Looking good

examples/generative/finetune_stable_diffusion.py
sayakpaul: This has been addressed. 

Since we're training the model from scratch, it'll take some time to get to an epoch checkpoint producing plausible image quality. Once we have the results, we'll update the PR with those.

sayakpaul:
@fchollet we could only fine-tune for 20 epochs and we are limited by the resources here. 

The results will likely improve if we fine-tune it for at least 70 epochs. 

Could someone on the Keras team spin up an 80GB A100 to help us do this?

examples/generative/finetune_stable_diffusion.py
```diff
@@ -0,0 +1,463 @@
...
+        self.vae = vae
+        self.noise_scheduler = noise_scheduler
+        self.max_grad_norm = max_grad_norm
+
+        self.vae.trainable = False
+
+    def train_step(self, inputs):
+        images = inputs["images"]
+        encoded_text = inputs["encoded_text"]
+        bsz = tf.shape(images)[0]
```
fchollet, line 264: Avoid any non standard abbreviations, use fully spelled out variable names whenever possible.

fchollet (commented):
Thanks for the PR! This is great!

> Could someone on the Keras team spin up an 80GB A100 to help us do this?

I think your best bet might be a TPU. We can help here. I actually tried running the example on TPU but ran into a few issues.

If that doesn't work we can try the 80GB A100.

examples/generative/finetune_stable_diffusion.py
```diff
@@ -0,0 +1,463 @@
...
+of brevity, we discarded those elements. More on this later in the tutorial.
+"""
+
+"""
+## Initialize the trainer and compile it
+"""
+
+image_encoder = ImageEncoder(RESOLUTION, RESOLUTION)
+diffusion_ft_trainer = Trainer(
+    diffusion_model=DiffusionModel(RESOLUTION, RESOLUTION, MAX_PROMPT_LENGTH),
```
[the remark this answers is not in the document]
sayakpaul: We actually did this ([here](https://github.com/sayakpaul/stable-diffusion-keras-ft/blob/main/finetune.py#L123) and [here](https://github.com/sayakpaul/stable-diffusion-keras-ft/blob/main/trainer.py#L137)) in the repository. But to keep the code part specific to the fine-tuning, we didn't include the checkpointing code here.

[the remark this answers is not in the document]
sayakpaul: Overriding the `save_weights()` method from the trainer class allows us to work directly with the diffusion model that is being fine-tuned.

examples/generative/finetune_stable_diffusion.py
```diff
@@ -0,0 +1,463 @@
...
+script provided by Hugging Face
+[here](https://github.com/huggingface/diffusers/blob/main/examples/text_to_image/train_text_to_image.py).
+
+We assume that you have a high-level understanding of the Stable Diffusion model.
+Following resources can be helpful if you're looking for resources in that regard:
+
+* [High-performance image generation using Stable Diffusion in KerasCV](https://keras.io/guides/keras_cv/generate_images_with_stable_diffusion/)
+* [Stable Diffusion with Diffusers](https://huggingface.co/blog/stable_diffusion)
+
+It's highly recommended that you use a GPU having at least 30 GBs of memory to execute
```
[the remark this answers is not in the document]
sayakpaul: Yes! Another bit that needs to be worked out is loss-scaling (as per the [official guide](https://www.tensorflow.org/guide/mixed_precision#training_the_model_with_a_custom_training_loop)). We followed the [same philosophy](https://github.com/keras-team/keras-io/pull/1176#discussion_r1059650956) here and didn't include the mixed-precision related code in the tutorial.

examples/generative/finetune_stable_diffusion.py
```diff
@@ -0,0 +1,463 @@
...
+            latents = self.sample_from_encoder_outputs(self.vae(images, training=False))
+            # Know more about the magic number here:
+            # https://keras.io/examples/generative/fine_tune_via_textual_inversion/
+            latents = latents * 0.18215
+
+            # Sample noise that we'll add to the latents.
+            noise = tf.random.normal(tf.shape(latents))
+
+            # Sample a random timestep for each image.
+            timesteps = tnp.random.randint(
```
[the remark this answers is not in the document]
sayakpaul: `drop_remainder=True` in the `batch()` call should work, but don't you think the data also needs to come from a GCS bucket? 

Apart from that, another XLA-incompatible bit is a backprop operation for an upsampling operation. I noticed it when I set `jit_compile=True` in `trainer.compile()` and then fit it.

examples/generative/finetune_stable_diffusion.py
sayakpaul: My bad. Very sorry about it. You have suggested this multiple times.

examples/generative/finetune_stable_diffusion.py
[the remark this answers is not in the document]
fchollet: > backprop operation for an upsampling operation

Yes, this keeps coming up, I've escalated it with the XLA team a while back. This is unfortunate.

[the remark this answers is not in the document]
sayakpaul: Much appreciated. Thank you!

fchollet:
> Could someone on the Keras team spin up an 80GB A100 to help us do this?

It appears that our team doesn't have access to a 80GB A100 at this time. Is there any way we could make the code TPU compatible or otherwise reduce its memory consumption via mixed precision?

examples/generative/finetune_stable_diffusion.py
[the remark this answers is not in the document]
fchollet: If the example can only realistically be run with mixed precision I'd recommend enabling it by default (same for checkpointing). It doesn't detract from the core of the example but it really helps readers make use of it.

sayakpaul:
> > Could someone on the Keras team spin up an 80GB A100 to help us do this?
> 
> It appears that our team doesn't have access to a 80GB A100 at this time. Is there any way we could make the code TPU compatible or otherwise reduce its memory consumption via mixed precision?

@fchollet while doing the training ourselves, we are using mixed-precision. Still, that takes time on a 40 GB A100. 

Regarding TPU compatibility, the TFRecord part needs to be worked out which is not a bi problem but [this one](https://github.com/keras-team/keras-io/pull/1176#discussion_r1059651771) is.

examples/generative/finetune_stable_diffusion.py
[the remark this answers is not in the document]
sayakpaul: @fchollet FYI, we will add the support for mixed-precision and checkpointing. Will let you know when that's ready.

fchollet:
Right now the best we have on our side (besides TPUs) is the 40GB A100, but even then we have very limited quota. TPUs are likely our best bet, or alternatively we could try to shard the model across multiple smaller GPUs.

For the upsampling issue, what is the line that causes the issue? You could actually use a very simple workaround which is to set `method="nearest"` in `tf.image.resize`, which is in fact implemented in XLA in reverse mode. The default of bilinear isn't implemented yet.

sayakpaul:
> For the upsampling issue, what is the line that causes the issue? You could actually use a very simple workaround which is to set method="nearest" in tf.image.resize, which is in fact implemented in XLA in reverse mode. The default of bilinear isn't implemented yet.

It actually comes from an upsampling layer inside the diffusion model. I fear we cannot do changes to the model level since it might impact the training. 

> Right now the best we have on our side (besides TPUs) is the 40GB A100, but even then we have very limited quota. TPUs are likely our best bet, or alternatively we could try to shard the model across multiple smaller GPUs.

Thank you for trying! We actually fine-tuned for 60 epochs (which is not too bad given the model is known give good results after being fine-tuned for at least 70 epochs). You can check some results with the latest model here: https://huggingface.co/spaces/sayakpaul/pokemon-sd-kerascv. We'll soon update this PR with the recent weights. How does that sound?

tanzhenyu:
> Right now the best we have on our side (besides TPUs) is the 40GB A100, but even then we have very limited quota. TPUs are likely our best bet, or alternatively we could try to shard the model across multiple smaller GPUs.
> 
> For the upsampling issue, what is the line that causes the issue? You could actually use a very simple workaround which is to set `method="nearest"` in `tf.image.resize`, which is in fact implemented in XLA in reverse mode. The default of bilinear isn't implemented yet.

My take is changing to nearest is fine in most cases, not sure if makes any difference in this model

fchollet:
> My take is changing to nearest is fine in most cases, not sure if makes any difference in this model

Could we update the StableDiffusion code with this change? Would this result in worse generation quality?

> We'll soon update this PR with the recent weights. How does that sound?

Sounds great, thank you!

sayakpaul:
@fchollet here is a summary of https://github.com/keras-team/keras-io/pull/1176/commits/66414740aee74deed46a149c50689092b5014289:

* Added new results.
* Added support for mixed-precision and model checkpointing.
* Added relevant commentary and code comments. 

Let me know if there's anything pending.

examples/generative/finetune_stable_diffusion.py
```diff
@@ -0,0 +1,505 @@
...
+![custom-pokemons](https://i.imgur.com/X4m614M.png)
+
+For the code, the tutorial relies on KerasCV 0.3.5 which is not yet available on PyPI.
+So, we need to install it from the source. Additionally, we need the latest stable 
+version of TensorFlow.
+"""
+
+"""shell
+pip install git+https://github.com/keras-team/keras-cv -q
+pip install -U tensorflow -q
```
sayakpaul, line 38: Need so that mixed-precision works with the experimental AdamW.

examples/generative/finetune_stable_diffusion.py
```diff
@@ -0,0 +1,507 @@
...
+* [Stable Diffusion with Diffusers](https://huggingface.co/blog/stable_diffusion)
+
+It's highly recommended that you use a GPU having at least 30 GBs of memory to execute
+the code.
+
+By the end of the guide, you'll be able to generate images of interesting pokemons:
+
+![custom-pokemons](https://i.imgur.com/X4m614M.png)
+
+For the code, the tutorial relies on KerasCV 0.3.5 which is not yet available on PyPI.
```
tanzhenyu, line 31: I believe 0.3.5 is already released? Or ask users to directly rely on 0.4 is better IMO

examples/generative/finetune_stable_diffusion.py
```diff
@@ -0,0 +1,507 @@
...
+So, we need to install it from the source. Additionally, we need the latest stable 
+version of TensorFlow.
```
tanzhenyu, line 33: or at least TF 2.11 for AdamW

examples/generative/finetune_stable_diffusion.py
```diff
@@ -0,0 +1,507 @@
...
+as an image encoder.
+* A diffusion model that refines an input latent and produces another latent in the
+presence of the encoded text prompt.
+* A decoder that generates images given a latent from the diffusion model.
+
+It's worth noting that during the process of generating an image from a text prompt, the
+image encoder is not typically employed.
+
+However, during the process of fine-tuning, the workflow goes like the following:
+
```
tanzhenyu, line 57: Maybe document the fact that image encoder and text encoders are pretrained before this, so the entire point of this is to only fine-tune the diffusion model.

examples/generative/finetune_stable_diffusion.py
```diff
@@ -0,0 +1,507 @@
...
+to predict the noise that was added to the image latent.
+5. A reconstruction loss is calculated between the predicted noise and the original noise
+added in step 3.
+6. Finally, the diffusion model parameters are then optimized w.r.t this loss using
+gradient descent.
+
+Note that only the diffusion model parameters are updated during fine-tuning while the
+text and the image encoders are kept frozen.
+
+Don't worry if this reads complicated. The code is much simpler than this.
```
tanzhenyu, line 72: not related to this PR, but we should maybe consider providing a train-from-scratch example under KerasCV

examples/generative/finetune_stable_diffusion.py
```diff
@@ -0,0 +1,507 @@
...
+"""
+Since we have only 833 {image,caption} pairs we can precompute the text embeddings from
+the captions. Moreover, the text encoder will be kept frozen during the course of
+fine-tuning. So, we can save some compute by doing this.
+
+Before we employ the text encoder, we need to tokenize the captions.
+"""
+
+PADDING_TOKEN = 49407
+MAX_PROMPT_LENGTH = 77
```
tanzhenyu, line 126: Sometimes those magic values will get lost -- consider documenting them, for example, 77 means maximum number of tokens the tokenizer / text encoder is able to accept.

examples/generative/finetune_stable_diffusion.py
```diff
@@ -0,0 +1,507 @@
...
+AUTO = tf.data.AUTOTUNE
+POS_IDS = tf.convert_to_tensor([list(range(MAX_PROMPT_LENGTH))], dtype=tf.int32)
+
+augmenter = keras_cv.layers.Augmenter(
+    layers=[
+        keras_cv.layers.CenterCrop(RESOLUTION, RESOLUTION),
+        keras_cv.layers.RandomFlip(),
+        tf.keras.layers.Rescaling(scale=1.0 / 127.5, offset=-1),
+    ]
+)
```
tanzhenyu, line 166: Nice!

examples/generative/finetune_stable_diffusion.py
```diff
@@ -0,0 +1,507 @@
...
+text_encoder = TextEncoder(MAX_PROMPT_LENGTH)
+
+
+def process_image(image_path, tokenized_text):
+    image = tf.io.read_file(image_path)
+    image = tf.io.decode_png(image, 3)
+    image = tf.image.resize(image, (RESOLUTION, RESOLUTION))
```
tanzhenyu, line 173: nit: kerascv might already have a resizing layer which you could add to the augmenter

examples/generative/finetune_stable_diffusion.py
```diff
@@ -0,0 +1,507 @@
...
+    return image, tokenized_text
+
+
+def apply_augmentation(image_batch, token_batch):
+    return augmenter(image_batch), token_batch
```
tanzhenyu, line 178: the good thing about KPL and tf.data is that you can simply do:
transformed_dataset = dataset.map(augmenter)

examples/generative/finetune_stable_diffusion.py
```diff
@@ -0,0 +1,507 @@
...
+    dataset = dataset.map(prepare_dict, num_parallel_calls=AUTO)
+    return dataset.prefetch(AUTO)
+
+
+"""
+The baseline Stable Diffusion model was trained using images of 512x512 resolution. It's
+unlikely for a model that's trained using higher resolution images to transfer well to
+lower resolution images. However, the current model will lead to OOM if we keep the
+resolution to 512x512 (without enabling mixed-precision). Therefore, in the interest of
+interactive demonstration, we kept the input resolution to 256x256.
```
tanzhenyu, line 212: might be able to workaround this with a batch_size=2 too?

examples/generative/finetune_stable_diffusion.py
```diff
@@ -0,0 +1,507 @@
...
+    plt.title(text)
+
+    plt.axis("off")
+
+"""
+## A trainer class for the fine-tuning loop
+"""
+
+
+class Trainer(tf.keras.Model):
```
tanzhenyu, line 248: same as above comment, we should consider providing a trainer in kerascv.
And thanks for the pointer here!

examples/generative/finetune_stable_diffusion.py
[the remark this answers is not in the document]
tanzhenyu: why not just define a macro `BATCH_SIZE=4`,
and dataset.batch(BATCH_SIZE)`?

examples/generative/finetune_stable_diffusion.py
```diff
@@ -0,0 +1,507 @@
...
+            latents = self.sample_from_encoder_outputs(self.vae(images, training=False))
+            # Know more about the magic number here:
+            # https://keras.io/examples/generative/fine_tune_via_textual_inversion/
+            latents = latents * 0.18215
+
+            # Sample noise that we'll add to the latents.
+            noise = tf.random.normal(tf.shape(latents))
+
+            # Sample a random timestep for each image.
+            timesteps = tnp.random.randint(
```
tanzhenyu, line 287: nit: timestep -> sampled_t?

examples/generative/finetune_stable_diffusion.py
```diff
@@ -0,0 +1,507 @@
...
+
+            # Get the target for loss depending on the prediction type
+            # just the sampled noise for now.
+            target = noise  # noise_schedule.predict_epsilon == True
+
+            # Predict the noise residual and compute loss.
+            timestep_embedding = tf.map_fn(
+                lambda t: self.get_timestep_embedding(t), timesteps, dtype=tf.float32
+            )
+            timestep_embedding = tf.squeeze(timestep_embedding, 1)
```
tanzhenyu, line 305: sometimes I wish this part can be automatically handled by the diffusion_model

examples/generative/finetune_stable_diffusion.py
```diff
@@ -0,0 +1,507 @@
...
+        return embedding
+
+    def sample_from_encoder_outputs(self, outputs):
+        mean, logvar = tf.split(outputs, 2, axis=-1)
+        logvar = tf.clip_by_value(logvar, -30.0, 20.0)
+        std = tf.exp(0.5 * logvar)
+        sample = tf.random.normal(tf.shape(mean), dtype=mean.dtype)
+        return mean + std * sample
+
+    def save_weights(self, filepath, overwrite=True, save_format=None, options=None):
```
tanzhenyu, line 341: ultimately do you still want to save the encoder part, so it could load and run inference later?

examples/generative/finetune_stable_diffusion.py
```diff
@@ -0,0 +1,507 @@
...
+    save_weights_only=True,
+    monitor="loss",
+    mode="min",
+)
+diffusion_ft_trainer.fit(training_dataset, epochs=epochs, callbacks=[ckpt_callback])
+
+"""
+## Inference
+
+We fine-tuned the model for 60 epochs on an image resolution of 512x512. To allow
```
tanzhenyu, line 422: IIUC -- the above 1 epoch training is using 256x256, but users at this point will load pre-fine-tuned 512x512 weights?

tanzhenyu (changes requested):
Sorry for the late review.
This PR looks great! Thanks for the contribution!

examples/generative/finetune_stable_diffusion.py
sayakpaul: I couldn't find a good way to do that. But included a phrase later to emphasize the point.

examples/generative/finetune_stable_diffusion.py
sayakpaul: I don't think it's necessary. And I think the variable names make it easier for readers to understand their usage.

examples/generative/finetune_stable_diffusion.py
sayakpaul: But we cannot batch images together if they are of different sizes. So, this initial resizing step is necessary.

examples/generative/finetune_stable_diffusion.py
sayakpaul: I am aware of it. I didn't do it to keep the data pipeline code easier.

examples/generative/finetune_stable_diffusion.py
sayakpaul: Did not get your point.

How does increasing batch size prevent an OOM?

examples/generative/finetune_stable_diffusion.py
sayakpaul: > why not just define a macro BATCH_SIZE=4

That impacts the dynamic behavior of the code IMO. In the current case, users can just pass on a single batch size and let the code take care of the rest. Also, increasing batch size, in this case, might easily lead to OOM, so that option is a bit restricted. 

> nit: timestep -> sampled_t?

Not consistent with the rest of the guides we have and the existing codebases on Stable Diffusion.

examples/generative/finetune_stable_diffusion.py
sayakpaul: See the inference code later to see how that might not be required. 

We just do:

```py
pokemon_model = keras_cv.models.StableDiffusion(
    img_width=img_width, img_height=img_height
)
# We just reload the weights of the fine-tuned diffusion model.
pokemon_model.diffusion_model.load_weights(weights_path)
```

examples/generative/finetune_stable_diffusion.py
sayakpaul: Yes.

sayakpaul:
@tanzhenyu addressed your comments and made some changes as per your feedback.

examples/generative/finetune_stable_diffusion.py
tanzhenyu: Makes sense

examples/generative/finetune_stable_diffusion.py
tanzhenyu: What do you mean? That layer is specifically used for resizing to the same shape, it's used before dataset.batch

examples/generative/finetune_stable_diffusion.py
tanzhenyu: Because you're using batch_size=4 here, halving it makes it less likely to get OOM?

examples/generative/finetune_stable_diffusion.py
```diff
@@ -0,0 +1,507 @@
...
+* [Stable Diffusion with Diffusers](https://huggingface.co/blog/stable_diffusion)
+
+It's highly recommended that you use a GPU having at least 30 GBs of memory to execute
+the code.
+
+By the end of the guide, you'll be able to generate images of interesting pokemons:
+
+![custom-pokemons](https://i.imgur.com/X4m614M.png)
+
+For the code, the tutorial relies on KerasCV 0.4.0 which is not yet available on PyPI.
```
tanzhenyu, line 31: It is available on PyPI

examples/generative/finetune_stable_diffusion.py
sayakpaul: `augmenter` is used after the batches have been formed. To form the initial batches, the images (to be present in the batches) need to be of the same shape. I know that the KerasCV resizing layer can be used for that purpose. But you mentioned adding the resizing layer to the `augmenter`. 

What am I missing?

examples/generative/finetune_stable_diffusion.py
tanzhenyu: Do you want to comment on the padding_token? It's used in KerasCV and here, but those numbers might change with different versions of text encoder / SD models.

examples/generative/finetune_stable_diffusion.py
sayakpaul: Oh the current code will OOM out anyway with 512x512 resolution (if not using an A100). Since we're already using 256x256, keeping the batch size to 4 won't hurt IMO.

examples/generative/finetune_stable_diffusion.py
sayakpaul: You're right. Just pushed the changes.

fchollet (commented):
Thanks for the updates! I pushed some copyedits. Please note that I changed the title, but the final title is up to you. I believe that something along the lines of "fine-tuning stable diffusion" will be easier for readers to immediately grasp what the tutorial is about.

Unless there are further comments, I think we can add the generated files?

sayakpaul:
@fchollet thank you! Will work on adding the generated files soon and let you know once done.

sayakpaul:
@fchollet,

Added the generated files.

tanzhenyu (approved):

fchollet:
Awesome, thank you for the great contribution! 👍

sayakpaul:
@fchollet is there a problem with the CD? I don't see the example live yet.

fchollet:
@sayakpaul It should be coming up later today.

PR #1178, opened 2022-12-28
Some minor fixes for keras-nlp docs

mattdangerw:
We will use the same install line everywhere which will update tensorflow to the latest version (useful for colab with old tf).

Also switch the optimizer from AdamW -> Adam in our quickstart, as AdamW does not converge on tf 2.9, and the quickstart may be copied and pasted into a lot of environments we don't control.


jbischof (approved):
Thank you!

fchollet (approved):
Thanks!

PR #1179, opened 2022-12-29
Fix broken link for KerasNLP Getting Started guide

jbischof:
Need absolute reference to link to guide from the API page.

TESTED: built site locally and checked link.


fchollet (approved):
LGTM, thanks!

PR #1182, opened 2022-12-30
Clean up class name and `self` in calls to `super()`

mbrukman:
PEP 3135 [^1] simplified the syntax for `super()` in Python 3.0 from:

    super(ClassName, self)

to the following very simple and equivalent [^2] syntax:

    super()

The current Keras codebase already requires Python 3+, and there's a mix of the two syntax formats above, sometimes both are used in a single file.

This change simplifies the entire code base by cleaning up the remaining explicit uses of the current class name and `self` and using the cleaner `super()` syntax everywhere consistently.

Since the new syntax is intended to be a shorthand for the old syntax, this change should have no semantic differences from before.

[^1]: https://peps.python.org/pep-3135/
[^2]: https://docs.python.org/3/library/functions.html#super


fchollet (changes requested):
Thanks for the PR! Please replicate the fixes in the corresponding `.md` and `.ipynb` files.

mbrukman:
> Thanks for the PR! Please replicate the fixes in the corresponding `.md` and `.ipynb` files.

Done.

fchollet (approved):
LGTM, thanks!

PR #1185, opened 2023-01-06
Use resnet50 'preprocess_input' function before feeding them to model

balvisio:
The keras source code shows that `keras.applications.resnet50` shows that the input mode is `caffe` instead of `tf`. (https://github.com/keras-team/keras/blob/e6784e4302c7b8cd116b74a784f4b78d60e83c26/keras/applications/resnet.py#L606-L613). Thus, normalizing the input to `[-1, 1]` might no be appropriate. It is suggested to use `tf.keras.applications.resnet50.preprocess_input()` instead. This will actually subtract the same mean from each channel used when training the model.


fchollet (approved):
LGTM, thanks. This will also better generalize when people attempt to use a different backbone.

PR #1206, opened 2023-01-21
Update API symbols + genapi specs

LukeWood:
Lets still hold off on the v0.4.0 publicity push until:

- keras.io/guides/retinanet is fixed
- keras.io/keras_cv is updated

I will update keras.io/keras_cv in a separate PR


fchollet (approved):
LGTM, thank you!

PR #1231, opened 2023-02-12
Update dreambooth.md

LukeWood:
Remove unneeded warning from dreambooth.md.

I recently fixed this in keras-cv - but we haven't performed a release yet.


fchollet (approved):
LGTM, thanks

PR #1237, opened 2023-02-14
Changes in Focal Modulation Network: self-attention eqn, norm for modulator, acknowledgement

ariG23498:
The self attention equation (in Equation 2) was incorrect (pointed out by [Jianwei](https://github.com/jwyang)) as we were using `Q` throughout the eqation. The current equation uses `Q(x)` , `K(x)`, and `V(x)`.

We change the `reduce_mean` to `l2 norm` for better viz of the modulator. This change was also pointed out by [Jianwei](https://github.com/jwyang).

The change in the acknowledgement reflects the contribution of [Jianwei's](https://github.com/jwyang) review.

CC: @fchollet @ritwikraha


fchollet (approved):
Thanks! Please update the generated files.

PR #1240, opened 2023-02-19
Added responsiveness to landing page images.

Neeshamraghav012:
Some images of the landing page are not responsive on some devices. So I played with the CSS of the landing page and found out that `min-height: 20rem` works better than `min-height: 25rem`. That's why I changed it to 20rem.


fchollet (approved):
SG, thanks

PR #1245, opened 2023-02-22
Update the keras-nlp getting started guide with compilation defaults

mattdangerw:


fchollet (approved):
LGTM, thanks! Please update the generate files.

jbischof:
Thanks @mattdangerw will look later today!

mattdangerw:
@fchollet I think @jbischof wanted to update some copy elsewhere in the guide. After that I will update the generated files and we can land.

PR #1268, opened 2023-03-08
Fix the index of the position in expert

hieuza:
The current position starts at 1 due to the cumulative sum, it should be start from 0.


ksalama:
LGTM

PR #1269, opened 2023-03-08
Fix the dispatch_tensor -- must be binary value

hieuza:


fchollet (commented):
Thanks for the PR. It looks good. Please replicate the fix in the ipynb and md files as well.

PR #1291, opened 2023-03-21
Add Token Classification Example for KerasNLP

shivance:
Closes [keras-nlp/#754](https://github.com/keras-team/keras-nlp/issues/754)

Using BERT to reduce the script running time, will change to RoBERTa later

### DIFF
--- a/examples/nlp/semantic_similarity_with_keras_nlp.py
+++ b/examples/nlp/semantic_similarity_with_keras_nlp.py
@@ -17,12 +17,13 @@ semantic similarity with HuggingFace Transformers library. In this tutorial we w
 learn how to use [KerasNLP](https://keras.io/keras_nlp/), an extension of the core Keras API, 
 for the same task. We'll also learn how KerasNLP reduces the boilerplate code and makes models easy to use.
 
-This guide is broken into three parts:
+This guide is broken into following parts:
 
 1. *Setup*, task definition, and establishing a baseline.
-2. *Training* a BERT model.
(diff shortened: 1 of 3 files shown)


examples/nlp/semantic_similarity_with_keras_nlp.py
```diff
@@ -0,0 +1,263 @@
+"""
```
mattdangerw, line 1: I think you pulled in your other guide by accident, you may want to rebase out the commits for this guide.

examples/nlp/token_classification.py
```diff
@@ -0,0 +1,127 @@
+from datasets import load_dataset
```
mattdangerw, line 1: I think we can go ahead and start filling out the rough text for this guide anytime. So introduce the text classification task, talk about the dataset, why we need to preprocess it for the BERT subword tokenizer in particular.

Even if we end up tweaking the preprocessing routines a bit during review, the overall flow and intro description can probably stay fairly static!

mattdangerw:
Thank you! @sampathweb and I will take a look, but left some high level comments for now.

shivance:
I've opened an issue [here](https://github.com/keras-team/keras-nlp/issues/927) which elaborates the potential and clear cut roadmap for this example, and needs review from @mattdangerw and @sampathweb

examples/nlp/token_classification.py
shivance: Sounds good @mattdangerw .

PR #1306, opened 2023-03-28
fix typos in KerasCV guide using RetinaNet

jhhan0:
- found some misspellings and punctuation errors in KerasCV guide, so I fixed them.

Comment down below if there are any other parts to change. I will update .ipynb and .md files accordingly if all things are set!


fchollet (commented):
Thanks for the PR! Please replicate the fix in the md and ipynb files.

LukeWood:
I'll incorporate these into the new guide!  We're pretty close to opening a PR for it!

Issue #1235, opened 2023-02-13
Add model saving and reloading sections to code example


innat:
It would be good practice to demonstrate model saving and re-loading options while writing code example. Practitioner use keras code example and in some cases, they face issue to successfully save and reload the model. Here are some issue asked in stack.

- [SO]: https://keras.io/examples/vision/shiftvit/ - https://stackoverflow.com/a/72484085/9215780
- [SO]: https://keras.io/examples/vision/shiftvit/ - https://stackoverflow.com/a/72496860/9215780
- [SO]: https://keras.io/examples/vision/nnclr/ - https://stackoverflow.com/a/75410780/9215780

nkovela1:
Thanks for the request! Would you be interested in contributing sections to these tutorials regarding saving and loading for these models?

innat:
Sure. I will.

Additionally, I think this should be documented in keras io contribution guidelines. WDYT?

nkovela1:
Good idea! We will add to the contribution guide under "Model development best practices." We are in the process of rolling out a new default saving format for Keras and will update the guide once this is in a new stable release.

SuryanarayanaY:
Hi @innat ,

The proposed PR has been merged now and the changes will affect soon. Shall we mark it as resolved ?

Thanks!

innat:
Not really. Or, I am not sure.
In the keras-core repo, contributors converts the keras code example to backend agnostoc api. Those update should now run on tf, torch and jax. I am not sure if they were suggested to add section for saving and reloading with keras_v3 format.

PR #1314, opened 2023-04-03
Modernization of the KerasCV Object Detection guide

LukeWood:
This is a huge overhaul to the KerasCV OD API!

Before publishing I have a few action items:

- [x] fix numerical differences between PyCOCOCallback and BoxCOCOMetrics (should be quick)
- [x] update the visualization utility to make text not overflow image
- [x] allow customization of font size in visualization utility
- [x] show confidence scores in the visualization utility

----

Demo colab: https://colab.sandbox.google.com/github/LukeWood/keras-io/blob/odguidev2/guides/ipynb/keras_cv/object_detection_keras_cv.ipynb


guides/keras_cv/object_detection_keras_cv.py
```diff
@@ -0,0 +1,750 @@
...
+)
+eval_ds = load_pascal_voc(split="test", dataset="voc/2007", bounding_box_format="xywh")
+eval_ds = load_pascal_voc(dataset="voc/2007", split="test", bounding_box_format="xywh")
+
+train_ds = train_ds.shuffle(BATCH_SIZE * 4)
+visualization.visualize_dataset(train_ds, value_range=(0, 255), rows=2, cols=2)
+"""
+Next, lets batch our data.
+
+In KerasCV object detection tasks it is recommended that
```
jbischof, line 320: Why do the batches themselves need to be ragged? That doesn't seem to be related to image size of having a different number of labels in a dict.

guides/keras_cv/object_detection_keras_cv.py
```diff
@@ -0,0 +1,750 @@
...
+visualization.visualize_dataset(eval_ds, bounding_box_format="xywh", path="eval.png")
+
+"""
+Looks like everything is structured as expected.
+Now we can move on to constructing our
+data augmentation pipeline.
+
+## Data augmentation
+
+One of the most labor-intensive tasks when constructing object detection
```
jbischof, line 362: s/labor-intensive/challenging

We're all having fun, right?

guides/keras_cv/object_detection_keras_cv.py
```diff
@@ -0,0 +1,750 @@
...
+dataset.
+
+KerasCV makes it easy to construct a `RetinaNet` with any of the KerasCV
+backbones.  Simply use one of the presets for the architecture you'd like!
+
+For example:
+"""
+
+model = keras_cv.models.RetinaNet.from_preset(
+    "resnet50_v2_imagenet",
```
jbischof, line 452: s/resnet50_v2_imagenet/resnet50_imagenet

guides/keras_cv/object_detection_keras_cv.py
```diff
@@ -0,0 +1,750 @@
...
+
+This matches what we have constructed in our input pipeline above.
+"""
+
+"""
+### Optimizer
+
+Next, let's construct a learning rate schedule and optimizer for our model.
+"""
+base_lr = 0.005 * (BATCH_SIZE / 16)
```
jbischof, line 473: Might be good to offer some more explanation and visualize the LR schedule. This is an advanced topic.

guides/keras_cv/object_detection_keras_cv.py
```diff
@@ -0,0 +1,750 @@
...
+lr_decay = tf.keras.optimizers.schedules.PiecewiseConstantDecay(
+    boundaries=[12000 * 16, 16000 * 16],
```
jbischof, line 475: The `boundaries` arg seems arbitrary. I would offer intuition and some rules of thumb. Why 16? Why 12K?

guides/keras_cv/object_detection_keras_cv.py
```diff
@@ -0,0 +1,750 @@
...
+optimizer = tf.keras.optimizers.SGD(
+    learning_rate=lr_decay, momentum=0.9, global_clipnorm=10.0
+)
+
+"""
+**Note: be sure to freeze BatchNormalization layers***
+
+And important but easy to miss step in training your own object detection model
+is freezing the `BatchNormalization` layers in your backbone.
+We felt it was a violation of user expectations to automatically do this,
```
jbischof, line 489: We felt it would be confusing to do this automatically

guides/keras_cv/object_detection_keras_cv.py
```diff
@@ -0,0 +1,750 @@
...
+Due to the high computational cost of computing COCO metrics, the KerasCV
+`BoxCOCOMetrics` component requires an `evaluate_freq` parameter to be passed to
+its constructor.  Every `evaluate_freq`-th call to `update_state()`, the metric
+will recompute the result.  In between invocations, a cached version of the
+result will be returned.
+
+To force an evaluation, you may call `coco_metrics.result(force=True)`:
+"""
+model.compile(
+    classification_loss="focal",
```
jbischof, line 519: These loss functions deserve some (even brief) explanation since they are custom to this problem.

guides/keras_cv/object_detection_keras_cv.py
```diff
@@ -0,0 +1,750 @@
...
+
+## Inference and Plotting results
+
+KerasCV makes object detection inference simple.  `model.predict(images)` returns a
+RaggedTensor of bounding boxes.  By default, `RetinaNet.predict()` will perform
+a non max suppression operation for you.
+
+In this section, we will use a `keras_cv` provided preset:
+"""
+model = keras_cv.models.RetinaNet.from_preset(
```
jbischof, line 621: Didn't we already define this on line 58?

guides/keras_cv/object_detection_keras_cv.py
```diff
@@ -0,0 +1,750 @@
...
+    optimizer=optimizer,
+    # We will use our custom callback to evaluate COCO metrics
+    metrics=None,
+)
+
+"""
+And run `model.fit()`!
+"""
+
+# If you want to train the fully model, uncomment `.take(20)` from each
```
jbischof, line 590: I would suggest showing how to instantiate a model from a pretrained backbone and then running a small number of steps without worrying about visualization and metric values. This is what most users will want to do and we want a "plug and play" example for them to copy.

guides/keras_cv/object_detection_keras_cv.py
```diff
@@ -0,0 +1,750 @@
...
+model training, visualization, and
+metric evaluation easy.
+
+Some follow up exercises for the reader:
+
+- add additional augmentation techniques to improve model performance
+- tune the hyperparameters and data augmentation used to produce high quality results
+- train an object detection model on your own dataset
+
+One last fun code snippet to showcase the power of KerasCV's modular API!
```
jbischof, line 736: I'm not sure this showcases modularity. We're passing the outputs of one monolithic API to another. It does showcase the diversity of our offerings however.

jbischof (commented):
Amazing start @LukeWood! Starting with some high level suggestions.

I think the next steps are to rearrange this guide as follows:
1. High level explanation of OD problem
2. Understanding the input data (both tensors and image visualizaton)
3. Understanding the metrics
4. Load a pretrained model, evaluate the metrics, and visualize some of the predictions
5. Show how to instantiate a model with pretrained backbone only
6. Explain how to set up the learning rate and loss functions
7. Run a couple of training steps for demonstration purposes and give some advice on how long the model would actually need to train. Do not bother with the metrics or predictions (they will be bad).

Finally, let's keep the tone light and helpful. This is supposed to be fun!

guides/keras_cv/object_detection_keras_cv.py
LukeWood: My honest answer is that anecdotally I found these to do well.

guides/keras_cv/object_detection_keras_cv.py
```diff
@@ -0,0 +1,750 @@
...
+For example:
+
+```python
+train_ds, ds_info = your_data_loader.load(
+    split='train', bounding_box_format='xywh', batch_size=8
+)
+```
+
+Clearly yields bounding boxes in the format `xywh`.  You can read more about
+KerasCV bounding box formats [in the API docs](https://keras.io/api/keras_cv/bounding_box/formats/).
```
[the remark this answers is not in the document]
LukeWood: Good catch, I've moved this higher up.

guides/keras_cv/object_detection_keras_cv.py
LukeWood: Last time we shipped the guide it felt pretty incomplete without a dedicated inference section.  

Personally, I say we keep it!

guides/keras_cv/object_detection_keras_cv.py
LukeWood: Yes, but by doing this we re-initialize the model to make sure we have a fully pretrained model.  Above we override it, and then we don't get strong convergence necessarily.

guides/keras_cv/object_detection_keras_cv.py
LukeWood: removed -modularity

LukeWood:
Ok @jbischof @fchollet @ianstenbit ready for another round of reviews!

guides/keras_cv/object_detection_keras_cv.py
```diff
@@ -0,0 +1,750 @@
...
+To get started, let's sort out all of our imports and define global
+configuration parameters.
+"""
+
+BATCH_SIZE = 4
+low, high = resource.getrlimit(resource.RLIMIT_NOFILE)
+resource.setrlimit(resource.RLIMIT_NOFILE, (high, high))
+
+"""
+## Data loading
```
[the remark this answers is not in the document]
LukeWood: Reorganized a bit, let me know what you think now

guides/keras_cv/object_detection_keras_cv.py
```diff
@@ -0,0 +1,868 @@
...
+images, and your labels are bounding boxes with optional class
+labels.
+
+**For example:**
+
+<img width="300" src="https://i.imgur.com/QTvxIZH.jpeg">
+
+The data often looks something like this:
+```
+image = [height, width, 3])
```
fchollet, line 48: Remove extra parens

guides/keras_cv/object_detection_keras_cv.py
```diff
@@ -0,0 +1,868 @@
...
+While this may seem simple, it is a critical piece of the KerasCV object
+detection API!
+Every component that is to process bounding boxes requires
+`bounding_box_format` argument.
+You can read more about
+KerasCV bounding box formats [in the API docs](https://keras.io/api/keras_cv/bounding_box/formats/).
+
+
+This is done because there is no one correct format for bounding boxes!
+Components in different pipelines expect different formats, and so by requiring them to be specified we ensure that our components remain readable, reusable, and clear.
```
fchollet, line 128: Break long line

guides/keras_cv/object_detection_keras_cv.py
```diff
@@ -0,0 +1,868 @@
...
+eval_ds = eval_ds.map(dict_to_tuple, num_parallel_calls=tf.data.AUTOTUNE)
+
+train_ds = train_ds.prefetch(tf.data.AUTOTUNE)
+eval_ds = eval_ds.prefetch(tf.data.AUTOTUNE)
+
+"""
+Our data pipeline is now complete!
+We can now move on to model creation and training.
+
+### Model creation
```
fchollet, line 508: Should be a lvl-2 section (or create an encompassing lvl-2 section for "model architecture", "optimizer", "loss functions", "metrics")

guides/keras_cv/object_detection_keras_cv.py
```diff
@@ -0,0 +1,868 @@
...
+class EvaluateCOCOMetricsCallback(keras.callbacks.Callback):
+    def __init__(self, data):
+        super().__init__()
+        self.data = data
+        self.metrics = keras_cv.metrics.BoxCOCOMetrics(
+            bounding_box_format="xywh",
+            # passing 1e9 ensures we never evaluate until
+            # `metrics.result(force=True)` is
+            # called.
+            evaluate_freq=1e9,
```
fchollet, line 661: Is this really necessary? Why not just remove the call to `result()` in `on_epoch_end`?

guides/keras_cv/object_detection_keras_cv.py
```diff
@@ -0,0 +1,868 @@
...
+model training, visualization, and
+metric evaluation easy.
+
+Some follow up exercises for the reader:
+
+- add additional augmentation techniques to improve model performance
+- tune the hyperparameters and data augmentation used to produce high quality results
+- train an object detection model on your own dataset
+
+One last fun code snippet to showcase the power of KerasCV's API!
```
fchollet, line 841: Nice touch!

guides/keras_cv/object_detection_keras_cv.py
```diff
@@ -0,0 +1,868 @@
...
+    scale=5,
+    font_scale=0.7,
+    bounding_box_format="xywh",
+    class_mapping=class_mapping,
+)
+
+"""
+Object detection, using a pretrained ML model, on ML generated images, all using
+KerasCV models!
+Awesome!
```
fchollet, line 867: You can drop that last one

fchollet (commented):
Thanks for the PR!

Since the guide URL is being changed, please add an entry in `redirects/` to redirect the old location to the old one.

guides/keras_cv/object_detection_keras_cv.py
```diff
@@ -0,0 +1,868 @@
...
+Finally, let's unpackage our inputs from the preprocessing dictionary, and
+prepare to feed the inputs into our model.  In order to be TPU compatible,
+bounding box Tensors need to be `Dense` instead of `Ragged`.  If training on
+GPU, you can omit the `bounding_box.to_dense()` call.  If ommitted,
+the KerasCV RetinaNet
+label encoder will automatically correctly encode Ragged training targets.
+"""
+
+def dict_to_tuple(inputs):
+    return inputs["images"], bounding_box.to_dense(
```
jbischof, line 494: Is this a static method on a class `bounding_box`? Is `bounding_box` just a namespace? Hard to tell

guides/keras_cv/object_detection_keras_cv.py
```diff
@@ -0,0 +1,868 @@
...
+eval_ds = eval_ds.prefetch(tf.data.AUTOTUNE)
+
+"""
+Our data pipeline is now complete!
+We can now move on to model creation and training.
+
+### Model creation
+
+We'll use the KerasCV API to construct a RetinaNet model.
+In this tutorial we use a pretrained ResNet50 backbone, initializing the
```
jbischof, line 511: you can just say "using pretrained weights produced"

guides/keras_cv/object_detection_keras_cv.py
```diff
@@ -0,0 +1,868 @@
...
+
+KerasCV makes it easy to construct a `RetinaNet` with any of the KerasCV
+backbones.  Simply use one of the presets for the architecture you'd like!
+
+For example:
+"""
+
+model = keras_cv.models.RetinaNet.from_preset(
+    "resnet50_imagenet",
+    # number of classes to be used in box classification
```
jbischof, line 523: This is probably a good opportunity to explain the `+ 1`

guides/keras_cv/object_detection_keras_cv.py
```diff
@@ -0,0 +1,868 @@
...
+
+**Note: be sure to freeze BatchNormalization layers when training a RetinaNet.***
+
+And important but easy to miss step in training your own object detection model
+is freezing the `BatchNormalization` layers in your backbone.
+We felt it would be confusing to do this automatically,
+so instead we recommend you do it in your training loops.
+"""
+
+for layer in model.backbone.layers:
```
jbischof, line 545: Should this be a method on `RetinaNet`? Seems very handy

```python
model.freeze_batchnorm_layers()
model.unfreeze_batchfrom_layers()
```

guides/keras_cv/object_detection_keras_cv.py
```diff
@@ -0,0 +1,868 @@
...
+    if isinstance(layer, keras.layers.BatchNormalization):
+        layer.trainable = False
+
+"""
+
+### Optimizer
+
+In this guide, we use a standard SGD optimizer and rely on the [`keras.callbacks.ReduceLROnPlateau`](https://keras.io/api/callbacks/reduce_lr_on_plateau/) callback to reduce the learning rate.
```
jbischof, line 553: These are really long lines.

guides/keras_cv/object_detection_keras_cv.py
```diff
@@ -0,0 +1,868 @@
...
+You may not be familiar with the `"focal"` or `"smoothl1"` losses.  While not common in other models, these losses are more or less staples in the object detection world.
+
+In short, ["Focal Loss"](https://arxiv.org/abs/1708.02002) places extra emphasis on difficult training examples.  This is useful when training the classification loss, as the majority of the losses are assigned to the background class.
+
+"SmoothL1 Loss" is used to [prevent exploding gradients](https://arxiv.org/abs/1504.08083) that often occur when attempting to perform the box regression task.
+
+In KerasCV you can use these losses simply by passing the strings `"focal"` and `"smoothl1"` to `compile()`:
+"""
+
+pretrained_model.compile(
```
jbischof, line 583: `model.compile(...)`?

guides/keras_cv/object_detection_keras_cv.py
jbischof: I think the untrained model creation should occur after you discuss loss and metrics on the pretrained model

guides/keras_cv/object_detection_keras_cv.py
```diff
@@ -0,0 +1,868 @@
...
+model.compile(
+    classification_loss="focal",
+    box_loss="smoothl1",
+    optimizer=optimizer,
+    # We will use our custom callback to evaluate COCO metrics
+    metrics=None,
+)
+
+"""
+If you want to train the fully model, uncomment `.take(20)` from each
```
jbischof, line 693: Thank you, this is great!

guides/keras_cv/object_detection_keras_cv.py
```diff
@@ -0,0 +1,868 @@
...
+            images, y_true = batch[0], batch[1]
+            y_pred = self.model.predict(images, verbose=0)
+            self.metrics.update_state(y_true, y_pred)
+
+        metrics = self.metrics.result()
+        logs.update(metrics)
+        return logs
+
+"""
+## Training our model
```
jbischof, line 676: I think this is the penultimate section of the guide. It's weird to go back and forth between the pretrained and random models.

jbischof (commented):
This is SO much better, thank you!

My biggest remaining suggestion is to essentially do everything with the pretrained model FIRST and only at the end show the extra steps needed to train one from scratch. It's fairly confusing to keep going back and forth between the pretrained and random models.

I have trouble reviewing the raw python personally because I can't see any of the images produced by the runtime. Would love to see a notebook version.

guides/keras_cv/object_detection_keras_cv.py
```diff
@@ -0,0 +1,868 @@
...
+```
+bounding_boxes = {
+  "classes": [num_boxes],
+  "boxes": [num_boxes, 4]
+}
+```
+
+This argument describes *exactly* what format the values in the `"boxes"`
+field of the label dictionary take in your pipeline.
+For example, a box in `xywh` format at coordinates (100, 100) with a width of 55 and a height of 70 would be represented by:
```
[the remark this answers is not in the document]
LukeWood: Upper left for XY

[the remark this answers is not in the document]
LukeWood: center_xy is middle

guides/keras_cv/object_detection_keras_cv.py
```diff
@@ -0,0 +1,868 @@
...
+
+"""
+To use the `RetinaNet` architecture, you'll need to resize your image
+to a size that is divisible by 64.
+In object detection, the approach to resizing images to comply
+with this constraint is *extremely* important.  If the resize operation distorts
+the input's aspect ratio, the model will perform signficantly poorer.  For the
+pretrained `"retinanet_resnet50_pascalvoc"` preset we are using, the final
+`MeanAveragePrecision` on the `pascalvoc/2012` evaluation set drops to `0.15`
+from `0.33` when using a naive resizing operation.
```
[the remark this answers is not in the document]
LukeWood: Removed a sentence and think its a bit clearer now

guides/keras_cv/object_detection_keras_cv.py
```diff
@@ -0,0 +1,868 @@
...
+"""
+
+## Train a Custom Object Detection Model
+
+![](https://storage.googleapis.com/keras-nlp/getting_started_guide/prof_keras_advanced.png)
+
+Whether you're an object detection amateur or a well seasoned veteran, assembling
+an object detection pipeline from scratch is a massive undertaking.
+Luckily, all KerasCV object detection APIs are built as modular components.
+Whether you need a complete pipeline, an object detection pipeline, or even just
```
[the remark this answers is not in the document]
LukeWood: Meant to say mode, thanks!

guides/keras_cv/object_detection_keras_cv.py
```diff
@@ -0,0 +1,868 @@
...
+
+In this guide, we'll assemble a full training pipeline for a KerasCV object
+detection model.  This includes data loading, augmentation, metric evaluation,
+and inference!
+
+To get started, let's sort out all of our imports and define global
+configuration parameters.
+"""
+
+BATCH_SIZE = 4
```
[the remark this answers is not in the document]
LukeWood: Ah, sorry this was needed to open multiple TFDS datasets.  For just 2007 its not needed

guides/keras_cv/object_detection_keras_cv.py
```diff
@@ -0,0 +1,868 @@
...
+low, high = resource.getrlimit(resource.RLIMIT_NOFILE)
+resource.setrlimit(resource.RLIMIT_NOFILE, (high, high))
+
+"""
+## Data loading
+
+To get started, let's discuss data loading and bounding box formatting.
+KerasCV has a predefined specificication for bounding boxes.
```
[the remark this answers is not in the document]
LukeWood: or just "format". updated

guides/keras_cv/object_detection_keras_cv.py
```diff
@@ -0,0 +1,868 @@
...
+pipelines is data augmentation.  Image augmentation techniques must be aware of the underlying
+bounding boxes, and must update them accordingly.
+
+Luckily, KerasCV natively supports bounding box augmentation with its extensive
+library
+of [data augmentation layers](https://keras.io/api/keras_cv/layers/preprocessing/).
+The code below loads the Pascal VOC dataset, and performs on-the-fly bounding box
+friendly data augmentation inside of a `tf.data` pipeline.
+"""
+augmenter = keras.Sequential(
```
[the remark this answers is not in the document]
LukeWood: Actually yes

guides/keras_cv/object_detection_keras_cv.py
LukeWood: bounding box is a namespace!

guides/keras_cv/object_detection_keras_cv.py
LukeWood: maybe in a follow up, for now lets leave it.

BatchNorm is something I think Keras in general could handle a bit better - (sync batch norm, normal batch norm, freezing, unfreezing)

guides/keras_cv/object_detection_keras_cv.py
LukeWood: Now that I moved this model doesn't exist anymore at this cell 👍

guides/keras_cv/object_detection_keras_cv.py
LukeWood: It just helps performance to make it 1e9, we only call result() once in this callback per epoch.  Thats the only place we read the metrics from.

guides/keras_cv/object_detection_keras_cv.py
LukeWood: I see what you are saying here, but I was struggling to come up with a better solution.

LukeWood:
Ok @jbischof @fchollet @ianstenbit ready for another round!

LukeWood:
> This is SO much better, thank you!
> 
> My biggest remaining suggestion is to essentially do everything with the pretrained model FIRST and only at the end show the extra steps needed to train one from scratch. It's fairly confusing to keep going back and forth between the pretrained and random models.
> 
> I have trouble reviewing the raw python personally because I can't see any of the images produced by the runtime. Would love to see a notebook version.

I will render one for you now.

LukeWood:
The visual results somehow got worse on my last training run despite the MaP remaining 0.33, I am going to try again w/o freezing batchnorm.  🤔

LukeWood:
I removed the tuning your prediction decoding section, because I realized my approach was nonsense.  You will NEVER get worse results by allowing your model to make more predictions if you sort by confidence score, so I was basically just selecting the most "flexible" parameterization.  Updated now

LukeWood:
Ok, @jbischof and @fchollet I think its finally ready!!!!!

LukeWood:
@jbischof to see the preview, please visit

https://colab.sandbox.google.com/github/LukeWood/keras-io/blob/odguidev2/guides/ipynb/keras_cv/object_detection_keras_cv.ipynb

You will need to swap out the pip install for:
```
!pip install --upgrade git+https://github.com/keras-team/keras-cv@rnv2
```

LukeWood:
Ok, need to re-train the model one more time with the softmax->sigmoid fix!

guides/keras_cv/object_detection_keras_cv.py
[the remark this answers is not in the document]
jbischof: It's still not clear to me why it has to be divisible by 64. Sorry if I'm being dense.

guides/keras_cv/object_detection_keras_cv.py
```diff
@@ -0,0 +1,859 @@
...
+    y_pred=y_pred,
+    scale=5,
+    font_scale=0.7,
+    bounding_box_format="xywh",
+    class_mapping=class_mapping,
+)
+
+"""
+In order to support easy this easy and intuitive inference workflow, KerasCV
+perform non-max suppression inside of the `RetinaNet` class.
```
jbischof, line 245: s/perform/performs

guides/keras_cv/object_detection_keras_cv.py
```diff
@@ -0,0 +1,846 @@
...
+        num_parallel_calls=tf.data.AUTOTUNE,
+    )
+    return ds
+
+
+train_ds = load_pascal_voc(
+    split="train", dataset="voc/2007", bounding_box_format="xywh"
+)
+eval_ds = load_pascal_voc(split="test", dataset="voc/2007", bounding_box_format="xywh")
+eval_ds = load_pascal_voc(dataset="voc/2007", split="test", bounding_box_format="xywh")
```
jbischof, line 410: You have this line twice

guides/keras_cv/object_detection_keras_cv.py
```diff
@@ -0,0 +1,846 @@
...
+"""
+
+train_ds = train_ds.ragged_batch(BATCH_SIZE, drop_remainder=True)
+eval_ds = eval_ds.ragged_batch(BATCH_SIZE, drop_remainder=True)
+
+"""
+Let's make sure our dataset is following the format KerasCV expects.
+By using the `visualize_dataset()` function, you can visually verify
+that your data is in the format that KerasCV expects.  If the bounding boxes
+are not visible, or are visible in the wrong locations that is a sign that your
```
jbischof, line 434: delete comma

guides/keras_cv/object_detection_keras_cv.py
```diff
@@ -0,0 +1,846 @@
...
+"""
+
+inference_resizing = keras_cv.layers.Resizing(
+    640, 640, bounding_box_format="xywh", pad_to_aspect_ratio=True
+)
+eval_ds = eval_ds.map(inference_resizing, num_parallel_calls=tf.data.AUTOTUNE)
+
+"""
+Due to the fact that the resize operation differs between the train dataset,
+which uses `JitteredResize()` to resize images, and the inference dataset which
```
jbischof, line 502: comma after "inference dataset".

You should likely start a new sentence with "It is good practice..."

guides/keras_cv/object_detection_keras_cv.py
```diff
@@ -0,0 +1,846 @@
...
+    ]
+)
+
+train_ds = train_ds.map(augmenter, num_parallel_calls=tf.data.AUTOTUNE)
+visualize_dataset(
+    train_ds, bounding_box_format="xywh", value_range=(0, 255), rows=2, cols=2
+)
+
+"""
+Great!  We now have a bounding box friendly data augmentation pipeline.
```
jbischof, line 489: When multiple words create a single adjective you should add dashes for clarity:
* state-of-the-art
* out-of-the-box

guides/keras_cv/object_detection_keras_cv.py
jbischof: Never would have guessed!

guides/keras_cv/object_detection_keras_cv.py
```diff
@@ -0,0 +1,846 @@
...
+eval_ds = eval_ds.map(inference_resizing, num_parallel_calls=tf.data.AUTOTUNE)
+
+"""
+Due to the fact that the resize operation differs between the train dataset,
+which uses `JitteredResize()` to resize images, and the inference dataset which
+uses `layers.Resizing(pad_to_aspect_ratio=True)`, it is good practice to
+visualize both datasets:
+"""
+
+visualize_dataset(
```
jbischof, line 507: I don't see any outputs in the notebook for this

guides/keras_cv/object_detection_keras_cv.py
```diff
@@ -0,0 +1,846 @@
...
+    eval_ds, bounding_box_format="xywh", value_range=(0, 255), rows=2, cols=2
+)
+
+"""
+Finally, let's unpackage our inputs from the preprocessing dictionary, and
```
jbischof, line 512: Does this mean that the notebook is not GPU compatible by default?

guides/keras_cv/object_detection_keras_cv.py
jbischof: Should we call it `pretrained_model`?

guides/keras_cv/object_detection_keras_cv.py
```diff
@@ -0,0 +1,846 @@
...
+
+stable_diffusion = keras_cv.models.StableDiffusionV2(512, 512)
+images = stable_diffusion.text_to_image(
+    prompt="A zoomed out photograph of a cool looking cat.  The cat stands in a beautiful forest",
+    negative_prompt="unrealistic, bad looking, malformed",
+    batch_size=4,
+    seed=1231,
+)
+y_pred = model.predict(images)
+visualization.plot_bounding_box_gallery(
```
jbischof, line 836: Still don't see output

jbischof (approved):
Looks great! Really high quality.

The biggest issue remaining is that I don't see output for a lot of print and visualization statements. I hope that can be remedied before pushing. The rest is mainly grammatical.

guides/keras_cv/object_detection_keras_cv.py
LukeWood: No it is GPU compatible either way

guides/keras_cv/object_detection_keras_cv.py
```diff
@@ -0,0 +1,846 @@
...
+    classification_loss="focal",
+    box_loss="smoothl1",
+    optimizer=optimizer,
+    metrics=[coco_metrics],
+)
+coco_metrics.reset_state()
+result = pretrained_model.evaluate(eval_ds.take(40), verbose=0)
+result = coco_metrics.result(force=True)
+
+print_metrics(result)
```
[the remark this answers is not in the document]
LukeWood: Notebooks don't include output on Keras.io (confirmed by @fchollet ).

this is done to simplify the process of rendering/including them on the website.

guides/keras_cv/object_detection_keras_cv.py
```diff
@@ -0,0 +1,848 @@
...
+    cols=1,
+    y_pred=y_pred,
+    scale=5,
+    font_scale=0.7,
+    bounding_box_format="xywh",
+    class_mapping=class_mapping,
+)
+
+"""
+That looks a lot better!
```
jbischof, line 296: This is quite confusing because NMS is set up as *suppressing* noisy detections (false positives) and instead it is recovering true positives. Is this a good example?

LukeWood: Updated the picture to reference the doc picture instead!  Thanks!

guides/img/object_detection_keras_cv/object_detection_keras_cv_16_1.png
jbischof, line 1: Shouldn't this file be out?

guides/keras_cv/object_detection_keras_cv.py
```diff
@@ -0,0 +1,848 @@
...
+    class_mapping=class_mapping,
+)
+
+"""
+In order to support easy this easy and intuitive inference workflow, KerasCV
+performs non-max suppression inside of the `RetinaNet` class.
+Non-max suppression is a traditional computing algorithm that solves the problem
+of a model detecting multiple boxes for the same object.
+
+**Example of pre-NMS detections:**
```
jbischof, line 252: Couldn't this come from the API itself? We just have to overwrite the prediction decoder twice, right?

guides/keras_cv/object_detection_keras_cv.py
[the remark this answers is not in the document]
jbischof: Looks great thanks!

guides/keras_cv/object_detection_keras_cv.py
LukeWood: it could, but I feel its weird to write to it for that demonstration purpose.  I wrote it out and it just made the guide look a bit strange.   Also, this isnt a truly non-nmsed image, its just a "barely nmsed" image.  Feel its burdening the user with maybe more logic than they need

guides/img/object_detection_keras_cv/object_detection_keras_cv_16_1.png
LukeWood: out of the guide?  removed?  This is still shown before customizing the decoder.  I feel we should update the default decoder now that my softmax->sigmoid fix is in.

guides/keras_cv/object_detection_keras_cv.py
LukeWood: actually, this is better !  Updated

fchollet (approved):
LGTM

guides/keras_cv/object_detection_keras_cv.py
```diff
@@ -0,0 +1,872 @@
...
+supported in all KerasCV preprocessing components.
+
+Let's load some data and verify that the data looks as we expect it to.
+"""
+
+
+def visualize_dataset(inputs, value_range, rows, cols, bounding_box_format):
+    inputs = next(iter(inputs.take(1)))
+    images, bounding_boxes = inputs["images"], inputs["bounding_boxes"]
+    visualization.plot_bounding_box_gallery(
```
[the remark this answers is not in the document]
martin-gorner: Ideally, you should not have to create a visualize_dataset function. This should be enough:
plot_bounding_box_gallery(ds.get_sigle_element(), value_range=(0,255), bounding_box_format="xywh", class_mapping=...)

guides/keras_cv/object_detection_keras_cv.py
```diff
@@ -0,0 +1,872 @@
...
+[`keras.callbacks.ReduceLROnPlateau`](https://keras.io/api/callbacks/reduce_lr_on_plateau/)
+callback to reduce the learning rate.
+
+You will always want to include a `global_clipnorm` when training object
+detection models.  This is to remedy exploding gradient problems that frequently
+occur when training object detection models.
+"""
+
+base_lr = 0.005
+# including a global_clipnorm is extremely important in object detection tasks
```
martin-gorner, line 573: In my experience, a global clipnorm will prevent training runs from going to NaNs, but they will still be bad. This will show as bumpy training curves. Essentially, the clip norm puts the training back on track when it diverges but the training epochs up to that point are lost. Once you have found correct LR settings, global clipnorm is no longer necessary. You can look at all the "arthropods" samples, even the 32-core TPU pod one using a batch size of 1024. They all converge nicely and none use global clipnorm.

guides/keras_cv/object_detection_keras_cv.py
```diff
@@ -0,0 +1,872 @@
...
+
+train_ds = train_ds.prefetch(tf.data.AUTOTUNE)
+eval_ds = eval_ds.prefetch(tf.data.AUTOTUNE)
+
+"""
+
+### Optimizer
+
+In this guide, we use a standard SGD optimizer and rely on the
+[`keras.callbacks.ReduceLROnPlateau`](https://keras.io/api/callbacks/reduce_lr_on_plateau/)
```
martin-gorner, line 564: There is no ReduceLROnPlateau in your .fit() !

guides/keras_cv/object_detection_keras_cv.py
```diff
@@ -0,0 +1,872 @@
...
+    print("-" * (maxlen + 1))
+    for k, v in metrics.items():
+        print(f"{k.ljust(maxlen+1)}: {v.numpy():0.2f}")
+
+
+"""
+Due to the high computational cost of computing COCO metrics, the KerasCV
+`BoxCOCOMetrics` component requires an `evaluate_freq` parameter to be passed to
+its constructor.  Every `evaluate_freq`-th call to `update_state()`, the metric
+will recompute the result.  In between invocations, a cached version of the
```
martin-gorner, line 639: "cached version": essentially, the metric is returning a wrong value most of the time. This is really not the ideal solution. Please implement a proper "eval-every-N-epochs" functionality instead.

guides/keras_cv/object_detection_keras_cv.py
```diff
@@ -0,0 +1,872 @@
...
+Luckily, there are two workarounds that allow you to still train a RetinaNet on TPU:
+
+- The use of a custom callback
+- Using a [SideCarEvaluator](https://www.tensorflow.org/api_docs/python/tf/keras/utils/SidecarEvaluator)
+
+Let's use a custom callback to achieve TPU compatibility in this guide:
+"""
+
+
+class EvaluateCOCOMetricsCallback(keras.callbacks.Callback):
```
martin-gorner, line 675: This is pretty bad. It means that all KerasCV OD examples will be TPU-incompatible out of the box.

guides/keras_cv/object_detection_keras_cv.py
```diff
@@ -0,0 +1,872 @@
...
+loss, as the majority of the losses are assigned to the background class.
+
+"SmoothL1 Loss" is used to [prevent exploding gradients](https://arxiv.org/abs/1504.08083)
+that often occur when attempting to perform the box regression task.
+
+In KerasCV you can use these losses simply by passing the strings `"focal"` and
+`"smoothl1"` to `compile()`:
+"""
+
+pretrained_model.compile(
```
martin-gorner, line 603: you are compiling your model twice before training it. Consider moving this "losses" block after the instantiation of the model, where a .compile step is relevant.

guides/keras_cv/object_detection_keras_cv.py
```diff
@@ -0,0 +1,872 @@
...
+    # Run for 10-35~ epochs to achieve good scores.
+    epochs=1,
+    callbacks=[EvaluateCOCOMetricsCallback(eval_ds.take(20))],
+)
+"""
+
+## Inference and plotting results
+
+KerasCV makes object detection inference simple.  `model.predict(images)`
+returns a RaggedTensor of bounding boxes.  By default, `RetinaNet.predict()`
```
martin-gorner, line 763: can you be more explicit about the retuned format here ? There are classes too aren't there ?

martin-gorner (commented):
This is a great tutorial. Thank you Luke for writing it!

Top comments (P0):
- Soft NMS is a top priority in object Detection
- There should be ONE format for images with bounding boxes and make sure everything in KerasCV uses it and is compatible: preprocessing layers, models, visualization utilities etc.
- JitteredResize needs renaming and width and height arguments before release to align it on other augmentation layers.
- remove the evaluate_freq arg from BoxCOCOMetric which essentially has the metric return wrong values, and implement a proper "ecval-every-N-epochs" functionality. (evaluate_freq can be a short-term workaround if needed)

Some other important stuff (P1):
- TPU compatibility ? Can we make it so that people do not have to write their own custom callback ?
- include_rescaling missing in RetinaNet instantiation ?
- educate people about the rampup-sustain-decay LR schedule for fine-tuning OD

One additional change that is cosmetic but will remove many lines that have no educational value: make sure plot_bounding_box_gallery does not need the scale, font_scale, rows and cols arguments out of the box.

LukeWood:
> This is a great tutorial. Thank you Luke for writing it!
> 
> Top comments (P0):
> 
> * Soft NMS is a top priority in object Detection
> * There should be ONE format for images with bounding boxes and make sure everything in KerasCV uses it and is compatible: preprocessing layers, models, visualization utilities etc.
> * JitteredResize needs renaming and width and height arguments before release to align it on other augmentation layers.
> * remove the evaluate_freq arg from BoxCOCOMetric which essentially has the metric return wrong values, and implement a proper "ecval-every-N-epochs" functionality. (evaluate_freq can be a short-term workaround if needed)
> 
> Some other important stuff (P1):
> 
> * TPU compatibility ? Can we make it so that people do not have to write their own custom callback ?
> * include_rescaling missing in RetinaNet instantiation ?
> * educate people about the rampup-sustain-decay LR schedule for fine-tuning OD
> 
> One additional change that is cosmetic but will remove many lines that have no educational value: make sure plot_bounding_box_gallery does not need the scale, font_scale, rows and cols arguments out of the box.

Thank you!  I will create a follow up PR for these.

innat-asj:
Hi, @LukeWood thanks for making such wonderful instructions. I have a couple of opinions that you might want to consider in the follow-up PR. 

1. The title (**Object Detection with KerasCV**) should be renamed with **2D Object ..**, if 3D OD should be in separated docs. 
2. `.from_pretset` API might be a new look to Keras users. Also, how many backbones are supported currently, and how to approach adding a custom backbone (i.e. output of backbone format) might be good if discussed in the doc.
3. Please use mixed precision, jit enabled, and GPU-memory growth at the beginning to showcase the feature complete document and encourage best practices. 
4. A complete TPU-compatible demonstration would be game-changing. This part should be emphasized more, (IMHO).
5. A demonstration of a custom data loader would be nice with tf.data API.
6. Augmentation like CutMix, MixUp, and RandAugment are mentioned in the description but aren't demonstrated in the code. Adding, if they can be part of the model to get GPU acceleration, no!
7. The parameter of the `model. compile` may need some discussion, i.e. loss, classification_loss, box_loss, etc, lots of params!
8. ..Would be nice if coco_metrics output can be formatted in a table in mode. fit log, no?
10. FasterRCNN should be discussed too as it's present at the time of writing the doc.
11. Saving the trained OD model with the latest format (`.keras`) and reloading successfully. (pls check https://github.com/keras-team/keras-io/issues/1235) 
12. Performance (speed/accuracy) compared to existing models!

---

For TPU, it's written as follows, where the same validation data set (`eval_ds` is passed to the model again via callback, even though it's already processed via `validation_data`). It reminds me of this feature request https://github.com/keras-team/keras/issues/17350 .
cc. @rchao 

```
model.fit(
    train_ds.take(20),
    validation_data=eval_ds.take(20),
    # Run for 10-35~ epochs to achieve good scores.
    epochs=1,
    callbacks=[EvaluateCOCOMetricsCallback(eval_ds.take(20))],
)
```

jbischof:
Yes @innat eventually we will write a "Getting Started" guide for KerasCV. We're waiting for a variety of tasks to be ready so that it is more than a few lines 😄

jbischof:
I will add @LukeWood that I still find that [image](https://keras.io/img/guides/object_detection_keras_cv/object_detection_keras_cv_16_1.png) with NMS set too high pretty confusing. I would just remove

innat-asj:
@LukeWood 

Thanks for the prompt response. 

> Please use mixed precision, jit enabled, and GPU-memory growth at the beginning to showcase the feature complete document and encourage best practices.
Thanks! Can you file a bug on kerasCV to verify this behavior for RetinaNet?

I tried mixed precision, the `keras_cv.layers.Resizing` caused some issues; but `JirtterResize` passed. Maybe bug!

```
filepath = tf.keras.utils.get_file(origin="https://i.imgur.com/gCNcJJI.jpg")
image = keras.utils.load_img(filepath)
image = np.array(image)

inference_resizing = keras_cv.layers.Resizing(
    640, 640, 
    pad_to_aspect_ratio=True, 
    bounding_box_format="xywh"
)

jittered_resize = keras_cv.layers.JitteredResize(
    target_size=(640, 640),
    scale_factor=(0.8, 1.25),
    bounding_box_format="xywh",
    )

keras_resize = keras.layers.Resizing(
    640, 640, crop_to_aspect_ratio=False
)

image_batch = jittered_resize([image]) # OK
image_batch = keras_resize([image]) # OK 
image_batch = inference_resizing([image]) # Error

InvalidArgumentError: Exception encountered when calling layer 'resizing' (type Resizing).

TensorArray dtype is float16 but Op is trying to write dtype float32 
guide
Call arguments received by layer 'resizing' (type Resizing):
  • inputs=['tf.Tensor(shape=(3024, 4032, 3), dtype=uint8)']
```

> Perhaps this is deserving of its own guide? The guide is TPU-friendly, so there's really no delta needed here. What would you like to see to show this off?

Agree, this might need a separate guide (maybe like [mult-device](https://keras.io/guides/distributed_training/) training/inference doc). Sometimes it happens that a well-optimized code perfectly run on GPU (s) but fails to run on TPU (hard experience from Kaggle -_-). That's why I was suggesting if there's a full-proof demonstration that would be really great (discussing possible issues and workaround i.e. like you did `to_ragged() to to_dense()`).

> For TPU, it's written ..
model.fit(
    train_ds.take(20),
    validation_data=eval_ds.take(20),
    # Run for 10-35~ epochs to achieve good scores.
    epochs=1,
    callbacks=[EvaluateCOCOMetricsCallback(eval_ds.take(20))],
)
Not sure how to fix this last one.

Understood. Right now the above way is the only standard way. This might become too heavy if `eval_ds` is large and `model` is big enough. I needed something similar which lead to this FR https://github.com/keras-team/keras/issues/17350.

PR #1325, opened 2023-04-06
Add missing dependency and make pip commands runnable on Google Colab

seungjun-green:
Without installing the 'rouge-score' package using 'pip install rouge-score', attempting to run 'rouge_l = keras_nlp.metrics.RougeL()' will result in an error message: 'ImportError: RougeL requires the rouge_score package. Please install it with pip install rouge-score.'


fchollet (commented):
Thanks for the PR! Please also replicate the fix in the md and ipynb files.

fchollet (approved):

PR #1334, opened 2023-04-19
Fix incorrectly rendered files

haifeng-jin:
fix the incorrectly rendered files in #1133.
The images are not rendered in https://keras.io/examples/nlp/multi_label_classification/#train-the-model.


fchollet (approved):
LGTM, thanks

PR #1346, opened 2023-04-30
Update image_captioning.py

aadesh-kumar:
The " <end>" token had a leading space which never matches with the vocabulary token "<end>" leading to bad inference.


fchollet (commented):
Thanks for the PR! Please replicate the fix in the ipynb and md file.

aadesh-kumar:
Sorry for a late response. I have replicated the changes to ipynb and md files.

Issue #1332, opened 2023-04-15
Add example - Keras Reproducibility Recipes


Frightera:
I am eager to write a tutorial for `Keras Reproducibility Recipes`. Here are some questions for this:
1) Do we always use the latest version of Keras in the examples? Because there will be some differences in reproducibility techniques between Keras versions (version <= Keras 2.7 and version > 2.7)

2) Does this fall under the category of quick recipes?i

When I have the confirmation, I can start writing this guide.

hertschuh:
@Frightera , 

1. Yes, you should use the latest Keras. And 2.7 is really old now.

2. Can you provide more details on what you mean by `Keras Reproducibility Recipes`?

Thanks!

Frightera:
@hertschuh 

I got the idea from: https://github.com/keras-team/keras-io/blob/master/call_for_contributions.md#keras-reproducibility-recipes

My goal is to provide a comprehensive guide that enables users to obtain consistent results in every run of their deep learning experiments. This issue is frequently raised in various Stackoverflow questions.

hertschuh:
Ok, yeah, I think quick recipes makes sense, but we can see once it's ready if it fits better somewhere else.

Thank you for contributing!

Frightera:
Hi @hertschuh, 

PR is ready.

Frightera:
Closing as PR is merged.

PR #1353, opened 2023-05-05
[Add Example] Create reproducibility_recipes.py

Frightera:
This quick Keras recipe demonstrates how Keras initializers behave, and shows what's needed to achieve identical results in subsequential runs in model training process.

#1332

### DIFF
--- a/examples/keras_recipes/reproducibility_recipes.py
+++ b/examples/keras_recipes/reproducibility_recipes.py
@@ -180,8 +180,17 @@ test_ds = tf.data.Dataset.from_tensor_slices((test_images, test_labels))
 """
 Remember we called `tf.config.experimental.enable_op_determinism()` at the
 beginning of the function. This makes the `tf.data` operations deterministic.
-However, making `tf.data` operations deterministic comes with a performance cost.
+However, making `tf.data` operations deterministic comes with a performance 
+cost.
 If you want to learn more about it, please check this [official guide](https://www.tensorflow.org/api_docs/python/tf/config/experimental/enable_op_determinism#determinism_and_tfdata).
+
+Small summary what's going on here. Models have `kernel_initializer` and
+`bias_initializer` parameters. Since we set random seeds using
+`keras.utils.set_random_seed` in the beginning of the notebook, the initializers
+will produce same results in the sequential runs. Additionally, TensorFlow
+operations have now become deterministic. Frequently, you will be utilizing GPUs
+that have thousands of hardware threads which causes non-deterministic behavior
+to occur.
 """
 
 
--- a/examples/keras_recipes/reproducibility_recipes.py
+++ b/examples/keras_recipes/reproducibility_recipes.py
@@ -72,7 +72,7 @@ for initializer in initializers_list:
         # you need to specify a seed value. Note that this is not related to
         # keras.utils.set_random_seed or tf.config.experimental.enable_op_determinism.
         # If you comment those lines, you will still get the same results.
-        result = initializer(seed=42)(shape=(1, 1)).numpy()
+        result = float(initializer(seed=42)(shape=(1, 1)))
         print(f"\tIteration --> {iteration} // Result --> {result}")
     print("\n")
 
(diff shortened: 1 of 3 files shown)


examples/keras_recipes/reproducibility_recipes.py
```diff
@@ -0,0 +1,274 @@
...
+from tensorflow import keras
+from tensorflow.keras import layers
+
+# Set the seed using keras.utils.set_random_seed. This will set:
+# 1) `numpy` seed
+# 2) `tensorflow` random seed
+# 3) `python` random seed
+keras.utils.set_random_seed(812)
+
+# This will make TensorFlow ops as deterministic as possible.
```
LukeWood, line 34: perhaps mention the performance loss that follows from this to explain why its not on by default

examples/keras_recipes/reproducibility_recipes.py
```diff
@@ -0,0 +1,274 @@
...
+initializing the weights of layer variables. The following built-in initializers
+are available as part of `tf.keras.initializers`:
+"""
+to_be_removed = ["Initializer", "Ones", "Zeros", "Identity"]
+initializers_list = [
+    string
+    for string in dir(keras.initializers)
+    if string[0].isupper() and string not in to_be_removed
+]
+print(initializers_list)
```
LukeWood, line 51: can you explain this code a bit more in comments?  I'm not following the logic of "why" some of these are picked and others are not!

The goal is to remove these from the initializers module?

examples/keras_recipes/reproducibility_recipes.py
```diff
@@ -0,0 +1,274 @@
...
+        if "seed" in initializer_signature.parameters:
+            result = initializer(seed=42)(shape=(3, 3))
+        else:
+            result = initializer()(shape=(3, 3))
+        results[initializer_name].append(result)
+
+# Check if the results are equal.
+all_equal = True
+for initializer_name, initializer_results in results.items():
+    if not tf.experimental.numpy.allclose(
```
LukeWood, line 79: so we're not quite removing modules, we're just checking to ensure that these output identical results?  I think that is reasonable!

LukeWood (commented):
Perhaps a useful guide - I think we could greatly simplify this code and make it very concise by removing a lot of the module inspection code and some of the more complex logic loops!

Frightera:
@LukeWood Thanks for the review, I simplified those parts, can you take a look at when you have time?

examples/keras_recipes/reproducibility_recipes.py
```diff
@@ -0,0 +1,269 @@
...
+
+for initializer in initializers_list:
+    print(f"Running {initializer}")
+
+    for iteration in range(2):
+        # In order to get same results across multiple runs from an initializer,
+        # you need to specify a seed value. Note that this is not related to
+        # keras.utils.set_random_seed or tf.config.experimental.enable_op_determinism.
+        # If you comment those lines, you will still get the same results.
+        result = initializer(seed=42)(shape=(1, 1)).numpy()
```
fchollet, line 75: You could cast to float instead of using `numpy()` since it's a single element

examples/keras_recipes/reproducibility_recipes.py
```diff
@@ -0,0 +1,269 @@
...
+
+    # Resize the image
+    image = tf.image.resize(image, (32, 32))
+
+    return image, label
+
+
+# Prepare the datasets, batch-map --> vectorized operations
+train_data = (
+    train_ds.shuffle(buffer_size=len(train_images))
```
fchollet, line 207: Data order during training is important so shuffle should be explained

fchollet (commented):
Thank you for the PR! Looking good! 👍

Frightera:
Hi @fchollet, thank you for your comments. Now I provided more explanation for the `shuffle` process.

fchollet (approved):
Thank you! Everything looks good to me. You can add the generated files.

Frightera:
@fchollet Done! I think this is ready to be merged 😄 Here's a snippet:

image

fchollet (approved):
LGTM, thank you for the great contribution! 👍

PR #1358, opened 2023-05-10
Fixes for Input Image Normalization and Optimizer Reference

chunduriv:
There are a couple of fixes recommended to run `Face image generation with StyleGAN` tutorial with the latest Tensorflow version (2.12).

1) The input images are `normalized` twice, so it generates black images. To fix this, we should exclude  `ds_train = ds_train.map(lambda x: x / 255.0)`

2) Since we are trying to restore a checkpoint from a legacy keras optimizer into latest version, we should update the optimizer reference as shown below

```
d_optimizer=tf.keras.optimizers.legacy.Adam(**opt_cfg),
g_optimizer=tf.keras.optimizers.legacy.Adam(**opt_cfg),
```
Thank you!


chunduriv:
Hello @LukeWood @fchollet,

Could you kindly review this pull request?

Thank you!

PR #1361, opened 2023-05-11
Update keras-nlp content

mattdangerw:
Misc updates to better highlight keras-nlp content we care about...
- Link new gpt2 generation example.
- Add redirect links at old layer locations.
- Add a quicklink for keras-nlp models.
- Remove some content on the base models page so the "table" is front and center. The removed content exists on individual symbol docstrings in greater detail.


redirects/api/keras_nlp/layers/fnet_encoder/index.html
```diff
@@ -0,0 +1,2 @@
+
+<meta http-equiv="refresh" content="0; URL='https://keras.io/api/keras_nlp/modeling_layers/'" />
```
mattdangerw, line 2: Is there a way to test this out?

scripts/examples_master.py
```diff
@@ -615,11 +615,16 @@
                 },
                 # Text generation
                 {
-                    "path": "text_generation_gpt",
-                    "title": "GPT text generation with KerasNLP",
+                    "path": "gpt2_text_generation_with_kerasnlp",
```
mattdangerw, line 618: One weird thing to deal with, our guide is currently sitting under `nlp` not `generative`. That means with this change it will list twice! Should we just move it under generative?

mattdangerw: Moved for now, seems most consistent.

redirects/api/keras_nlp/layers/fnet_encoder/index.html
chenmoneygithub: this should be `/modeling_layers/fnet_encoder`?

No idea how to test - -"

chenmoneygithub (approved):
Thanks!

fchollet (approved):

fchollet (approved):

PR #1366, opened 2023-05-15
Update the KerasNLP translation example for keras-nlp 0.5.x

mattdangerw:
Replaces old `keras_nlp.utils.xx_search` API with `keras_nlp.samplers`.


fchollet (approved):
LGTM

PR #1391, opened 2023-05-31
Update link to timeseries dataset

fex95:
The dataset used has been moved by the creator and packaged in a slightly different structure.

On the example page, the link to the data set is also still incorrect.
https://keras.io/examples/timeseries/timeseries_traffic_forecasting/

Tested only with colab


nkovela1 (approved):

nkovela1:
@fchollet I do not have write access, please approve thanks!

fchollet (approved):
Thanks!

PR #1392, opened 2023-05-31
Adding header Accelerator field details

hamidriasat:
Every example header needs to have 6 rows. This condition is implemented in the code, but the 6th line should contain info about `Accelerator`, this thing was missing.

So I added additional checks with existing header conditions to verify two things

1. 6th row contains info about `Accelerator`
2. Its value is one of the possible values `GPU`, `TPU`, or `None`

Info about `Accelerator` and its possible values is also added in `README`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,6 +78,7 @@ Author: (could be `Authors`: as well, and may contain markdown links)
 Date created: (date in yyyy/mm/dd format)
 Last modified: (date in yyyy/mm/dd format)
 Description: (one-line text description)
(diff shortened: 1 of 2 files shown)


nkovela1:
Hi @hamidriasat, all keras.io tutorials should be able to work on all these accelerators by default. If there is an exception, that can be marked. Is there a specific reason why the accelerator field is needed?

fchollet (approved):
Thanks!

PR #1397, opened 2023-06-07
fix broken links in sequential_model.py

bombasticman:
hyperlink missing /keras/
edit: also /guides/ should be /guide/


qlzh727:
https://keras.io/guides/training_with_built_in_methods/ seems to the correct URL, and https://keras.io/guide/keras/training_with_built_in_methods/ doesn't work.

bombasticman:
my bad I suggested these changes for the tensorflow.org:
https://www.tensorflow.org/guide/keras/sequential_model
I didn't realize I was on keras.io

PR #1398, opened 2023-06-08
Remove extra activation layer

### LINKED ISSUE #254: A minor issue: Extra Activation Layer
In this tutorial: 
https://github.com/keras-team/keras-io/blob/master/examples/vision/oxford_pets_image_segmentation.py#L119

This line of code is redundant and can be removed:
https://github.com/keras-team/keras-io/blob/678471522035c988f98c82ce86e4c892d40163a3/examples/vision/oxford_pets_image_segmentation.py#L119

Cause there is the following line right after the above line:
https://github.com/keras-team/keras-io/blob/678471522035c988f98c82ce86e4c892d40163a3/examples/vision/oxford_pets_image_segmentation.py#L125

So, there is an extra activation layer in the model.

sachinprasadhs:
Closes #254


fchollet (commented):
Thanks! Please replicate the fix in the md and ipynb files.

fchollet (approved):
LGTM

PR #1410, opened 2023-06-23
remove unwanted code from transformer_asr.py

### LINKED ISSUE #1394: `transformer_asr.py`: positional embedding unutilized in source CNN downsampler.
# Issue
As stated in the title, the positional embedding `self.pos_emb` in the class `SpeechFeatureEmbedding` is initialized but not used, apparently. This is a curious case, though, as it seems the model is still able to perform well without the positional embedding. Makes me wonder if CNNs are able to extract certain positional information into their outputs.

# Problem Code

```python
class SpeechFeatureEmbedding(layers.Layer):
    def __init__(self, num_hid=64, maxlen=100):
        super().__init__()
        self.conv1 = tf.keras.layers.Conv1D(
            num_hid, 11, strides=2, padding="same", activation="relu"
        )
        self.conv2 = tf.keras.layers.Conv1D(
            num_hid, 11, strides=2, padding="same", activation="relu"
        )
        self.conv3 = tf.keras.layers.Conv1D(
            num_hid, 11, strides=2, padding="same", activation="relu"
        )
        self.pos_emb = layers.Embedding(input_dim=maxlen, output_dim=num_hid)

    def call(self, x):
        x = self.conv1(x)
        x = self.conv2(x)
        return self.conv3(x)
```
Lastly, I would like to extend a thank you to @apoorvnandan for creating this code example. It has been a great resource for leaning.

SuryanarayanaY:
In the example tutorial of `Automatic Speech Recognition with Transformer`, positional embedding `self.pos_emb` in the class `SpeechFeatureEmbedding` has initialised but not used anywhere.

Hence I am proposing to remove this line of code to avoid unnecessary confusion.

Fixes #1394 

Thanks

### DIFF
--- a/examples/audio/ipynb/transformer_asr.ipynb
+++ b/examples/audio/ipynb/transformer_asr.ipynb
@@ -110,7 +110,6 @@
     "        self.conv3 = tf.keras.layers.Conv1D(\n",
     "            num_hid, 11, strides=2, padding=\"same\", activation=\"relu\"\n",
     "        )\n",
-    "        self.pos_emb = layers.Embedding(input_dim=maxlen, output_dim=num_hid)\n",
     "\n",
     "    def call(self, x):\n",
     "        x = self.conv1(x)\n",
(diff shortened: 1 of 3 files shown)


fchollet (commented):
LGTM, thank you!

fchollet (approved):

PR #1418, opened 2023-07-06
Remove extra activation layer in image_classification_from_scratch.py

### LINKED ISSUE #1119: Problems with the image classification example
According to the code, x is activated twice when input into the residual block right?
Screenshot 2022-10-19 114157

SuryanarayanaY:
The `image_classification_from_scratch.py` tutorial has one extra `relu activation layer` which is being applied on the layer that already gone through relu activation which makes one code of line redundant. Though it may not affect output it unnecessarily creates confusion for the users and also consumes some resources if used for training.Hence I am proposing to remove the unwanted line of code.

Fixes #1119

### DIFF
--- a/examples/vision/image_classification_from_scratch.py
+++ b/examples/vision/image_classification_from_scratch.py
@@ -235,11 +235,11 @@ def make_model(input_shape, num_classes):
     x = layers.Rescaling(1.0 / 255)(inputs)
     x = layers.Conv2D(128, 3, strides=2, padding="same")(x)
     x = layers.BatchNormalization()(x)
-    x = layers.Activation("relu")(x)
 
     previous_block_activation = x  # Set aside residual
 
     for size in [256, 512, 728]:
+        x = layers.Activation("relu")(x)
         x = layers.SeparableConv2D(size, 3, padding="same")(x)
         x = layers.BatchNormalization()(x)
 
--- a/examples/vision/image_classification_from_scratch.py
+++ b/examples/vision/image_classification_from_scratch.py
@@ -240,7 +240,6 @@ def make_model(input_shape, num_classes):
     previous_block_activation = x  # Set aside residual
 
     for size in [256, 512, 728]:
-        x = layers.Activation("relu")(x)
         x = layers.SeparableConv2D(size, 3, padding="same")(x)
         x = layers.BatchNormalization()(x)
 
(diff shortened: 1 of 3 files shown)


divyashreepathihalli (commented):
Thank you for the PR. It is recommended to remove the first call at line 238 instead of the one in the loop.

SuryanarayanaY:
Actually line no.240 `previous_block_activation = x ` (where x = layers.Activation("relu")(x) defined at line no.238 ) stored under variable `previous_block_activation` as residual and used it later at Line No.254.

https://github.com/keras-team/keras-io/blob/8c48c9f892373c571eae92a77b06ecdecb47520f/examples/vision/image_classification_from_scratch.py#L238-L256

divyashreepathihalli:
@SuryanarayanaY in that case this change will break the implementation. I will close this PR. Thanks!

SuryanarayanaY:
@divyashreepathihalli ,

I am thinking in perspective of user request for removing extra layer. But here issue seems more than that. Here author used activation layer output as residual. I think it's better to pass residuals without activation. In that case removing the first activation call at line no.138 itself makes sense. Like the amendment 

If the author intention is to pass activation layer output as residual then earlier proposal is the solution. Anyways passing already activated output to same activation layer again gives same result.Hence the proposed changes will not break. I have tested the code with earlier proposal and it executes fine as per attached [gist](https://colab.sandbox.google.com/gist/SuryanarayanaY/dee2100f0186706934d876ba72894510/modified_-image_classification_from_scratch.ipynb). Correct me if I am missing anything here.

I think what we need to pay attention more here is whether we have to pass residuals as with relu activation or without activation. If we need to pass residuals without going through activation then we need to remove line no.238 as proposed by you.

Please suggest.

SuryanarayanaY:
@divyashreepathihalli ,

Can I go ahead with the proposal you referred earlier in this [comment](https://github.com/keras-team/keras-io/pull/1418#pullrequestreview-1528932255) ?

divyashreepathihalli:
> @divyashreepathihalli ,
> 
> Can I go ahead with the proposal you referred earlier in this [comment](https://github.com/keras-team/keras-io/pull/1418#pullrequestreview-1528932255) ?

Sure, please go ahead and update the changes.

SuryanarayanaY:
@divyashreepathihalli , Done the changes.May please review.

gbaned:
Hi @divyashreepathihalli, Can you please review this PR ? Thank you!

divyashreepathihalli (approved):

gbaned:
Hi @fchollet, Can you please review this PR ? Thank you!

sachinprasadhs:
Fines are outdated, can you please make the changes to the latest file https://keras.io/examples/vision/image_classification_from_scratch/

PR #1421, opened 2023-07-07
fixed internal link

tricki:
First content link here: https://keras.io/api/keras_nlp/models/


tricki:
Take it or leave it, I'm not going to sign anything for adding a single slash. I promise I won't sue if you steal it.

fchollet (approved):

PR #1422, opened 2023-07-08
KerasNLP: Abstractive Summarization with BART

abheesht17:


examples/nlp/abstractive_summarization_with_bart.py
```diff
@@ -0,0 +1,249 @@
...
+"""
+## Setup
+
+Before we start implementing the pipeline, let's install and import all the
+libraries we need. We'll be using the KerasNLP library. We will also need a
+couple of utility libraries.
+"""
+
+"""shell
+pip install rouge-score -q
```
mattdangerw, line 41: we can probably make this install line a one liner, at least after we land whatever fixes we need for keras-nlp

examples/nlp/abstractive_summarization_with_bart.py
```diff
@@ -0,0 +1,249 @@
...
+
+import gdown
+import py7zr
+import time
+
+import keras_nlp
+import tensorflow as tf
+import tensorflow_datasets as tfds
+
+from tensorflow import keras
```
mattdangerw, line 55: we can try rewriting this to keras-core (can ditch mixed precision for now while we debug that if it helps)

examples/nlp/abstractive_summarization_with_bart.py
```diff
@@ -0,0 +1,249 @@
...
+"""
+## Dataset
+
+Let's load the [SAMSum dataset](https://arxiv.org/abs/1911.12237). This dataset
+contains around 15,000 pairs of conversations/dialogues and summaries.
+"""
+
+# Download the dataset.
+gdown.download(
+    "https://drive.google.com/uc?id=1u4L2tgfbd5RLYPws47LTE9GeIJ6bZA9g",
```
mattdangerw, line 77: where is this drive link from? can we not use keras.utils.get_file?

examples/nlp/abstractive_summarization_with_bart.py
```diff
@@ -0,0 +1,249 @@
...
+expects the input format to be.
+"""
+
+train_ds = (
+    samsum_ds.map(
+        lambda dialogue, summary: {"encoder_text": dialogue, "decoder_text": summary}
+    )
+    .batch(BATCH_SIZE)
+    .cache()
+    .prefetch(tf.data.AUTOTUNE)
```
mattdangerw, line 111: I think the prefetch call only really makes sense right before the model, e.g. after tokenization.

examples/nlp/abstractive_summarization_with_bart.py
```diff
@@ -0,0 +1,249 @@
...
+for dialogue, generated_summary, ground_truth_summary in zip(
+    dialogues[:5], generated_summaries[:5], ground_truth_summaries[:5]
+):
+    print("Dialogue:", dialogue)
+    print("Generated Summary:", generated_summary)
+    print("Ground Truth Summary:", ground_truth_summary)
+    print("=============================")
+
+"""
+Qualitatively, the generated summaries look awesome! Let's see if that's the
```
mattdangerw, line 227: Figuring out these metrics is going to be a little tricky. Since they are text manipulations we will need the tf backend (as torch and jax lack string types). What we could do is update our metrics in keras-nlp to convert any inputs with something like `tf.convert_to_tensor`, then this would work on all backends. Might need a little playing around.

Alternately, if we don't want the headache, we could just ditch this section for now.

mattdangerw (commented):
This looks good! Should we try converting it to a multi-backend example?

examples/nlp/abstractive_summarization_with_bart.py
abheesht17: Followed the manual download instructions here: https://www.tensorflow.org/datasets/catalog/samsum. The provided link intermittently fails with `wget` (and `keras.utils.get_file`, by extension). I uploaded it to my GDrive. `keras.utils.get_file` might not work on a GDrive link (wget wasn't working), probably because GDrive allows downloading files below a certain size

examples/nlp/abstractive_summarization_with_bart.py
abheesht17: Let's ditch it for now, then. I'll open an issue on KerasNLP for porting metrics over

abheesht17:
@mattdangerw - this is ready!

examples/nlp/abstractive_summarization_with_bart.py
mattdangerw: Probably best to avoid this if we could. If the provided link intermittently fails, maybe try the huggingface location?

https://huggingface.co/datasets/samsum/blob/main/data/corpus.7z

```
filename = keras.utils.get_file(
    "corpus.7z",
    origin="https://huggingface.co/datasets/samsum/resolve/main/data/corpus.7z"
)
```

mattdangerw (approved):
Looks good! Just a few comments.

abheesht17:
Ready for merge!

mattdangerw:
All looks good to me! Note that we have not actually stuck BART in a release yet, but hopefully we can ready one with gpt-neox, pythia, bart and whisper in the next week or two.

I am fine if we just release the guide based on the master branch for now. We could also wait for the release to better promote on social.

Maybe easiest to release this as it's ready, and tweet about it again when the next KerasNLP release is out.

fchollet (approved):
Thank you for the contribution!

PR #1424, opened 2023-07-10
Update guides for Keras V3 Saving Compatibility

nkovela1:
This PR updates existing keras-io guides ensuring that the models created are saveable via the new Keras V3 saving format (.keras).

Most changes involve the addition of the `@keras.utils.register_keras_serializable()` decorator, which should be standard practice for custom objects. Additionally, these guides were updated to use the new standard `import keras` instead of `from tensorflow import keras`.


fchollet (commented):
Nice! Please also regenerate the corresponding md/ipynb files via the add_guide command.

nkovela1:
Regenerated all the md/ipynb files corresponding with the changes, thanks!

fchollet (commented):
Thanks for the PR!

Please revert the changes to the images -- the image source is shared with the keras core guides and we should use their version to stay up to date.

nkovela1:
Reverted all image changes, thanks!

MarkDaoust (approved):
Francois' comments are all resolved and this looks good to me.

PR #1431, opened 2023-07-12
Updated the broken links in TF doc

sushreebarsa:
Changed the broken links with correct links for the following texts in making_new_layers_and_models_via_subclassing.py file in documentation. 
1.guide to writing a training loop from scratch, 
2. Functional model 
3.guide to saving and serializing models
4.understanding padding and masking

Thank you!

### DIFF
--- a/guides/making_new_layers_and_models_via_subclassing.py
+++ b/guides/making_new_layers_and_models_via_subclassing.py
@@ -294,7 +294,8 @@ for x_batch_train, y_batch_train in train_dataset:
 
 """
 For a detailed guide about writing training loops, see the
-[guide to writing a training loop from scratch](/guides/writing_a_training_loop_from_scratch/).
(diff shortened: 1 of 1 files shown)


MarkDaoust (changes requested):
Thanks for pointing this out. 

The files in the first level of /guides/ are mirrored to tensorflow.org.

This will be best is I just fix the import tool, since there are similar errors in a bunch of other files. 

We should be consistent here.  Either all links should or the `(/guides/{name}/)` form (no change), or you could update them all to be keras.io URLs, so they also work in colab.

I'll fix tensorflow.org

MarkDaoust:
cl/549120156

MarkDaoust:
The links on tensorflow.org are fixed, let's close this.

sushreebarsa:
@MarkDaoust Thank you for the update!

PR #1435, opened 2023-07-12
Document `AdamW` and `Adafactor` using the `keras.optimizers` namespa…

hertschuh:
…ce instead of the experimental namespace.


fchollet (approved):
LGTM, thank you!

PR #1436, opened 2023-07-12
Add Keras Core guides for distributed training with JAX and PyTorch.

hertschuh:


hertschuh:
> Please add the generated files via python autogen.py add_guide keras_core/distributed... -- should be run on a 2-GPU VM so that we get the right logs.

Done

fchollet (approved):
Thank you!

PR #1439, opened 2023-07-14
Add sigmoid activation to transfer_learning.py

### LINKED ISSUE #463: Please add basic prediction example to transfer learning guide
Unless I'm missing something basic, I don't see any example of actually using the transfer learning guide to make a prediction

https://github.com/keras-team/keras-io/blob/master/guides/transfer_learning.py
https://keras.io/guides/transfer_learning/

For example, if I run all the code in the "end-to-end" example at the end, my naive attempt at making some form of prediction is the following:

```
for obj, _ in train_ds:
    print(model.predict(obj))
    break

[[ 5.7956142]
 [ 7.138669 ]
 [ 6.004242 ]
 [-7.5267677]
 [ 4.4232373]
 [ 4.33535  ]
 [-6.3878946]
 [-5.303356 ]
 [ 4.605398 ]
 [ 8.426602 ]
 [10.56328  ]
 [-7.899274 ]
 [-9.870649 ]
 [ 3.5054545]
 [-9.1837   ]
 [-8.082974 ]
 [-9.05807  ]
 [ 7.5205717]
 [-6.4828777]
 [ 8.651134 ]
 [ 7.1191254]
 [-5.9992075]
 [ 5.146449 ]
 [ 4.319687 ]
 [-8.059815 ]
 [-8.574115 ]
 [ 9.458901 ]
 [ 6.465415 ]
 [ 6.0897965]
 [10.061241 ]
 [10.10572  ]
 [ 3.1187348]]
```

I'm quite confused as to what these values represent. I believe the `base_model` had more classes than just cat or dog and so maybe it's giving me all the classes. That's just speculation anyway. I think there should be a final example that shows an actual prediction resulting in a value "cat" or "dog" to complete this example. To me this seems like maybe the single most important piece of everything.

Anyway I'll probably sort this out myself, but if someone reading this can explain how to make predictions here I'd be grateful (and I could even take it upon myself to expand the `transfer_learning.py` guide myself).

SuryanarayanaY:
The `transfer_learning.py` tutorial has been used cat vs dog dataset which is a `binary classification` problem. But the model output dont have `sigmoid` activation and hence leading to outputs of out of the range `[0,1]` that also have negative values  which is difficult to interpret. Hence added sigmoid activation to output layer.

Also for binary classification problem `from_logits=False` makes more sense for loss function but currently it is `from_logits=True`.

Hence done both the changes and now model predicts outputs within the range of `[0,1]`.

Please refer the attached [gist](https://colab.research.google.com/gist/SuryanarayanaY/8a57cabc653aa74a127ad44a6cffd4c9/kerasio_463-transfer_learning.ipynb#scrollTo=kpD8KPNyTG7V).

Fixes #463

### DIFF
--- a/guides/transfer_learning.py
+++ b/guides/transfer_learning.py
@@ -235,7 +235,7 @@ x = base_model(inputs, training=False)
 # Convert features of shape `base_model.output_shape[1:]` to vectors
 x = keras.layers.GlobalAveragePooling2D()(x)
 # A Dense classifier with a single unit (binary classification)
-outputs = keras.layers.Dense(1)(x)
+outputs = keras.layers.Dense(1,activation='sigmoid')(x)
 model = keras.Model(inputs, outputs)
 ```
 
@@ -243,7 +243,7 @@ Train the model on new data.
 
 ```python
 model.compile(optimizer=keras.optimizers.Adam(),
-              loss=keras.losses.BinaryCrossentropy(from_logits=True),
+              loss=keras.losses.BinaryCrossentropy(from_logits=False),
               metrics=[keras.metrics.BinaryAccuracy()])
 model.fit(new_dataset, epochs=20, callbacks=..., validation_data=...)
 ```
@@ -281,7 +281,7 @@ base_model.trainable = True
 # to the `trainable` attribute of any inner layer, so that your changes
 # are take into account
 model.compile(optimizer=keras.optimizers.Adam(1e-5),  # Very low learning rate
-              loss=keras.losses.BinaryCrossentropy(from_logits=True),
+              loss=keras.losses.BinaryCrossentropy(from_logits=False),
               metrics=[keras.metrics.BinaryAccuracy()])
 
 # Train end-to-end. Be careful to stop before you overfit!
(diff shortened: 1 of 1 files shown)


MarkDaoust:
Hi, @SuryanarayanaY thanks for tanking the time to make a PR.

The interpretation of the outputs is negative values are class zero, positive values are class one. you can always run it through a sigmoid yourself.

There are good arguments for both ways of doing it. 

There's nothing wrong with using `activation=None + from_logits=True`. This approach avoids some numerical problems that can break `activation="sigmoid" + from_logits=False`. Keras usually handles that transparently.  If you're using eager execution keras can't apply that fix. In some cases it will have almost no effect (since keras does use an `eps` on the loss calculation so you won't get NaNs), but in some cases (if you have a large number of classes) the effect could be important. 

Anyway, there's no clear standard here, in this repository:

- [from_logits=True](https://github.com/search?q=repo%3Akeras-team%2Fkeras-io++from_logits%3DTrue&type=code) has ~170 uses.
- [activation='sigmoid'](https://github.com/search?q=repo%3Akeras-team%2Fkeras-io++activation%3D%22sigmoid%22&type=code) or [activation='softmax'](https://github.com/search?q=repo%3Akeras-team%2Fkeras-io++activation%3D%22softmax%22&type=code) has ~210.

So I'm -1 on changing just one for personal preferences.

PR #1449, opened 2023-07-21
Fixed the broken link in understanding_masking_and_padding.ipynb

tilakrayal:
Fixed the broken link of  **serialization** in `understanding_masking_and_padding.ipynb`


nkovela1 (approved):

MarkDaoust:
Hi, thanks for the fix, yes, the `/guides` link only works on keras.io, full URLs will work in keras.io, GitHub, and on Colab. 

But the notebooks here are generated from the PY files in guides/. If you don't update the PY file, this will get overwritten next time someone regenerates the notebook.

Issue #1328, opened 2023-04-09
Typo in the Learnable Resizer example


IaBogdi:
Dear Sir or Madam,

there is a typo in the code of [Learnable Resizer](https://github.com/keras-team/keras-io/blob/master/examples/vision/learnable_resizer.py).  In the lines 155-156, it seems like residual blocks take input from bottleneck what contradicts with the architecture of original [paper](https://arxiv.org/pdf/2103.09950v1.pdf), where the residual blocks are connected in series. It works fine, when the number of residual blocks is 1, which is the case of this example ;)

I would change the lines 155-156  as follows:

```python
    x = res_block(bottleneck)
    # Residual passes.
    for _ in range(num_res_blocks - 1):
        x = res_block(x)
```

SuryanarayanaY:
Hi @IaBogdi ,

Thanks for reporting the issue. I have gone through the source code and found that the way it is implemented makes first `res_block` to take inputs from bottleneck and subsequent outputs of resblocks again passed to bottleneck layer before feeding to next `res_block`. IMO, this is definitely contradicting to the architecture of standard `Resizer model`. 

In keras_io also the same picture used from that of paper which indicates there is no bottleneck layer in between `res_blocks`.
<img width="792" alt="Screenshot 2023-05-31 at 6 47 31 PM" src="https://github.com/keras-team/keras-io/assets/116063290/ec30e9c7-baef-4bf1-bf89-fbb9208176ea">


IMO as per the standard architecture the implementation seems wrong for me and your proposed code change will fit the standard architecture.However we need confirmation from Developer team whether they want to implement same architecture or they want to really add `bottleneck layer` after each `res_block`. So let's hear confirmation from Dev team.

Thanks!

SuryanarayanaY:
@sayakpaul , Could you please have a look into this and comment your views. Thanks!

PR #1455, opened 2023-07-27
Changes to support multiple residual blocks for learnable_resizer

SuryanarayanaY:
In the current implementation of learnable_resizer author used no of residual blocks=1 and accordingly code was implemented.

Consider the code below at line no.155

for _ in range(num_res_blocks):
        x = res_block(bottleneck)
Author used num_res_blocks=1 for this example but he intends to extend support to num_res_blocks>1 also hence considered for loop otherwise the above code can be replaced by just x = res_block(bottleneck).

However as per the Architecture the each res_block should get input from its previous res_block. But with current implementation each res_block is getting bottle_neck layer output as input which breaks the implementation if num_res_blocks>1.

Hence I proposed the changes and added notes so that user can able to change the num_res_blocks to >1 also and able to get correct output as expected which lacks in current implementation.

The issue was highlighted in issue https://github.com/keras-team/keras-io/issues/1328 by the issue author.

The proposed changes will make learnable_resizer more generalized.

Thanks!

### DIFF
--- a/examples/vision/learnable_resizer.py
+++ b/examples/vision/learnable_resizer.py
@@ -132,7 +132,7 @@ def res_block(x):
     x = conv_block(x, 16, 3, 1, activation=None)
     return layers.Add()([inputs, x])
 
-
+#Note: user can change num_res_blocks to >1 also if needed
 def get_learnable_resizer(filters=16, num_res_blocks=1, interpolation=INTERPOLATION):
     inputs = layers.Input(shape=[None, None, 3])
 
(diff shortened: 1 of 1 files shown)


MarkDaoust:
This Looks good to me, can you generate updates doe the markdown and notebook files too please?

SuryanarayanaY:
Sure. Actually I am awaiting approval for proposed changes so that  I can go ahead to do changes in .md and .ipynb too.

SuryanarayanaY:
Updated `.md` and `.ipynb `files also.

MarkDaoust:
Thanks!

PR #1456, opened 2023-07-28
Update KerasNLP getting started guide for multi-backend keras

mattdangerw:


mattdangerw:
This is blocked on https://github.com/keras-team/keras-nlp/pull/1185, but should be ready for review.

guides/keras_nlp/getting_started.py
```diff
@@ -486,8 +500,8 @@ def preprocess(inputs, label):
 pretraining_model.summary()
 pretraining_model.compile(
     loss=keras.losses.SparseCategoricalCrossentropy(from_logits=True),
-    optimizer=keras.optimizers.experimental.AdamW(learning_rate=5e-4),
-    weighted_metrics=keras.metrics.SparseCategoricalAccuracy(),
+    optimizer=keras.optimizers.AdamW(learning_rate=5e-4),
```
ianstenbit, line 503: Isn't this just `Adam` now with the `weight_decay` parameter set?

ianstenbit (approved):
Maybe update PR title to mention KerasNLP?

guides/keras_nlp/getting_started.py
mattdangerw: Interesting! I didn't know this. But it is still exported as a separate name.

Since `AdamW` will work both in `tf.keras` and `keras-core`, that's probably the safe way to leave this for now.

guides/keras_nlp/getting_started.py
```diff
@@ -231,17 +236,24 @@
 In this workflow we train the model over three epochs using `tf.data.Dataset.cache()`,
 which computes the preprocessing once and caches the result before fitting begins.
 
-**Note:** this code only works if your data fits in memory. If not, pass a `filename` to
-`cache()`.
+**Note:** that we can use `tf.data` for preprocessing while running on the
+Jax or PyTorch backend. The input dataset will automatically be converted to
+backend native tensor types during fit. In fact, given the efficiency of `tf.data`
+for running preprocessing, this is good practice, on all backends.
```
jbischof, line 242: no comma after practice

guides/keras_nlp/getting_started.py
jbischof: IIUC AdamW is *decoupled* weight decay. There's a bunch of sthick in the [paper](https://arxiv.org/pdf/1711.05101.pdf) about how you have to be very careful adding weight decay in Adam.

jbischof (approved):
Thanks!

Making JAX the default is an interesting and maybe correct choice. Anyway it's easy to change.

guides/keras_nlp/getting_started.py
mattdangerw: I think what @ianstenbit is pointing out is that https://github.com/keras-team/keras-core/blob/main/keras_core/optimizers/adamw.py is literally just an empty subclass. Not sure if that reflects a bug in `AdamW` in keras-core, or if `keras_core.optimizers.Adam` is correctly handling the decoupled weight decay.

PR #1461, opened 2023-08-02
Update custom_object documentation and __init__() arg serialization_and_saving.py

SuryanarayanaY:
In the example of a `CustomLayer` class its `__init__` method passed argument `sublayer` whereas while assigning to self it is using the variable `layer` instead of `sublayer`. Hence corrected the same.

Also please check the documentation below.

> If the arguments passed to the constructor (`__init__()` method) of the custom object aren't Python objects (anything other than base types like ints, strings, etc.), then you **must** also explicitly deserialize these arguments in the `from_config()` class method.

Actually we need to also add the statement like you must explicitly serialize the `non_python arguments` in `get_config()` method also so that it can be deserialized in `from_config()` method. As it is explicitly mentioned for `from_config()` it should be explicitly mentioned for `get_config()` also which is missing now.

This makes users well aware of this. Because without serializing `non_python` object in `get_config()` , it is not possible to reload that back into model.

### DIFF
--- a/guides/serialization_and_saving.py
+++ b/guides/serialization_and_saving.py
@@ -133,8 +133,8 @@ When saving a model that includes custom objects, such as a subclassed Layer,
 you **must** define a `get_config()` method on the object class.
 If the arguments passed to the constructor (`__init__()` method) of the custom object
 aren't Python objects (anything other than base types like ints, strings,
-etc.), then you **must** also explicitly deserialize these arguments in the `from_config()`
-class method.
+etc.), then you **must** serialize these arguments in `get_config()` method and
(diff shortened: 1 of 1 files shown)


MarkDaoust:
I think this is right. @nkovela1 can you confirm?

nkovela1 (approved):
This is great! Thank you for fixing the typo as well! Approved.

fchollet (approved):
LGTM -- @nkovela1 can you regenerate the guide and send a PR for the updated files?

nkovela1:
@fchollet Sure! Here is the PR: https://github.com/keras-team/keras-io/pull/1465

PR #1467, opened 2023-08-05
"FeatureSpace advanced use cases" Keras code example - Fixing typos and code formatting

dimitreOliveira:
Hi, I am the author of the [FeatureSpace advanced use cases](https://keras.io/examples/structured_data/feature_space_advanced/) Keras code example, I am creating this PR to fix a typo and fix some other minor code formatting, that is the reason I am uploading all 3 files here.

### DIFF
--- a/examples/structured_data/md/feature_space_advanced.md
+++ b/examples/structured_data/md/feature_space_advanced.md
@@ -1,8 +1,8 @@
-# FeatureSpace avanced use cases
+# FeatureSpace advanced use cases
 
 **Author:** [Dimitre Oliveira](https://www.linkedin.com/in/dimitre-oliveira-7a1a0113a/)<br>
 **Date created:** 2023/07/01<br>
-**Last modified:** 2023/07/01<br>
+**Last modified:** 2023/08/04<br>
 **Description:** How to use FeatureSpace for advanced preprocessing use cases.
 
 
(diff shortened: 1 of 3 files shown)


examples/structured_data/md/feature_space_advanced.md
```diff
@@ -125,7 +118,7 @@ target label), here's a preview of a few samples:
 
 ```python
```
MarkDaoust, line 119: There's something still wrong with the markdown output in this section. 

It might have to do with the fact that this cell both prints some text and displays the DataFrame.

MarkDaoust (changes requested):
Hi, thanks for the fixes!

Something's not rendering correctly in the markdown file, can you take a look?

examples/structured_data/md/feature_space_advanced.md
dimitreOliveira: Hey thanks for the review.

It is weird that my code is not rendering properly here, [this other code example here has the same exact two lines](https://github.com/keras-team/keras-io/blob/474cbcbdb2e9310a7605c08fade0b15a2ee8a632/examples/structured_data/wide_deep_cross_networks.py#L98) and there it seems to work. I am having some trouble starting the local website at my machine (Macos) with 
```
python autogen.py make
python autogen.py serve
```

but I will try to fix it.

PR #1470, opened 2023-08-06
Introducing Quit Flag for Unzipping Process

hamidriasat:
Hi,

In the download data section, While unzipping, It was printing a lot of file names on a terminal window which was affecting the readability of example. So I added a quit flag during unzipping to stop printing file names which will result in improved readability.

Thanks


fchollet (approved):
Thank you for the fix! LGTM

PR #1473, opened 2023-08-08
Port sentence embedding example to keras-core

abuelnasr0:
Port sentence embedding example to keras-core. keras-team/keras-nlp#1190
I will generate other files and add them once these changes get accepted. 
@mattdangerw

### DIFF
--- a/examples/nlp/sentence_embeddings_with_sbert.py
+++ b/examples/nlp/sentence_embeddings_with_sbert.py
@@ -1,17 +1,18 @@
 """
-Title: Sentence embeddings using Siamese RoBERTa-networks
+Title: Sentence embeddings using Siamese RoBERTa-networks with kerasNLP
 Author: [Mohammed Abu El-Nasr](https://github.com/abuelnasr0)
 Date created: 2023/07/14
-Last modified: 2023/07/14
-Description: Fine-tune a RoBERTa model to generate sentence embeddings using KerasNLP.
+Last modified: 2023/08/08
+Description: Use kerasNLP to Fine-tune a RoBERTa model to generate sentence embeddings for
+semantic similarity and clustering tasks.
 Accelerator: GPU
 """
 
 """
 ## Introduction
 
-BERT and RoBERTa can be used for semantic textual similarity tasks, where two sentences
-are passed to the model and the network predicts whether they are similar or not. But
+BERT or RoBERTa can be used for semantic textual similarity tasks, where two sentences
+are passed to the model, and the network predicts whether they are similar or not. But
 what if we have a large collection of sentences and want to find the most similar pairs
 in that collection? That will take n*(n-1)/2 inference computations, where n is the
 number of sentences in the collection. For example, if n = 10000, the required time will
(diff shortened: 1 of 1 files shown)


MarkDaoust:
Hi, thanks for the PR. We really appreciate it when authors maintain their examples.

I have one question about this change: 

It adds +175 lines, but at first glance it feels like most of the added lines are packing and repacking dictionaries that weren't necessary before.

Is the upgrade to keras-core forcing you to do this, or is it purely driven by the decision to move the text-preprocessing into the `tf.data.Dataset`? (I assume keras tensor-masks don't make it through). I know that can be good from an accelerator efficiency POV, but these examples should  be short and readable.

abuelnasr0:
@MarkDaoust Torch and jax doesn't have string dtype. so I was forced to move text-preprocessing to tf.data.Dataset. I will try to make it shorter.

examples/nlp/sentence_embeddings_with_sbert.py
```diff
@@ -101,35 +159,61 @@
...
 
 
 def prepare_dataset(dataset, num_batchs, batch_size):
+    dataset = dataset.batch(batch_size)
+    dataset = dataset.take(num_batchs)
     dataset = dataset.map(
-        lambda z: (
-            [z["sentence1"], z["sentence2"]],
-            [tf.cast(change_range(z["label"]), tf.float32)],
+        lambda x: (
```
MarkDaoust, line 209: Please just write this out as a stand alone function, it will be clearer.

examples/nlp/sentence_embeddings_with_sbert.py
```diff
@@ -101,35 +159,61 @@
...
+preprocessor = keras_nlp.models.RobertaPreprocessor.from_preset("roberta_base_en")
 
-AUTOTUNE = tf.data.experimental.AUTOTUNE
+
+def one_dictionary(x):
+    """
+    Returns the data as one dictionary to be appropriate input for
+    `keras_core.Model`
+    """
+    return {
```
MarkDaoust, line 193: Does keras_core support nested dictionaries?

Shouldn't this work:

{'input0': x[0], 'input1':x[1]}?

examples/nlp/sentence_embeddings_with_sbert.py
```diff
@@ -75,12 +82,63 @@
...
 best results. We will use it in our examples.
+
+We will use two methods for fine-tuning:
+
+- Fine-tune using the regression objective function
+- Fine-tune using the triplet Objective function
 """
 
+"""
+### Build the encoder model.
```
MarkDaoust, line 97: I think this section made more sense where it was, in logical order:
- Convert text to tokens.
- Encode the tokens.

MarkDaoust (changes requested):
That's too bad.  One of the things I love about the old approach is that you can export the model with all it's preprocessing, so that you can just throw strings at it and don't need to worry about separately exporting and running the string processing. I wonder if it would make sense to fork so people can see both versions.

> try to make it shorter.

LMK if I'm wrong, but I think you can save a bit in all the spots where you're flattening dictionaries. keras supports nested dicts, does keras_core?

abuelnasr0:
@MarkDaoust sorry for late reply. I think forking will be great. I have seen that in keras_core library there is an example folder where a lot of TF examples are being ported into keras_core. so I can add this example there ? or should I create new version of it here?

>  keras supports nested dicts, does keras_core?

keras_core doesn't support that unfortunately.
I think adding the example in keras_core/examples would make sense more. if it is better to add it here I will create a new file with keras_core example.

PR #1478, opened 2023-08-11
fixed typos and broken cells in KerasCV object detection guide

pcoet:
I had to change the code in two places to make the cells run. One involved an API that expected a numpy array rather than a list (fixed). The other was this:

```
- result = pretrained_model.evaluate(eval_ds.take(40), verbose=0)
+ result = pretrained_model.evaluate(eval_ds.take(1), verbose=0)
```

If you try to `take` 40 examples, you get the following error:

```
tensorflow.python.framework.errors_impl.InvalidArgumentError: {{function_node __wrapped__ConcatV2_N_20_device_/job:localhost/replica:0/task:0/device:CPU:0}} ConcatOp : Dimension 1 in both shapes must be equal: shape[0] = [4,2,4] vs. shape[1] = [4,4,4] [Op:ConcatV2] name: concat
```

I can make the code run successfully by taking only 1 example, but that does alter the intention of the cell. If there's a better solution, please let me know.


guides/keras_cv/object_detection_keras_cv.py
```diff
@@ -379,7 +380,7 @@ class mapping now.
 )
 ```
 
-Clearly yields bounding boxes in the format `xywh`.  You can read more about
+clearly yields bounding boxes in the format `xywh`.  You can read more about
```
MarkDaoust, line 383:
```suggestion
This clearly yields bounding boxes in the format `xywh`.  You can read more about
```

MarkDaoust (approved):

PR #1479, opened 2023-08-11
Typos were fixed in TF doc

sushreebarsa:
Minimal typos were fixed. Thank you!


MarkDaoust (changes requested):
Yes, thanks! 

But the `.py` file is the source of truth. Please update it and run `cd scripts; python autogen.py add_guide guides/transfer_learning`.

PR #1484, opened 2023-08-14
Added missing .self to unit_2 in "Custom cell that supports nested input/output"

pckSF:
A quick fix for a "typo" that I found in: [Define a custom cell that supports nested input/output](https://www.tensorflow.org/guide/keras/working_with_rnns#define_a_custom_cell_that_supports_nested_inputoutput).

I think there is a `self.` missing for `unit_2` in `return {"unit_1": self.unit_1, "unit_2": unit_2, "unit_3": self.unit_3}`

### DIFF
--- a/guides/ipynb/working_with_rnns.ipynb
+++ b/guides/ipynb/working_with_rnns.ipynb
@@ -759,7 +759,7 @@
     "        return output, new_states\n",
     "\n",
     "    def get_config(self):\n",
-    "        return {\"unit_1\": self.unit_1, \"unit_2\": unit_2, \"unit_3\": self.unit_3}\n",
+    "        return {\"unit_1\": self.unit_1, \"unit_2\": self.unit_2, \"unit_3\": self.unit_3}\n",
     ""
    ]
   },
--- a/guides/ipynb/working_with_rnns.ipynb
+++ b/guides/ipynb/working_with_rnns.ipynb
@@ -717,7 +717,7 @@
    "outputs": [],
    "source": [
     "\n",
-    "@keras.saving.register_keras_serializable()\n",
+    "@keras.utils.register_keras_serializable()\n",
     "class NestedCell(keras.layers.Layer):\n",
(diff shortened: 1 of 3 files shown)


pcoet:
Thanks for the PR, @pckSF. LGTM.

pcoet (approved):

pcoet:
Note: In order to run the updated cell, I had to change `@keras.saving.register_keras_serializable()` to `@keras.utils.register_keras_serializable()`.

pckSF:
Ah interesting, does that mean that [this](https://www.tensorflow.org/api_docs/python/tf/keras/saving/register_keras_serializable) also needs to be updated to match the [Keras API reference](https://keras.io/api/saving/serialization_utils/)? Or is it the other way around since there is a

```python
@keras_export(
    "keras.utils.serialize_keras_object",
    "keras.utils.legacy.serialize_keras_object",
)
```
in [.../keras/saving/legacy/serialization.py](https://github.com/keras-team/keras/blob/v2.12.0/keras/saving/legacy/serialization.py#L280) and 

```python
@keras_export(
    "keras.saving.register_keras_serializable",
    "keras.utils.register_keras_serializable",
)
```
in [.../keras/saving/object_registration.py](https://github.com/keras-team/keras/blob/v2.13.1/keras/saving/object_registration.py#L98-L158).

pcoet:
I'll have to look into this and see what docs need to be updated. Thanks for the fix!

haifeng-jin:
@nkovela1 Would you please take a look at this PR? Thanks!

nkovela1:
Hi @pckSF , thanks for the PR! The `@keras.saving.register_keras_serializable` is the correct namespace so it should work, would it be possible to only make changes for the typo (adding "self.")?

This namespace is new to the TF 2.13 release, I believe, so if you're running versions prior it likely won't be recognized.

pckSF:
Yes, I've changed it back so that it now only adds the `self.`.

nkovela1:
Great, thank you @pckSF! LGTM, @MarkDaoust please approve and merge, thanks!

Issue #857, opened 2022-04-19
Move `KerasTuner` tab higher up


LukeWood:
The page linked here is great!  But it is very low down - below even Contributing to Keras!  We should move this, KerasCV, and KerasNLP higher.

<img width="896" alt="Screen Shot 2022-04-19 at 12 48 11 PM" src="https://user-images.githubusercontent.com/12191303/164083904-0d7bf9c6-bb1c-4137-a27d-27d746b929eb.png">

innat:
@LukeWood I also think that `KerasTuner - KerasNLP - KerasCV` should be enough higher. Currently, it somewhat looks less important or not highly recommended.

PR #1486, opened 2023-08-14
Move KerasTuner, KerasCV, KerasNLP higher

sachinprasadhs:
Moved KerasTuner, KerasCV, KerasNLP to top for more visibility.
Closes: https://github.com/keras-team/keras-io/issues/857
image


scripts/master.py
```diff
@@ -529,18 +529,6 @@
...
-        },
-        {
-            "path": "governance",
-            "title": "Community & governance",
-        },
-        {
-            "path": "contributing",
-            "title": "Contributing to Keras",
-        },
         {
```
mattdangerw, line 532: I might find things more scannable if we grouped things better. And potentially added a caption for all our libraries (similar to what we have for KerasCore?).

About
Getting Started
Guides
Examples
API Reference
Keras Core: Keras for TensorFlow, JAX, and PyTorch
Keras Tuner: Hyperparameter Tuning
Keras CV: Computer Vision Workflows
Keras NLP: Natural Language Workflows

Then we are keeping "like next to like." E.g. guides next to examples. Ecosystem libraries next to each other. @fchollet what do you think?

mattdangerw: Oh also, let's rename the `Keras API Reference` section to just `API Reference`. It's implied that everything is Keras, and we also list the domain package APIs there.

sachinprasadhs: Thanks for reviewing, I made the changes as you suggested, please review the below screenshot before I make commit.

image

fchollet (approved):
This looks like a good idea. Thanks for the PR.

PR #1487, opened 2023-08-14
Update KerasCV OD guide to use Keras Core

ianstenbit:
Original author: Tirth Patel
(When I autogen the MD and notebook files I will attribute that commit to him so he gets authorship on this PR)


guides/keras_cv/object_detection_keras_cv.py
```diff
@@ -17,15 +17,27 @@
...
 """shell
-!pip install --upgrade git+https://github.com/keras-team/keras-cv
+!pip install -U keras-core git+https://github.com/keras-team/keras-cv.git
 """
-import tensorflow as tf
+import os
+
+os.environ["KERAS_BACKEND"] = "tensorflow"
+
+from tensorflow import data as tf_data
```
jbischof, line 33: Why not `tf.data`?

guides/keras_cv/object_detection_keras_cv.py
```diff
@@ -586,139 +607,60 @@ def dict_to_tuple(inputs):
...
+easy-to-use suite of COCO metrics under the `keras_cv.callbacks.PyCOCOCallback`
+symbol. Note that we use a Keras callback instead of a Keras metric to compute
+COCO metrics. This is because computing COCO metrics requires storing all of a
+model's predictions for the entire evaluation dataset in memory at once, which
+is impractical to do during training time.
 """
 
-coco_metrics = keras_cv.metrics.BoxCOCOMetrics(
-    bounding_box_format="xywh", evaluate_freq=20
+coco_metrics_callback = keras_cv.callbacks.PyCOCOCallback(
```
jbischof, line 637: It looks like we're no longer evaluating the pretrained model using the COCO metrics. The original goal here was to show the user some realistic values for these without having to train the model e2e. Is this something we could add back?

guides/keras_cv/object_detection_keras_cv.py
```diff
@@ -586,139 +607,60 @@ def dict_to_tuple(inputs):
...
-)
-coco_metrics.reset_state()
-result = pretrained_model.evaluate(eval_ds.take(1), verbose=0)
-result = coco_metrics.result(force=True)
-
-print_metrics(result)
-
-"""
-
-**A note on TPU compatibility:**
```
jbischof, line 660: Is the TPU issue no longer relevant?

jbischof (commented):
Look great in general but we're also losing some content and would like to understand why.

guides/keras_cv/object_detection_keras_cv.py
ianstenbit: The idea was to demonstrate that TF isn't used except for tf data, but I switched it back because this is a departure from normal convention

guides/keras_cv/object_detection_keras_cv.py
ianstenbit: We are still doing this, because when we show how to train the model we're using the PyCOCOCallback during training and we're training an already-pretrained model, so we can see the numerical results from the callback in that block.

IMO this whole block was very weird because it added a bunch of code that doesn't exist in KerasCV which really distracts from our API

ianstenbit: Whoops -- this is actually not true. I forgot that I had switched to using a pre-trained backbone without a pre-trained model head.

If you'd like I can add a block about how to just evaluate the metrics, or I can just add a text blurb about expected values. Wdyt @jbischof

jbischof: Yes, it should be relatively low friction to add some metrics to the completely pretrained model, I hope!

guides/keras_cv/object_detection_keras_cv.py
jbischof: Very cool!

ianstenbit:
@jbischof this is ready for another look

guides/keras_cv/object_detection_keras_cv.py
ianstenbit: yep -- added.

jbischof (approved):
Thanks for the update!

ianstenbit:
I used the autogen script to update the rendered versions.

@fchollet please take a look when you have a chance -- thank you!

sampathweb:
Closing this in favor of #1662 that updates guide to Keras 3

PR #1493, opened 2023-08-17
Align guides with docstring of Layer.add_loss():

arnoegw:
Use the mean (not the sum) of input-dependent losses, matching PR https://github.com/keras-team/keras-io/pull/1144.

### DIFF
--- a/guides/intro_to_keras_for_researchers.py
+++ b/guides/intro_to_keras_for_researchers.py
@@ -411,14 +411,14 @@ Here's a layer that creates an activity regularization loss:
 class ActivityRegularization(keras.layers.Layer):
     """Layer that creates an activity sparsity regularization loss."""
 
-    def __init__(self, rate=1e-2):
+    def __init__(self, rate=0.1):
         super().__init__()
         self.rate = rate
 
     def call(self, inputs):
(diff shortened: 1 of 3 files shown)


arnoegw:
Friendly ping... Should be easy, it's a 5-line change that just aligns example code with the docstring at https://github.com/keras-team/keras/blob/r2.14/keras/engine/base_layer.py#L1450

grasskin (changes requested):
Thank you for the ping! The reduce_mean changes LGTM. Is there a reason for the switch to 0.1 as our learning rate? I couldn't find a mention in the docstring and I believe we usually use 1e-2 (https://keras.io/api/optimizers/)

arnoegw:
It's not the learning rate for the optimizer; it's the coefficient of this particular loss term. Growing it by an order of magnitude is meant to balance the shrinking by an order of magnitude in the loss term itself (going from a sum over some tens of examples to a mean).

grasskin (approved):
LGTM!

arnoegw:
@fchollet @MarkDaoust , I believe this needs approval from one of you as code owners.

arnoegw:
Thank you, François and Gabriel!

Issue #1492, opened 2023-08-17
Duplicate Image Search example with broken link


under-score:
The example [https://github.com/keras-team/keras-io/blob/master/examples/vision/near_dup_search.py](https://github.com/keras-team/keras-io/blob/master/examples/vision/near_dup_search.py)
does not work due to a broken link to the dataset
`!wget -q https://git.io/JuMq0 -O flower_model_bit_0.96875.zip`
Fortunately the model can be recreated here [https://github.com/sayakpaul/near-dup-parser/blob/main/bit-supervised-training.ipynb](https://github.com/sayakpaul/near-dup-parser/blob/main/bit-supervised-training.ipynb)

sachinprasadhs:
Hi,

Thanks for reporting the issue.

I was able to make some changes like loading the pre-trained model from TF-Hub  and run the notebook successfully.

Please refer the below Colab till we make the changes in the example document.

https://gist.github.com/sachinprasadhs/d1672b5badfa9df9b07786ab2dfa3cc0

under-score:
million thanks

sachinprasadhs:
Hi, 

The PR has been merged with only change to the pre-trained model link https://github.com/keras-team/keras-io/pull/1508.
You can refer the PR and close the issue. Thanks!

PR #1499, opened 2023-08-18
Change dataset link with TFHub

sachinprasadhs:
Change dataset link with TFHub and below changes.
* Removed previously generated .png files.
 * Removed unwanted cell outputs
 * Made some minor code changes
 
Closes: https://github.com/keras-team/keras-io/issues/1492
Here is the actual Gist for reference: https://gist.github.com/sachinprasadhs/d1672b5badfa9df9b07786ab2dfa3cc0
cc: @sayakpaul


sayakpaul:
I don't think you should be changing the model as the model was specifically suited for flowers. I will provide the exact model link to you that you can just update in the markdown, py, and ipynb files.

sayakpaul:
https://github.com/sayakpaul/near-dup-parser/releases/download/v0.1.0/flower_model_bit_0.96875.zip here you go.

sachinprasadhs:
Created a new PR with only changes to model link and tensorrt setup here https://github.com/keras-team/keras-io/pull/1508

PR #1075, opened 2022-09-20
Display test image

suvigyajain0101:
Can add a simple line at the end to show what the image looks like along with it's predictions

PR #1500, opened 2023-08-19
Display test image

suvigyajain0101:
Adding a line to display the test image used for prediction. The reason is to provide a better understanding of how the predictions are linked with image being tested.

Original PRs - #1091 , #1075


pcoet (approved):
LGTM. The image won't appear in the HTML page until we regenerate the markdown file, but I think that's fine, in this case.

suvigyajain0101:
@pcoet Sounds good. Thanks a lot!

PR #1506, opened 2023-08-30
Remove unwanted loss function in depth_estimation.py

### LINKED ISSUE #969: Confused about loss in depth estimation example
I'm following this example:
https://github.com/keras-team/keras-io/blob/master/examples/vision/depth_estimation.py

However, I'm confused as to why the model is compiled with the cross-entropy loss when a custom loss function is being used for optimization?

SuryanarayanaY:
The **depth_estimation.py** example implements `custom_loss` function (i.e. calculate_loss) using 1. Structural similarity index(SSIM),  2. L1-loss, or Point-wise depth in our case and 3. Depth smoothness loss. The same loss function used in `train_step` also. But still `model.compile()` was given another loss function  `cross_entropy = tf.keras.losses.SparseCategoricalCrossentropy(from_logits=True, reduction="none")` which has to be removed as it has no effect and also creates unwanted confusion.

Hence deleted the unwanted loss defined in `model.compile()`

Fixes #969


pcoet (approved):

pcoet:
@SuryanarayanaY Do you know how to resolve the docker-image failure? Then we can merge this...

PR #1516, opened 2023-09-08
fix encoding

fsx950223:
Fix 'utf-8' codec can't decode byte 0xe9 in position 3114: invalid continuation byte


divyashreepathihalli:
@shivalikasingh95 sorry this Pr has slipped through the cracks. Since Keras has changed from tf backend to multi-backend. the PR changes would no longer be up to date. I will close this out. please feel free to open a new one. Thank you!

PR #1534, opened 2023-09-14
Add DeepLab example

aaudevart:
Add a new example using DeepLabV3+ pretrained model for semantic segmentation in healthcare

Google colab : https://colab.research.google.com/drive/17UQr7-UM_Et0A7g7638Gq7Vb-e1171kC?usp=sharing

### DIFF
--- a/examples/vision/semantic_segmentation_deeplab_v3.py
+++ b/examples/vision/semantic_segmentation_deeplab_v3.py
@@ -1,0 +1,340 @@
+"""
+Title: Semantic segmentation using DeepLabV3+
+Author: [Alexia Audevart](https://github.com/aaudevart)
+Date created: 2023/13/04
+Last modified: 2023/13/04
+Description: Use the DeepLabV3+ pre-trained models via the Keras CV API for Semantic Segmentation.
+Accelerator: GPU
+"""
+
+"""
+## Introduction
+
+Semantic segmentation, with the goal to assign semantic labels to every pixel in an
+image, is an essential computer vision task.
+
+In this example, we implement
+the **DeepLabV3+** model for semantic segmentation in healthcare. This model is a
+fully-convolutional
+architecture that performs well on semantic segmentation benchmarks.
+
+The goal is to detect blood vessels on 2D PAS-stained histology images from healthy human
+kidney tissue slides.
+
+
+### References:
+
+- [Encoder-Decoder with Atrous Separable Convolution for Semantic Image
+Segmentation](https://arxiv.org/abs/1802.02611)
+- [Rethinking Atrous Convolution for Semantic Image
+Segmentation](https://arxiv.org/abs/1706.05587)
+- [DeepLab: Semantic Image Segmentation with Deep Convolutional Nets, Atrous Convolution,
+and Fully Connected CRFs](https://arxiv.org/abs/1606.00915)
+
+### Dataset:
+
+We will use a simplified extract of the [HuBMAP - Hacking the Human Vasculature
+Dataset](https://www.kaggle.com/competitions/hubmap-hacking-the-human-vasculature)
+for training our model. This dataset is obtained from Periodic acid-Schiff (PAS) stained
+tissue slides from human kidneys.
+Each image is labeled with pixel-wise annotations for 2 categories: background and blood
+vessel.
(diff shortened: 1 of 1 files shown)


qlzh727:
Adding Ian as reviewer and Jon for visibility.

examples/vision/semantic_segmentation_deeplab_v3.py
```diff
@@ -0,0 +1,340 @@
+"""
+Title: Semantic segmentation using DeepLabV3+
+Author: [Alexia Audevart](https://github.com/aaudevart)
+Date created: 2023/13/04
```
ianstenbit, line 4: Is this meant to be September 13? (13/09)?

Either way -- YYYY/MM/DD format is the standard here so let's use that.

examples/vision/semantic_segmentation_deeplab_v3.py
```diff
@@ -0,0 +1,340 @@
...
+Each image is labeled with pixel-wise annotations for 2 categories: background and blood
+vessel.
+This dataset focuses on microvasculature in the kidney, including arterioles, capillaries
+and venules. Larger vessels structures, i.e. arteries and veins, were not segmented in
+this dataset.
+"""
+
+"""shell
+!pip install keras-core --upgrade
+!pip uninstall -y keras-cv
```
ianstenbit, line 49: I don't think we should need this uninstall for the example

examples/vision/semantic_segmentation_deeplab_v3.py
```diff
@@ -0,0 +1,340 @@
...
+print("TensorFlow version:", tf.__version__)
+print("KerasCV version:", keras_cv.__version__)
+
+"""
+## Accelerator
+
+Detect hardware, return appropriate distribution strategy
+"""
+
+try:
```
ianstenbit, line 83: For the example, I think we can exclude this bit and forego a strategy altogether. TF will just default to using the GPU in this case so we don't need to worry about distribution

examples/vision/semantic_segmentation_deeplab_v3.py
```diff
@@ -0,0 +1,340 @@
...
+            yield (image, mask)
+
+    return gen
+
+
+# Prepare the training dataset.
+train_gen = tf.data.Dataset.from_generator(
+    generator=gen_pairs(train_img_list),
+    output_signature=(
+        tf.TensorSpec(shape=(IMG_WIDTH, IMG_HEIGHT, 3), dtype=tf.uint8),
```
ianstenbit, line 164: Not critical here, since images are square, but this should be IMG_HEIGHT, IMG_WIDTH (order reversed)

examples/vision/semantic_segmentation_deeplab_v3.py
```diff
@@ -0,0 +1,340 @@
...
+        tf.TensorSpec(shape=(IMG_WIDTH, IMG_HEIGHT, 2), dtype=tf.uint8),
+    ),
+)
+train_gen = train_gen.batch(BATCH_SIZE, drop_remainder=True)
+
+# Prepare the validation dataset.
+val_gen = tf.data.Dataset.from_generator(
+    generator=gen_pairs(val_img_list),
+    output_signature=(
+        tf.TensorSpec(shape=(IMG_WIDTH, IMG_HEIGHT, 3), dtype=tf.uint8),
```
ianstenbit, line 174: (Same here re: height,width)

examples/vision/semantic_segmentation_deeplab_v3.py
```diff
@@ -0,0 +1,340 @@
...
+plot_samples(train_gen.take(2), colormap)
+
+"""
+## Training
+
+We train the model using sparse categorical crossentropy as the loss function, and
+Adam as the optimizer.
+"""
+
+EPOCHS = 16
```
ianstenbit, line 251: For examples, we often try to keep the number of epochs low -- how long does this take to train?
If it's possible to get decent results with fewer epochs we should do that.

ianstenbit (commented):
Generally looks good to me -- thanks for turning this into an example!
Just a few mostly minor comments

divyashreepathihalli:
Hi @aaudevart! Since KerasCV has been deprecated, I will be closing the PR. If you are interested in opening one using KerasHub, please go ahead.

PR #1547, opened 2023-10-10
Update keras_CV guides_master.py

divyashreepathihalli:


scripts/guides_master.py
```diff
@@ -23,6 +23,18 @@
             "path": "custom_image_augmentations",
             "title": "Custom Image Augmentations with BaseImageAugmentationLayer",
         },
+        {
+            "path": "semantic_segmentation_deeplab_v3_plus",
+            "title": "Semantic Segmentation with KerasCV",
+        },
+        {
+            "path": "coco_metrics",
```
ianstenbit, line 31: I'm not sure if we want to publish these other 2 right now -- I'm not certain of the state they are in, and at least for the COCO one we already discuss this in the OD guide.

ianstenbit (approved):
LGTM for adding the segmentation one, but I think we can drop the other 2.

scripts/guides_master.py
divyashreepathihalli: got it! I'll remove them

divyashreepathihalli:
> LGTM for adding the segmentation one, but I think we can drop the other 2.

done!

fchollet (approved):

PR #1562, opened 2023-10-30
Update about.md

ckulkarni97:
Fixing the broken link on this page https://keras.io/about/ under the "Support" section for "our guidelines"

Corrected link : https://github.com/keras-team/keras-io/blob/master/templates/contributing.md


sachinprasadhs:
The previous link has been moved under tf-keras, you can access it here https://github.com/keras-team/tf-keras/blob/master/CONTRIBUTING.md.
But I'm not sure if we can still refer to tf-keras contributing guidelines.

sachinprasadhs (approved):

Issue #1579, opened 2023-11-08
Use read-only permissions in the continuous_integration.yml workflow


pnacht:
Almost all of keras-io's workflows run with safe permissions. However, `continuous_integration.yml` is currently running with the default write-all token.

This issue can be solved in two ways:
- add top-level read-only permissions to the workflow; and/or
- set the default token permissions to read-only in the repo settings.

I'll be sending a PR along with this issue that sets the top-level permissions. If you instead (or also) wish to modify the default token permissions:

1. Open the repo settings
2. Go to [Actions > General](https://github.com/keras-team/keras-io/settings/actions)
3. Under "Workflow permissions", set them to "Read repository contents and packages permissions"

This setting can also be set at the org level to protect all of keras-team's repositories. However, this may break workflows in other repos that are currently implicitly relying on the write-all token.

---

**Disclosure:** My name is Pedro and I work with Google and the [Open Source Security Foundation (OpenSSF)](https://www.openssf.org/) to improve the supply-chain security of the open-source ecosystem.

I'm also working on some of the other keras-team repos to fix similar issues there (https://github.com/keras-team/keras-core/pull/882, https://github.com/keras-team/keras-cv/pull/2075, https://github.com/keras-team/keras-tuner/pull/930, with more to come!).

PR #1580, opened 2023-11-08
Token permissions

pnacht:
Fixes https://github.com/keras-team/keras-io/issues/1579.

This PR adds read-only permissions to continuous_integration.yml.

It also adds some future-proofing to the other workflows by moving write permissions to job-level instead of top-level, keeping only read-only top-level permissions. This way the added permissions aren't available to other jobs potentially added to the workflows in the future.


hertschuh (approved):
LGTM. Thanks for the change!

PR #1596, opened 2023-11-09
Add token_learner.py example for Keras 3

haifeng-jin:
Changed epochs from 20 to 1. To avoid timeout, and for faster demo.


fchollet (approved):
LGTM, thanks

Note: when changing the number of epochs, leave a code comment (or a text comment) saying how many epochs are actually recommended.

PR #1606, opened 2023-11-10
Fix rendering issue for >10 code blocks

mattdangerw:
Maybe not the most beautiful fix, but we now support up to 9999 code blocks per docstring.


fchollet (approved):
Thanks for the fix!

PR #1607, opened 2023-11-10
Update MirNet example for Keras 3.

hertschuh:


hertschuh:
Note: this is run after applying this fix: https://github.com/keras-team/keras/pull/18760

fchollet (approved):
LGTM, thank you

PR #1608, opened 2023-11-10
Keras 3 involution example

mattdangerw:
tf-only

only fix needed was to avoid list of list output for a functional model at the end of the example


fchollet (approved):
LGTM, thank you

PR #1610, opened 2023-11-11
Keras 3 ner_transformers

mattdangerw:
Tensorflow only with preprocessing and some modeling ops in tf, but it should not be too hard to port this to multi backend with tf-only preprocessing in the future.

I believe there was a bad bug in this examples, where it computed a loss of softmax outputs with from_logits=True. Fixed.


fchollet (approved):
LGTM, thank you!

PR #1612, opened 2023-11-11
Keras 3 wgan_gp

mattdangerw:
custom train_step with tf ops and gradient tape, so this should probably stay tf only


fchollet (approved):
LGTM, thank you

PR #1617, opened 2023-11-13
Vision Siamese network

qlzh727:
The accuracy seems to be reduced comparing from the existing example. Not sure if this is Keras issue (the existing example started with very high acc in epoch 1.)


hertschuh:
Oh, I was working on it. I'm seeing the same accuracy issue, which I don't understand.

qlzh727:
ops, seems that I converted the wrong guide, but the performance issue seems to be a problem.

fchollet (approved):
Usually these accuracy issues come from training stability problems. The original code was likely not well reproducible and the training run was cherry picked out of several ones. LGTM

PR #1618, opened 2023-11-13
Bump KerasNLP version to 0.6.3

mattdangerw:
This doesn't actually change any symbols (it's only for adding Keras 3 support), but still good to track the latest release here.


fchollet (approved):
LGTM, thank you

PR #1624, opened 2023-11-18
Updated masked image modeling example for Keras 3.

hertschuh:


hertschuh:
I noticed that the loss is going up during training instead of going down like before. So, I'm not sure what's wrong.

However the MAE metric is going down just like before.

fchollet (approved):
Yeah, weird numbers. Anyhow, LGTM, thank you!

PR #1639, opened 2023-11-22
Update the KerasNLP getting started guide for Keras 3

mattdangerw:


fchollet (approved):
LGTM, thank you!

learn2manage:
This guide is currently very hard to get started.
`pip install -q --upgrade keras-nlp`
`pip install -q --upgrade keras  # Upgrade to Keras 3.`

This will throw dependancy error:
`ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
tensorflow 2.15.0.post1 requires keras<2.16,>=2.15.0, but you have keras 3.0.4 which is incompatible.`

So instead changed to use tf_nightly,
Then in below sections all require tensorflow-text:
1. Inference with a pretrained classifier
2. Fine tuning a pretrained BERT backbone
3. Fine tuning with a custom model
4. Build and train your own transformer from scratch

Following the steps will throw exception like:
`Exception encountered: BertTokenizer requires the `tensorflow-text` package. Please install with `pip install tensorflow-text`.`

Installing tensorflow-text will not work as it is 2.15 that is conflict with tensorflow 2.16 again.

PR #1640, opened 2023-11-22
Re-render transformer pretraining for keras 3

mattdangerw:
Only minor fixes here, was already updated to keras-core a bit ago


fchollet (approved):
LGTM, thanks!

PR #1642, opened 2023-11-23
Keras 3 semantic similarity

mattdangerw:


fchollet (approved):
LGTM, thanks! Will scrub the warning in post

PR #1643, opened 2023-11-23
Add install line for neural_machine_translation_with_keras_nlp

mattdangerw:
Looks like this was already ported, so just adding the pip install line in keeping with other examples.


fchollet (approved):
LGTM, thanks!

PR #1645, opened 2023-11-23
Port fnet with keras-nlp to keras-3

mattdangerw:


fchollet (approved):
LGTM, thanks! Thoughtful of you to also update `examples_master.py`.

PR #1647, opened 2023-11-27
Port sentence_embeddings_with_sbert to Keras 3

mattdangerw:
Only tf backend for now, and `jit_compile=False` to avoid a XLA error with variables on the wrong device.
https://www.tensorflow.org/xla/known_issues#tfvariable_on_a_different_device

The long term fix is probably to move all preprocessing out of the call graph of these models (no `Input(dtype="string")`), but for now we can do the simple fix.


examples/nlp/sentence_embeddings_with_sbert.py
```diff
@@ -47,23 +47,26 @@
 Let's install and import the libraries we need. We'll be using the KerasNLP library in
 this example.
 
-We will also enable [mixed perceciosn](https://www.tensorflow.org/guide/mixed_precision)
+We will also enable [mixed precision](https://www.tensorflow.org/guide/mixed_precision)
```
hertschuh, line 50: I can't comment on correct line, but can you fix this typo: `batchs` -> `batches` (both lowercase and upper case).

hertschuh: Oh there is another typo at the bottom:
```
How to work and ean money through internet?
```
`ean` -> `earn`

hertschuh (changes requested):

examples/nlp/sentence_embeddings_with_sbert.py
mattdangerw: Thanks! Fixed

examples/nlp/ipynb/sentence_embeddings_with_sbert.ipynb
```diff
@@ -564,22 +568,21 @@
...
+    "        positive = keras.Input(shape=(1,), dtype=\"string\")\n",
+    "        negative = keras.Input(shape=(1,), dtype=\"string\")\n",
     "\n",
     "        ea = encoder(anchor)\n",
     "        ep = encoder(positive)\n",
     "        en = encoder(negative)\n",
     "\n",
-    "        positive_dist = tf.math.reduce_sum(tf.math.pow(ea - ep, 2), axis=1)\n",
-    "        negative_dist = tf.math.reduce_sum(tf.math.pow(ea - en, 2), axis=1)\n",
+    "        positive_dist = keras.ops.sum(keras.ops.power(ea - ep, 2), axis=1)\n",
```
hertschuh, line 579: Oh, another nit: `ops.square` works too here (and the next line).

hertschuh (approved):

fchollet (approved):
LGTM, thank you!

PR #1648, opened 2023-11-27
Port data parallel training with KerasNLP to Keras 3

mattdangerw:
Changing the title so this is clear this is for tf.distribute


fchollet (approved):
LGTM, thank you.

PR #1655, opened 2023-11-28
Update dependencies

mattdangerw:
This still won't include Keras 3 (because of tf conflict), but includes the correct versions of tensorflow and keras-cv to render everything.

This should work (without GPU support):
pip install -r requirements.txt
pip install -U keras

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -10,10 +10,10 @@ pandas
 jupyter
 pycocotools
 pydot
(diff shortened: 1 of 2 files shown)


sachinprasadhs:
Hi @mattdangerw , Can we change from keras-core to keras 3.0 in the requirements.txt file.
Also, keras-tuner is having latest release 1.4.6

PR #1669, opened 2023-12-04
update Patch convnet to use Keras 3 [tf-only example]

divyashreepathihalli:
The example will be tf-only because it uses `tf.GradientTape()`


fchollet (approved):
LGTM, thank you for the PR!

PR #1678, opened 2023-12-08
Update "Text classification with switch transformer" to use Keras 3 

divyashreepathihalli:
Works on all backends!


fchollet (approved):
Awesome, thanks for the PR! LGTM

PR #1683, opened 2023-12-13
Update RandAugment example to Keras 3

sachinprasadhs:
Updated RandAugment to Keras 3 and made below changes.

* Removed the dependency of imgaug package and replaced it with keras_cv Randomaugment implementation.
* Removed the usage of function `augment`, since it is no longer required.
* Removed the usage of `tf.py_function` and it's discerption , and changed the implementation to `RandAugment layer`.
* Changed saved weight file format to `.weights.h5 ` for Keras 3.
* Removed referenced notebook of the author, since it uses Keras 2 implementation.


fchollet (approved):
LGTM, thank you! 👍

PR #1699, opened 2023-12-30
Fix notebooks using Colab TPU

jeffcarp:
There are 4 notebooks that specify `Accelerator: TPU`, however their IPython metadata still specified `GPU` as the accelerator, so the Colab would open with the incorrect accelerator.

Example at HEAD:
https://colab.research.google.com/github/keras-team/keras-io/blob/master/examples/vision/ipynb/xray_classification_with_tpus.ipynb

Validation of fix:
https://colab.research.google.com/github/jeffcarp/keras-io/blob/fix-colab-tpu-metadata/examples/vision/ipynb/xray_classification_with_tpus.ipynb

Fixed by re-exporting the .py files to .ipynb, using command:

```
python3 tutobooks.py py2nb ../examples/keras_recipes/tfrecord.py ../examples/keras_recipes/ipynb/tfrecord.ipynb
```

### DIFF
--- a/examples/keras_recipes/ipynb/tfrecord.ipynb
+++ b/examples/keras_recipes/ipynb/tfrecord.ipynb
@@ -159,7 +159,9 @@
     "            \"target\": tf.io.FixedLenFeature([], tf.int64),\n",
     "        }\n",
     "        if labeled\n",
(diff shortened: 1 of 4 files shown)


fchollet (approved):
LGTM, thank you!

PR #1708, opened 2024-01-04
Transform double dashes to n-dash using the Smarty plugin.

hertschuh:
This way code blocks, anchors, etc. are not affected.


fchollet (approved):
LGTM

Issue #1711, opened 2024-01-05
using test set as validation set


bz36912:
### Issue Type

Bug

### Source

source

### Keras Version

Keras2.13

### Custom Code

Yes

### OS Platform and Distribution

Windows 11

### Python version

3.8.5

### Current Behavior?

In examples/vision/pointnet.py. You seem to use used your test set as validation set in line 263:
`model.fit(train_dataset, epochs=20, validation_data=test_dataset)`
This would cause the performance and predictive ability of your model to be overly optimistic, since the test set is not separate from your training process. The model is likely to perform poorly with unseen data/points.

### Standalone code to reproduce the issue or tutorial link

```shell
examples/vision/pointnet.py line 263:
`model.fit(train_dataset, epochs=20, validation_data=test_dataset)`
```

simeetnayan81:
I'd like to contribute. @sachinprasadhs

PR #1719, opened 2024-01-10
Add validation dataset to pointnet example

sachinprasadhs:
Added Validation dataset to the model.fit instead of test_dataset.
Fixes: https://github.com/keras-team/keras-io/issues/1711


fchollet (approved):
LGTM, thank you!

PR #1720, opened 2024-01-10
Repeat code for the kerasCV image classification guide

sitamgithub-MSIT:
The reason for these modifications is actually not an error but rather a section of code that was repeated twice during the guide's writing. Here is a picture of it

Screenshot (324)

This code attempts to resize the evaluation dataset for inference; however, it is duplicated twice throughout the py, md, and ipynb files. Additionally, resizing was already completed for the train dataset in the form of an augmentation layer. Consequently, not for the train dataset either.

I did not raise any issues about it because it did not result in any errors. I have just made a pull request and eliminated the recurring instances.

cc: @fchollet

### DIFF
--- a/guides/ipynb/keras_cv/classification_with_keras_cv.ipynb
+++ b/guides/ipynb/keras_cv/classification_with_keras_cv.ipynb
@@ -46,7 +46,7 @@
   },
   {
    "cell_type": "code",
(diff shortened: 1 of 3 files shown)


fchollet (approved):
LGTM, thank you for the fix!

PR #1726, opened 2024-01-12
Added the remaining bounding box formats from keras_cv

singhsegv:
Currently keras_cv supports 3 more bounding box formats than what is mentioned in the documentation. This PR adds those remaining paths in `cv_api_master.py`.


fchollet (approved):
LGTM, thank you!

PR #1729, opened 2024-01-13
Updated classification_with_tfdf.py to fix broken UCI db link

mxm2go:
The gentle folks at UCI seem to have relocated the files to a new URL kdd.ics.edu/databases. Updated BASE_PATH to reflect the new URL.


fchollet (commented):
Thanks for the PR! Please also update the corresponding md and ipynb files with the same.

PR #1744, opened 2024-01-23
Remove keyword argument name `new_shape` from uses of `op.reshape`.

hertschuh:
This is in preparation for the renaming of `new_shape` to `newshape`.


fchollet (approved):
LGTM, thank you!

PR #1757, opened 2024-02-08
Fixed the broken link in handwriting_recognition.py

tilakrayal:
Fixed the broken link of `Pre-processing` in **handwriting_recognition.py**


fchollet (commented):
Thanks for the fix! please replicate the same fix in the `md` and `ipynb` files.

fchollet (approved):

PR #1766, opened 2024-02-17
gnn_citations.(py|md|ipynb) GRU code path does not work

maxwbuckley:
There is an optional GRU code path in the gnn_citations.(py|md|ipynb). However it doesn't currently work.

The combination_type "gated" should be "gru"

If you set the combination_type to "gated" you get back a ValueError:

"Invalid combination type: gated"  from the aggregate method call

Fixing that issue exposes another issue in that "layers.GRU" units must be an integer and not a list as it currently is `hidden_units`

To resolve this I made a method to match the signature of the create_ffn and add multiple GRU layers

However it complicates the example a bit...

Also the use of hidden units like this doesn't work if  the hidden_layer sizes are different.

If you are happy with this change I will also update the ipynb and md files.

### DIFF
--- a/examples/graph/gnn_citations.py
+++ b/examples/graph/gnn_citations.py
@@ -415,15 +415,8 @@ class GraphConvLayer(layers.Layer):
         self.normalize = normalize
 
         self.ffn_prepare = create_ffn(hidden_units, dropout_rate)
-        if self.combination_type == "gated":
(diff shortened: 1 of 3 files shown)


fchollet (commented):
Thanks for the PR! The code looks good -- please add the `md` and `ipynb` files.

fchollet (approved):

PR #1767, opened 2024-02-17
Updated Semi-supervised SimCLR flag for Keras v3

lpizzinidev:
This example already seems to support Keras v3, I just updated the related flag and fixed the formatting.


fchollet (approved):
LGTM, thanks!

PR #1779, opened 2024-03-02
Updated Large-scale multi-label text classification example for Keras v3

lpizzinidev:


fchollet (commented):
Thanks for the PR!

This example will need more changes, for instance there are `tf.keras` references remaining. I think the whole example can be made backend-agnostic (with preprocessing in tf.data).

> examples/nlp/.DS_Store

Please revert this file.

lpizzinidev:
@fchollet
Thanks for the feedback! (and sorry for messing up the files)
Maybe I'm missing something, but I don't think that [ragged tensors](https://github.com/keras-team/keras-io/blob/faf8ec1c12e168ec0d0274993144d64de78cb58b/examples/nlp/multi_label_classification.py#L147) are currently supported by Keras v3 so I'm not sure we can make this backend-agnostic.

examples/nlp/multi_label_classification.py
```diff
@@ -30,8 +30,9 @@
 ## Imports
```
jeffcarp, line 30: Can you update `Last modified` above and add yourself to the authors list?

examples/nlp/multi_label_classification.py
```diff
@@ -145,7 +146,7 @@
 """
 
 terms = tf.ragged.constant(train_df["terms"].values)
```
jeffcarp, line 148: Would it work to use padding here instead of ragged tensors?
https://keras.io/api/ops/numpy/#pad-function

PR #1783, opened 2024-03-09
Updated Proximal Policy Optimization example for Keras v3

lpizzinidev:
Updates the Proximal Policy Optimization example for Keras v3.
The example will be tf-only as the `GradientTape` function is used.


fchollet (commented):
Thanks for the PR! The code looks good. Please add the generated files.

lpizzinidev:
@fchollet Thanks for reviewing 👍
I added the generated files and updated the script as well (was working in Colab but not with `autogen.py`) by migrating to Gymnasium (OpenAI gym is deprecated) and fixing other issues.

fchollet (approved):
LGTM, thank you!

PR #1793, opened 2024-03-13
Update API generation for losses

sachinprasadhs:


scripts/cv_api_master.py
```diff
@@ -257,6 +257,49 @@
 #     ],
 # }
 
+LOSSES_MASTER = {
+    "path": "losses/",
+    "title": "Losses",
```
fchollet, line 262: KerasCV Losses

scripts/cv_api_master.py
```diff
@@ -388,5 +431,5 @@
     "path": "keras_cv/",
     "title": "KerasCV",
     "toc": True,
-    "children": [LAYERS_MASTER, MODELS_MASTER, BOUNDING_BOX_MASTER],
+    "children": [LAYERS_MASTER, MODELS_MASTER, BOUNDING_BOX_MASTER, LOSSES_MASTER],
```
fchollet, line 434: Do we also need metrics?

fchollet (commented):
Thanks for the PR!

scripts/cv_api_master.py
sachinprasadhs: This will anyways generate as `https://keras.io/api/keras_cv/losses/` and will be listed under KerasCV API like below.
Do you still want me to make changes as KerasCV Losses?


image

scripts/cv_api_master.py
sachinprasadhs: We don't have any public class or any metrics exported in the Keras CV repo.

fchollet (approved):

PR #1799, opened 2024-03-15
Update results as we changed torch compile to optimal setting

haifeng-jin:
We have used `torch.compile(model, mode="reduce-overhead")` in the benchmarks. Results are updated accordingly.
The numbers almost remain the same except BERT.
Changed BERT batch size because native PyTorch would run OOM with the old batch sizes using the given compile setting.


fchollet (approved):
LGTM, thank you!

PR #1801, opened 2024-03-16
update torch compile settings

haifeng-jin:
The change was to compile the train or inference function instead of just the model.
Only SegmentAnything train & inference and Stable Diffusion train are affected.
The rest relies on Trainer from HuggingFace or no custom train or inference functions (relies on the torch model call function).

Had to change the batch size for SegmentAnything inference since it runs OOM with the old batch size.


fchollet (approved):
LGTM, thank you for the update.

PR #1803, opened 2024-03-17
Updated Deep Q-Learning for Atari Breakout example for Keras v3

lpizzinidev:
Updates the Deep Q-Learning for Atari Breakout example for Keras v3.
The example will be tf-only as the `GradientTape` function is used.
It also updates the API used to [Gymnasium](https://gymnasium.farama.org/) as OpenAI Gym has been deprecated.


fchollet (commented):
Nice work! The code looks good, you can add the generated files.

fchollet (approved):

PR #1812, opened 2024-03-23
Updated Deep Deterministic Policy Gradient (DDPG) example to Keras v3

lpizzinidev:
Updates the Deep Deterministic Policy Gradient (DDPG) example for Keras v3.
The example will be tf-only as the GradientTape function is used.
It also updates the API used to [Gymnasium](https://gymnasium.farama.org/) as OpenAI Gym has been deprecated.


fchollet (commented):
Looking good, thank you! Please add the md and ipynb files.

fchollet (approved):

PR #1814, opened 2024-04-01
update benchmarks

haifeng-jin:
Please check [here](https://github.com/haifeng-jin/keras-io/blob/4de841a2a8c456b4e8170f173fb174e859db0f26/templates/getting_started/benchmarks.md) for the rendered table.


fchollet (approved):
Thank you for the update!

PR #1818, opened 2024-04-03
remove keras_core .ipynb dir

sachinprasadhs:
Remove the `keras_core` `.ipynb` generated directory since it is no longer used to render, since we have all the tutorials directly under `guides/ipynb` directory here https://github.com/keras-team/keras-io/tree/master/guides/ipynb which is rendered in keras.io


fchollet (approved):
Thanks for the cleanup!

PR #1819, opened 2024-04-03
update benchmarks

haifeng-jin:
Rendered view: https://github.com/haifeng-jin/keras-io/blob/61a9e554c6062b45a358c91231c27d4bc06d861b/templates/getting_started/benchmarks.md


fchollet (approved):
LGTM, thank you!

PR #1820, opened 2024-04-04
fix broken url in trainer_pattern.py

hvgazula:
as the title says


fchollet (commented):
Thanks for the PR! Please also update the same line in the ipynb and md files.

hvgazula (commented):
Done

fchollet (approved):
LGTM, thanks!

PR #1822, opened 2024-04-05
Fixed types in nlp examples

SuryanarayanaY:
Fixed types in nlp examples


fchollet (approved):
LGTM, thanks for the fixes!

PR #1857, opened 2024-05-08
Migrate Review classification using Active learning to Keras 3.

sachinprasadhs:
Migrate Review classification using Active learning to Keras 3 backend agnostic .


fchollet (approved):
Nice work, thank you!

PR #1871, opened 2024-05-24
Fix markdown in float8 example

james77777778:
Just a simple fix to better align the list more neatly

EDITED:
Fix `PROJECT_URL` for KerasNLP

### DIFF
--- a/examples/keras_recipes/md/float8_training_and_inference_with_transformer.md
+++ b/examples/keras_recipes/md/float8_training_and_inference_with_transformer.md
@@ -20,11 +20,10 @@ floating point with nearly no degradation in accuracy.
 
 In detail, there are two distinct types of FP8: E4M3 and E5M2, useful in
 different parts of training.
-- E4M3: It consists of 1 sign bit, 4 exponent bits and 3 bits of mantissa. It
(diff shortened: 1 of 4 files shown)


examples/keras_recipes/md/float8_training_and_inference_with_transformer.md
```diff
@@ -53,7 +55,7 @@ Note: The dependency on TensorFlow is only required for data processing.
 
 
 ```python
```
fchollet, line 57: What is going on with this block? I would not expect it to render in this form

fchollet (commented):
Thanks for the PR!

examples/keras_recipes/md/float8_training_and_inference_with_transformer.md
james77777778: image

The list doesn't render correctly.
The format is referenced from this one:
https://github.com/keras-team/keras-io/blob/master/examples/nlp/parameter_efficient_finetuning_of_gpt2_with_lora.py#L314-L324

fchollet (approved):

PR #1872, opened 2024-05-24
BUGFIX: "indices" does not shuffling

amir-saniyan:
`tf.keras.random.shuffle` returns random elements and does not apply in-place


fchollet (commented):
LGTM, thanks -- can you also update the md and ipynb files with the same?

amir-saniyan:
> LGTM, thanks -- can you also update the md and ipynb files with the same?

md and ipynb files updated.

fchollet (approved):
LGTM, thanks!

PR #1877, opened 2024-05-27
fix: Correct typo in classification_with_keras_cv.py

chanway-speedoc:


fchollet (commented):
Thanks for the PR -- please also apply the fix to the md and ipynb files

chanway-speedoc:
Updated as per requested. I am wondering whether there is a more automated manner to update all 3 files at once.

chanway-speedoc:
> Thanks for the PR -- please also apply the fix to the md and ipynb files


Updated as per requested. I am wondering whether there is a more automated manner to update all 3 files at once.

fchollet (approved):

Issue #1876, opened 2024-05-26
Timeseries Classification Transformer: Last -> First


duck-bongos:
### Issue Type

Documentation Bug

### Source

source

### Keras Version

2.13.1

### Custom Code

No

### OS Platform and Distribution

Google Colab

### Python version

Python 3

### GPU model and memory

N/A

### Current Behavior?

The tutorial's (https://keras.io/examples/timeseries/timeseries_classification_transformer/) got a line in it that doesn't match the output network example. This is the line in the 3rd code block:
```
...
x = layers.GlobalAveragePooling1D(data_format="channels_last")(x)
...
```
I copied and pasted the tutorial to play around with and found it returns a network output _different_ than what is expected in the tutorial.

<img width="837" alt="Screen Shot 2024-05-26 at 4 11 57 PM" src="https://github.com/keras-team/keras-io/assets/10129855/fa37b5e8-f793-4d71-a7a8-9f0413de280f">

When I changed the line to 
```
...
x = layers.GlobalAveragePooling1D(data_format="channels_first")(x)
...
```
The network parameters lined up with the tutorial!
<img width="848" alt="Screen Shot 2024-05-26 at 4 11 26 PM" src="https://github.com/keras-team/keras-io/assets/10129855/9cf64c9f-0c54-4e00-935e-bbaace521be9">

### Standalone code to reproduce the issue or tutorial link

```shell
It's in the tutorial :)
```

### Relevant log output

```shell
No additional log output
```

chunduriv:
@duck-bongos, 

Thank you for reporting the issue. It will be fixed in [PR](https://github.com/keras-team/keras-io/pull/1882).

PR #1882, opened 2024-06-03
Update data_format to `channels_first`

chunduriv:
To match the output as per the tutorial, update data_format to `"channels_first"`.

Fixes: https://github.com/keras-team/keras-io/issues/1876


fchollet (approved):

fchollet (commented):
There's a problem here -- the input should be `(batch, time, features)` and the output should be `(batch, features)`. So it seems `"channels_last"` is correct.

PR #1883, opened 2024-06-10
fix metric.reset_state spelling error

ketvector:
The method is called reset_state in the source.

Source: https://github.com/keras-team/keras/blob/master/keras/src/metrics/metric.py


fchollet (approved):
LGTM

PR #1904, opened 2024-08-01
Correct minor issue in the guide "Understanding masking & padding"

palc001:
Correct the axis for computing the denominator of softmax in the example which creates a `TemporalSoftmax` class.

### DIFF
--- a/guides/ipynb/understanding_masking_and_padding.ipynb
+++ b/guides/ipynb/understanding_masking_and_padding.ipynb
@@ -526,7 +526,7 @@
     "        broadcast_float_mask = tf.expand_dims(tf.cast(mask, \"float32\"), -1)\n",
     "        inputs_exp = tf.exp(inputs) * broadcast_float_mask\n",
     "        inputs_sum = tf.reduce_sum(\n",
-    "            inputs_exp * broadcast_float_mask, axis=-1, keepdims=True\n",
+    "            inputs_exp * broadcast_float_mask, axis=1, keepdims=True\n",
     "        )\n",
     "        return inputs_exp / inputs_sum\n",
     "\n",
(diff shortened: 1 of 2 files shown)


palc001:
@fchollet As per my understanding, [a softmax over the time dimension (axis 1) of an input sequence](https://www.tensorflow.org/guide/keras/understanding_masking_and_padding#:~:text=a%20layer%20that%20computes%20a%20softmax%20over%20the%20time%20dimension%20(axis%201)%20of%20an%20input%20sequence) implies that the softmax should be applied along the sequence length dimension over each embedding dimension.

To illustrate with a better example:
You'd typically use such a `TemporalSoftmax` function to compute attention weights over the given input sequence. Here, you'd want the weights across the sequence to sum to 1 (and not the embeddings for each time-step to sum to 1, as is achieved in the notebook using `axis=-1`)
Rewriting the [code snippet given in the example](https://www.tensorflow.org/guide/keras/understanding_masking_and_padding#:~:text=inputs%20%3D%20keras.Input(shape%3D(None%2C)%2C%20dtype%3D%22int32%22)%0Ax%20%3D%20layers.Embedding(input_dim%3D10%2C%20output_dim%3D32%2C%20mask_zero%3DTrue)(inputs)%0Ax%20%3D%20layers.Dense(1)(x)%0Aoutputs%20%3D%20TemporalSoftmax()(x)%0A%0Amodel%20%3D%20keras.Model(inputs%2C%20outputs)%0Ay%20%3D%20model(np.random.randint(0%2C%2010%2C%20size%3D(32%2C%20100))%2C%20np.random.random((32%2C%20100%2C%201)))) to compute attention weights (I've only changed the variable names, everything else remains the same):
```
inputs = tf.keras.Input(shape=(None,), dtype="int32")
embeddings = tf.keras.layers.Embedding(input_dim=10, output_dim=32, mask_zero=True)(inputs)
weights = tf.keras.layers.Dense(1)(embeddings)
attention_weights = TemporalSoftmax()(weights)

model = tf.keras.Model(inputs, attention_weights)
y = model(np.random.randint(0, 10, size=(32, 100)))
```
With my suggested correction to `axis`, you get the attention weights correctly; and you can further use them to weigh the embeddings for each time-step.

PR #1971, opened 2024-10-21
fix build failure

sachinprasadhs:
`STFTSpectrogram` is not part of Keras 3.6 yet and causing build failure, commenting the part till it will be available as part of next stable release.


sachinprasadhs:
This is good to merge @fchollet

PR #1987, opened 2024-11-11
Add SD3 latent walk example

james77777778:
This example is adapted from https://keras.io/examples/generative/random_walks_with_stable_diffusion/  with the following updates:
- Using `"stable_diffusion_3_medium"` preset
- Using slerp for the interpolation

Note that the GIFs of this example are quite large. Should I reduce the number of steps for generation?

cc @fchollet @divyashreepathihalli


fchollet (approved):
Wonderful! It looks better than the original. Thanks a lot for the contribution!

PR #1997, opened 2024-11-27
Fixes to the "English-to-Spanish Translation with a Sequence-to-Sequence Transformer" Code Example

tomjelen:
This pull request addresses several issues in the "English-to-Spanish translation with a sequence-to-sequence Transformer" code example:

1) Swapped input tensors:
    - Issue: The encoder and decoder inputs were swapped, causing either the model to be trained in the wrong translation direction, or inference happening in the wrong direction.
    - Fix: Aligned the encoder and decoder inputs correctly, so the model is trained and used to translate from English to Spanish. This results in meaningful translations when trained to convergence.

2) Padding handling:
    - Issue: The padding masks were computed but never used.
    - Fix: Propagate the masks out of the embedding layer.
    - Fix: Updated the decoder to handle multiple masks, accommodating padding in both Spanish and English sequences.
    - Fix: Modified the loss computation to exclude padding tokens, preventing them from affecting the training process.

I identified these issues while using the example to build my own Keras-based transformer models. I noticed that the model wasn't producing meaningful translations, which these corrections address. By fixing these issues, this pull request ensures that the example functions properly, allowing others to use it as a reliable foundation or learning resource for their projects.

It seems these issues may have arisen due to regression errors over time, and this update aims to restore the example to its intended functionality.

### DIFF
--- a/examples/nlp/ipynb/neural_machine_translation_with_transformer.ipynb
+++ b/examples/nlp/ipynb/neural_machine_translation_with_transformer.ipynb
@@ -10,7 +10,7 @@
     "\n",
     "**Author:** [fchollet](https://twitter.com/fchollet)<br>\n",
     "**Date created:** 2021/05/26<br>\n",
-    "**Last modified:** 2023/02/25<br>\n",
+    "**Last modified:** 2024/11/18<br>\n",
     "**Description:** Implementing a sequence-to-sequence Transformer and training it on a machine translation task."
    ]
   },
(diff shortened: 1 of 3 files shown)


fchollet (commented):
Thanks for the fixes! The code looks good. Please add the generated files as well.

tomjelen:
Perfect, here you go.

The generated markdown file contained thousands of lines of progress bar updates, so I removed them by hand and checked it was nice and readable.

fchollet (approved):
LGTM, thank you. Good call on cleaning up the logs.

jorcisai:
Many thanks, Tom! As soon as I can, I will test it.

PR #2003, opened 2024-12-10
Add Keras 3 example for "Audio track separation"

johacks:
Hi, I saw "Audio track separation" in the call for contributions, so I implemented an example that separates the vocal track from songs in the MUSDB18 dataset.

Some notes:

- The code has been tested on all frameworks.
- I have also uploaded the notebook running for 3 epochs on all frameworks with A100 GPU:
    - [JAX](https://colab.research.google.com/drive/1R_tpMar6qA6lvcrvNhyFnrQa1F19lka-?usp=sharing): 405s on average per epoch.
    - [Torch](https://colab.research.google.com/drive/1A0BbbSvkfOGeTY-5niBpFZXCmJUGvfys?usp=sharing): 615s on average per epoch.
    - [Tensorflow](https://colab.research.google.com/drive/1NUcp3G1Mjou68W3hV3xyQtJn6QhuJv1v?usp=sharing): 261s on average per epoch.
- I can confirm the convert script runs ok and the keras page generates correctly.

Please let me know if there is anything wrong with the provided script I may have missed.

Thanks!


examples/audio/vocal_track_separation.py
```diff
@@ -0,0 +1,611 @@
...
+## Training
+"""
+
+# Load or create the model
+if path.exists(MODEL_PATH):
+    model = saving.load_model(MODEL_PATH)
+else:
+    model = tfc_tdf_net(keras.Input(sample_batch_x.shape[1:]), name="tfc_tdf_net")
+
+model.summary()
```
fchollet, line 569: Would it be useful to plot the model, or is the result too busy? e.g. `keras.utils.plot_model`

fchollet (commented):
Thanks for the PR! This is excellent work, I enjoyed reading through it 👍

examples/audio/vocal_track_separation.py
```diff
@@ -0,0 +1,611 @@
...
+import typing
+from os import path
+
+import audiomentations as aug
+import ffmpeg
+import keras
+import numpy as np
+import soundfile as sf
+from IPython import display
+from keras import callbacks, layers, ops, optimizers, saving, utils
```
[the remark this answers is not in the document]
johacks: I ended up needing many more uses of saving in refactor, so I kept that import, removed the rest

examples/audio/vocal_track_separation.py
johacks: This is how it would look that after a refactoring i've made grouping the TFC_TDF blocks.
It's a little long, but actually simple in structure. Regardless, It can be further refactored to group into decoder and encoder blocks, but perhaps it wont be as informative.

model

johacks:
Hi @fchollet ,

Thanks for the feedback!

I did a small refactor to further group TFC_TDF and Downsample/Upsample blocks into custom layers so they could be better visualized in a plot. I also renamed symbols to further avoid abbreviations and updated some docstrings to reflect the changes.

The code is currently working on all frameworks after these changes. Also the convert script still runs correctly.

examples/audio/vocal_track_separation.py
fchollet: Ok -- your call on whether to include it.

fchollet (commented):
Thanks for the update -- the changes are looking good! Please add the generated files.

johacks:
Hi again,

I have just pushed the generated files. I also updated layer `TimeFrequencyTransformBlock` to use a flat list instead of list of tuples, because the latter was resulting in weights not being properly tracked if saving the model.

fchollet (approved):
LGTM - Thank you for the excellent contribution!

PR #2014, opened 2024-12-22
Add activations and ops method

shashaka:
I have added descriptions for some activation functions that were not yet documented on keras.io, even though they were already released in Keras 3.7.0.


fchollet (approved):
LGTM, thanks!

PR #2017, opened 2024-12-27
Update the link to the example of the random walk for SD3

james77777778:
We should update the highlight to replace KerasCV SD with KerasHub SD3.

https://keras.io/examples/generative/


fchollet (approved):
LGTM

PR #2019, opened 2025-01-01
migrating multimodal_entailment.py from Keras2 to Keras3 and fixing other issues

Humbulani1234:
Migrating the script `multimodal_entailment.py` from Keras2 to Keras3, and addressing various encountered errors.

The function [make_bert_preprocessing_model](https://github.com/keras-team/keras-io/blob/master/examples/nlp/multimodal_entailment.py#L227) from the script has been motivated by the Tensorflow [example](https://www.tensorflow.org/text/tutorials/bert_glue) according to the script [doc](https://github.com/keras-team/keras-io/blob/master/examples/nlp/multimodal_entailment.py#L219) string. Both the Tensorflow example and the script's function do not work, they produce the following error:

```
  File "/home/humbulani/tensorflow-env/env/lib/python3.11/site-packages/tensorflow/core/function/trace_type/trace_type_builder.py", line 185, in from_value
    ndarray = value.__array__()
              ^^^^^^^^^^^^^^^^^
  File "/home/humbulani/tensorflow-env/env/lib/python3.11/site-packages/keras/src/backend/common/keras_tensor.py", line 108, in __array__
    raise ValueError(
ValueError: Exception encountered when calling layer 'tokenizer' (type KerasLayer).

A KerasTensor is symbolic: it's a placeholder for a shape an a dtype. It doesn't have any actual numerical value. You cannot convert it to a NumPy array.
```

The error is due to calling `tokenizer` with a symbolic tensor argument. Even when tried with `Kerashub` WordPiece tokenizer I get the same error.

**Suggested fix:**

- Removed the function `make_bert_preprocessing_model` and used a `BertTextClassiferPreprocessor` with weights loaded using `from_preset` method.

- Used KerasHub `BertBackbone` instead of the URL path provided to [tfhub](https://tfhub.dev/tensorflow/small_bert/bert_en_uncased_L-2_H-256_A-4/1).

- Added some various changes to the script to get it to work.

- Besides the `ResourceExhaustion` error on my side, the script seems to be working.


fchollet (commented):
Thanks for the PR! Very nice work. Does it run on all backends?

Humbulani1234:
Find the generated `.ipynb` and `.md` files.

**Note**:

- Due to dependence on `tf.data` API, the script still runs on `Tensorflow`. Future work will be to create a `Dataloader` class which is backend-agnostic.

Humbulani1234:
Attached PR that adapts the script `multimodal_entailment.py` to `jax` and `torch`, i.e., make the script ***backend-agnostic***

I have also generated the `.md` and `.ipynb` files.

fchollet (approved):
LGTM, thank you

PR #2020, opened 2025-01-06
Adapting the script fnet_classification_with_keras_hub.py to be Backend-Agnostic

Humbulani1234:
This PR adapts the scripts `fnet_classification_with_keras_hub.py` to `jax` and `torch`, i.e., mak the script ***Backend-Agnostic***.

**Approach**:

- The function [text_dataset_from_directory](https://github.com/keras-team/keras/blob/master/keras/src/utils/text_dataset_utils.py#L14) is actually a wrapper around the [tf.data.Dataset.from_tensor_slices](https://github.com/keras-team/keras/blob/master/keras/src/utils/text_dataset_utils.py#L262)function which is a `tensorflow` function.
- Removed the above function and relied on `pandas` and `python std library` to load the `.txt` files as a `dataframe`.
- Created the `UnifiedPyDataset` which subclasses [keras.utils.PyDataset](https://github.com/keras-team/keras/blob/master/keras/src/trainers/data_adapters/py_dataset_adapter.py#L18) to create data batches during training.
- Generated the `.ipynb` and `.md` files

### DIFF
--- a/examples/nlp/fnet_classification_with_keras_hub.py
+++ b/examples/nlp/fnet_classification_with_keras_hub.py
@@ -133,17 +133,13 @@ def load_data_from_directory(directory, label):
 
 # Function to create Pandas dataframes
 def create_df(directory_pos, directory_neg):
-
     # Load data from both directories
     pos_data = load_data_from_directory(directory_pos, label=1)
     neg_data = load_data_from_directory(directory_neg, label=0)
-
     # Combine data from both pos and neg
     all_data = pos_data + neg_data
-
     # Shuffle the data randomly
     random.shuffle(all_data)
-
     # Create a pandas DataFrame
     df = pd.DataFrame(all_data, columns=["text", "label"])
     return df
@@ -226,7 +222,7 @@ less than the specified sequence length. Otherwise, the sequence is truncated.
 
 tokenizer = keras_hub.tokenizers.WordPieceTokenizer(
     vocabulary=vocab,
-    lowercase=False,
+    lowercase=True,
     sequence_length=MAX_SEQUENCE_LENGTH,
 )
 
@@ -296,13 +292,10 @@ class UnifiedPyDataset(keras.utils.PyDataset):
         low = index * self.batch_size
         # Cap upper bound at array length; the last batch may be smaller
         # if the total number of items is not a multiple of batch size.
-
         high_text = min(low + self.batch_size, len(self.text_x))
-
         # text files batches
         batch_text_x = self.text_x[low:high_text]
         batch_text_y = self.text_y[low:high_text]
-
         # Tokenize the data in dataframe and stack them into an nd.array of axis_0=batch
         text = np.array([[tokenizer(text) for text in batch_text_x]])
         return (
--- a/examples/nlp/fnet_classification_with_keras_hub.py
+++ b/examples/nlp/fnet_classification_with_keras_hub.py
@@ -2,9 +2,10 @@
 Title: Text Classification using FNet
 Author: [Abheesht Sharma](https://github.com/abheesht17/)
 Date created: 2022/06/01
-Last modified: 2022/12/21
+Last modified: 2025/01/06
 Description: Text Classification on the IMDb Dataset using `keras_hub.layers.FNetEncoder` layer.
 Accelerator: GPU
+Made Backend agnostic by: [Humbulani Ndou](https://github.com/Humbulani1234)
 """
 
 """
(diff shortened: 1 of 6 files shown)


Humbulani1234:
**Upon further observation**

This script was already running on all the backends and I was just inserting the line `os.environ['KERAS_BACKEND']=<some backend>` at the wrong place. However, it is a different approach to the problem which indeed strives to remove `tensorflow`, especially the `tf.data.Dataset` API, and thereby make the script backend independent from surface level.

It demonstrates the usage of keras [PyDataset](https://github.com/keras-team/keras/blob/master/keras/src/trainers/data_adapters/py_dataset_adapter.py#L18) for data loading and processing.

examples/nlp/fnet_classification_with_keras_hub.py
```diff
@@ -55,18 +56,25 @@
...
-import os
 
 keras.utils.set_random_seed(42)
 
 """
 Let's also define our hyperparameters.
 """
 BATCH_SIZE = 64
-EPOCHS = 3
+EPOCHS = 1  # maybe adjusted as desired
```
hertschuh, line 77: Please revert and re-run with the 3 epoch training.

examples/nlp/fnet_classification_with_keras_hub.py
```diff
@@ -105,41 +113,58 @@
 """
 
 """
-We'll use the `keras.utils.text_dataset_from_directory` utility to generate
-our labelled `tf.data.Dataset` dataset from text files.
+We'll use Pandas to avoid dependency on `keras.utils.text_dataset_from_directory`
```
hertschuh, line 116: I feel like this adds a lot of code that distracts from the original goal.

You can use torch as a backend and still train with a `tf.data.Dataset`, it just works transparently. You can also train directly with a Pandas series.

examples/nlp/fnet_classification_with_keras_hub.py
```diff
@@ -193,18 +219,21 @@ def train_word_piece(ds, vocab_size, reserved_tokens):
 all sequences are padded to the same length, if the length of the sequence is
 less than the specified sequence length. Otherwise, the sequence is truncated.
 """
+
 tokenizer = keras_hub.tokenizers.WordPieceTokenizer(
     vocabulary=vocab,
-    lowercase=False,
+    lowercase=True,
```
hertschuh, line 225: Why did you have to change this?

examples/nlp/fnet_classification_with_keras_hub.py
```diff
@@ -213,26 +242,89 @@ def train_word_piece(ds, vocab_size, reserved_tokens):
...
-
-
-def make_dataset(dataset):
-    dataset = dataset.map(format_dataset, num_parallel_calls=tf.data.AUTOTUNE)
-    return dataset.shuffle(512).prefetch(16).cache()
+Create the final datasets, method adapted from `PyDataset` doc string
+"""
+
+
+class UnifiedPyDataset(keras.utils.PyDataset):
```
hertschuh, line 249: Something similar to this already exists in Keras. Keras can natively consume Pandas DataFrames (now I'm realizing the documentation doesn't even mention it). You just pass DataFrames as the `x` and `y` parameters to `model.fit()`.

examples/structured_data/classification_with_grn_and_vsn.py
```diff
@@ -2,7 +2,7 @@
 Title: Classification with Gated Residual and Variable Selection Networks
 Author: [Khalid Salama](https://www.linkedin.com/in/khalid-salama-24403144/)
 Date created: 2021/02/10
-Last modified: 2021/02/10
+Last modified: 2025/01/03
```
hertschuh, line 5: This seems unrelated. Can you split these changes of this other examples in a separate PR?

hertschuh (commented):
@Humbulani1234 ,

Thank you for the PR! It seems to me that this could be done with fewer changes for the sake of readability of the example. See my comments in the code.

Humbulani1234:
In all honesty, this PR was solving a problem that was not there. I was making a mistake inserting the line `os.environ['KERAS_BACKEND']=<some backend>` at the wrong place, hence I diagnosed incorrectly that the script only runs on tensorflow. I also feel like it is introducing more complexity, especially not adhering to keras functionality.

hertschuh:
> In all honesty, this PR was solving a problem that was not there. I was making a mistake inserting the line `os.environ['KERAS_BACKEND']=<some backend>` at the wrong place, hence I diagnosed incorrectly that the script only runs on tensorflow. I also feel like it is introducing more complexity, especially not adhering to keras functionality.

Meaning we should just close this PR then?

Humbulani1234:
Yes, I will proceed and close it.

PR #2022, opened 2025-01-08
Adapting the script classification_with_grn_and_vsn to be Backend-Agnostic

Humbulani1234:
This PR adapts the script `classification_with_grn_and_vsn.py`  from `structured_data` examples to `jax` and `torch`, i.e., make the script ***Backend-Agnostic***.

**Approach**:

- Modified the model architecture and removed the `Preprocessing` layers and use them with the `tf.data.Datasets`. Refer to [this](https://github.com/keras-team/keras/blob/master/keras/src/layers/preprocessing/string_lookup.py#L42) Keras doc string for more.
- Generated the `.ipynb` and `.md` files


Humbulani1234:
Resend

PR #2030, opened 2025-01-17
fixed a small mistake in the transformer translation example

yasserhcn:


SamanehSaadat (approved):
Thanks!

fchollet (commented):
Thanks for the fix! Please apply the same fix in the corresponding md and ipynb files.

fchollet (approved):
LGTM, thank you!

PR #2054, opened 2025-02-28
Adapt masked language modeling.py script to backend-agnostic

Humbulani1234:
Here is my attempt to making this script backend-agnostic


examples/nlp/masked_language_modeling.py
```diff
@@ -117,7 +119,7 @@ def get_data_from_text_files(folder_name):
 train_df = get_data_from_text_files("train")
 test_df = get_data_from_text_files("test")
 
-all_data = train_df.append(test_df)
+all_data = train_df._append(test_df)
```
hertschuh, line 122: I wouldn't use a protected method like `_append`. It looks like the official replacement for `append` is `concat`.

examples/nlp/masked_language_modeling.py
```diff
@@ -301,27 +304,142 @@ def bert_module(query, key, value, i):
 
 
 class MaskedLanguageModel(keras.Model):
-    def train_step(self, inputs):
+
+    def train_step(self, *args, **kwargs):
```
hertschuh, line 308: Looking at the code for the custom train step below, I don't think a custom train step is needed at all, I think it can be removed.

The only things that might need to be overridden are `compute_loss` and `compute_metrics`, and those can be implemented in a backend agnostic way. This would greatly simplify the code.

My guess is that this example was created before `compute_loss` existed.

examples/nlp/masked_language_modeling.py
```diff
@@ -475,16 +593,34 @@ def create_classifier_bert_model():
 When you want to deploy a model, it's best if it already includes its preprocessing
 pipeline, so that you don't have to reimplement the preprocessing logic in your
 production environment. Let's create an end-to-end model that incorporates
-the `TextVectorization` layer, and let's evaluate. Our model will accept raw strings
-as input.
+the `TextVectorization` layer inside evalaute method, and let's evaluate.
```
hertschuh, line 596: Typo: `evaluate`.

hertschuh (changes requested):
Thanks for the PR!

I think this can be greatly simplified with modern Keras, no need for a custom `train_step` from what I can tell.

hertschuh (approved):

Humbulani1234:
Find PR.

Thou I've managed to get this to work on the `tensorflow` backend, `jax` and `torch` are still returning errors. And to my assessment the issue is that the `loss_fn` has a `None` reduction strategy and hence a scalar value is not returned for gradients calculations under `jax` and `torch`.

This also extends to the `loss_tracker` since the loss must not be a scalar. Previously i had tackled this issue by adjusting the logics accordingly under the `train_step` function.

Maybe we should change the reduction strategy and also change the `loss_tracker`, or probably I'm getting something wrong here in my implementation.

hertschuh:
> Find PR.
> 
> Thou I've managed to get this to work on the `tensorflow` backend, `jax` and `torch` are still returning errors. And to my assessment the issue is that the `loss_fn` has a `None` reduction strategy and hence a scalar value is not returned for gradients calculations under `jax` and `torch`.

On JAX, you can only take the gradient of a function that returns a scalar value, unlike on Tensorflow. I see that in the earlier version of your PR, you were doing the `sum` on both JAX and Torch.
 
> This also extends to the `loss_tracker` since the loss must not be a scalar. Previously i had tackled this issue by adjusting the logics accordingly under the `train_step` function.

Do you understand why the loss must not be a scalar and why `loss_tracker` exists in the first place?
 
> Maybe we should change the reduction strategy and also change the `loss_tracker`, or probably I'm getting something wrong here in my implementation.

I think this could work:
- keep the reduction as `None` on the loss object
- compute the unreduced loss
- update `loss_tracker` with it
- do a `sum` on the loss before returning it from `compute_loss`

Basically change the last line of code in `compute_loss`: `return keras.ops.sum(loss)`.

Would that work?

Notes:
1. There was an error in a check, you'll need to reformat the code
2. You'll need to regenerate the ipynb and md files.

hertschuh:
```
/home/humbulani/tensorflow-env/env/lib/python3.11/site-packages/keras/src/models/functional.py:252: UserWarning: The structure of `inputs` doesn't match the expected structure.
Expected: ['keras_tensor_48']
Received: inputs=Tensor(shape=torch.Size([32, 256]))
  warnings.warn(msg)
```
This will fail with in a future version of Keras.

I believe it's because of this:
```
    inputs = classifer_model.inputs
    outputs = classifer_model.outputs
    end_to_end_model = ModelEndtoEnd(inputs, outputs, name="end_to_end_model")
```
I believe `model.inputs` is always a list (even though you didn't create the model with a list) but then you're passing tensors directly.

I think this should fix it:
```
    inputs = classifer_model.inputs[0]
    outputs = classifer_model.outputs
    end_to_end_model = ModelEndtoEnd(inputs, outputs, name="end_to_end_model")
```

Humbulani1234:
Find PR

PR #2055, opened 2025-03-04
Adds new landing page and updates internal page styling

alexabbott:
This pull request revises the landing page code to align with new designs.

This pull request also updates the internal pages to ensure consistent navigation and styling with the landing page.


theme/css/colors.css
```diff
@@ -0,0 +1,27 @@
...
+  --border-grey: #DCDCDC;
+  --border-red: #A80000;
+
+  /* Shadow Colors */
+  --shadow-grey: #9C9C9C;
+
+  /* Text Colors */
+  --text-black: #292321;
+  --text-red: #A30000;
+}
```
mattdangerw, line 27: lots of missing newlines at end of files.

theme/icons/.DS_Store
mattdangerw, line 1: remove all `.DS_Store` files

theme/js/index.js
```diff
@@ -0,0 +1,126 @@
...
+  mobileNavControls.style.display = 'flex';
+  mobileNavSearchForm.style.display = 'none';
+});
+
+mobileNavMenu.addEventListener('click', () => {
+  mobileNavControls.style.display = 'flex';
+  mobileNavSearchForm.style.display = 'none';
+});
+
+// Parallax functionality
```
mattdangerw, line 68: I wonder if this whole code block would be less jerky if you dynamically updated the element css to switch between position fixed and relative--so the render engine will keep fixed rather than trying to dynamically set a height in js. Not a web dev expert though!

theme/landing.html
```diff
@@ -1,259 +1,686 @@
...
+        <div class="explore stacked__item">
+          <div class="explore__wrapper">
+            <div class="explore__column">
+              <div class="code__container">
+                <button class="code__copy--button">
+                  <i class="icon--copy"></i>
+                </button>
+                <pre>
+                  <code class="language-python">
+    inputs = keras.Input(shape=(32, 32, 3))
```
mattdangerw, line 205: We should remove the tab from these pre blocks so they copy without extra spaces. Instead just style the code card padding to get whatever left alignment we want.

mattdangerw (changes requested):
Most of the comments are on that tracking sheet, but adding some comments here that are more code specific as well!

PR #2056, opened 2025-03-05
Adapt text classification from scratch

Humbulani1234:
Find PR with `.md` and `.ipynb` generated.


SamanehSaadat:
Hi @Humbulani1234 

Could you share more details about why this change is needed?

Humbulani1234:
Good day,

My understanding is that the final end-to-end model must accept strings as inputs. JAX and Torch do not process strings as part of the compiled model - only Tensorflow does.

So, I created an end-to-end model that accepts strings by overriding evaluate to process strings first and not as part of the model.

PR #2062, opened 2025-03-14
 Add API doc for BasNet,Efficientnet,ViT,SegFormer,RetinaNet,MobileNet

laxmareddyp:
Adding Missing API doc for Keras Hub Models BasNet,Efficientnet,ViT,SegFormer,RetinaNet,MobileNet.

### DIFF
--- a/scripts/hub_master.py
+++ b/scripts/hub_master.py
@@ -509,52 +509,6 @@ MODELS_MASTER = {
                 },
             ],
         },
-        """
-        # Code is not available in keras-hub v0.19.1(stable),
-        # Will be enabled once it is available to generate the API doc.
-        {
-            "path": "cspnet/",
-            "title": "CSPNet",
-            "toc": True,
-            "children": [
-                {
-                    "path": "cspnet_image_converter",
-                    "title": "CSPNetImageConverter",
-                    "generate": [
-                        "keras_hub.layers.CSPNetImageConverter",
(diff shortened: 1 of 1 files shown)


scripts/hub_master.py
```diff
@@ -467,6 +509,52 @@
                 },
             ],
         },
+        """
+        # Code is not available in keras-hub v0.19.1(stable),
```
mattdangerw, line 513: Don't submit a commented out thing like this, it's more likely to just get out of date. Just drop this or stick it in a separate pr we can land whenever things are ready.

scripts/hub_master.py
```diff
@@ -1949,6 +2247,6 @@
         {
             "path": "presets/",
             "title": "Pretrained models list",
-       },
+        },
     ],
-}
+}
```
mattdangerw, line 2252: why are you removing the newline here?

scripts/hub_master.py
laxmareddyp: Sure.Will drop and update the PR. Thanks

scripts/hub_master.py
laxmareddyp: I think i should run black command here to eliminate this.

divyashreepathihalli (approved):

hertschuh (approved):

PR #2070, opened 2025-03-27
Fix CI

SamanehSaadat:
`tf-keras==2.19` requires `tensorflow==2.19` but the latest version of `tf-text` is `2.18` and requires `tensorflow==2.18` which creates a requirements conflict. To workaround this issue, we pin `tf-keras==2.18` to remove the strict dependency on `tensorflow==2.19`.

This PR also updates the `keras` and `keras-hub` to their latest version.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -14,5 +14,5 @@ tensorflow==2.18.1
 keras-cv
 keras-nlp
 keras-tuner
-tf_keras==2.18.0
+tf-keras==2.18.0
 keras-hub
--- a/requirements.txt
+++ b/requirements.txt
@@ -16,3 +16,4 @@ keras-nlp
 keras-tuner
 tf-keras==2.18.0
 keras-hub
+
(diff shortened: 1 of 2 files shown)


requirements.txt
```diff
@@ -10,9 +10,9 @@ pandas
...
 pydot
 boto3
-tensorflow
+tensorflow==2.18.1
 keras-cv
 keras-nlp
 keras-tuner
-tf_keras
-keras-hub-nightly  # TODO: update to keras-hub.
+tf_keras==2.18.0
```
mattdangerw, line 17: random nit, but `tf-keras`. the hyphen is the actual package name, though pip accepts both

requirements.txt
```diff
@@ -10,9 +10,9 @@ pandas
...
+keras-hub
```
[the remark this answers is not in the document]
SamanehSaadat: Done

requirements.txt
SamanehSaadat: changed to `tf-keras`

PR #2072, opened 2025-03-31
Update transformer_asr.py

MeysamMoghaddam:
Add small epsilon (1e-10) to stddevs to prevent division by zero

Modified the normalization step by adding a small constant to stddevs  to avoid potential division by zero errors when standard deviation is zero.


examples/audio/transformer_asr.py
```diff
@@ -372,7 +372,7 @@ def path_to_audio(path):
     # normalisation
     means = tf.math.reduce_mean(x, 1, keepdims=True)
     stddevs = tf.math.reduce_std(x, 1, keepdims=True)
-    x = (x - means) / stddevs
+    x = (x - means) / (stddevs + 1e-10)
```
mattdangerw, line 375: does using keras.epsilon work? and can you add the generated files?

PR #2078, opened 2025-04-01
Fix incorrect reference to model in Transformer class

MeysamMoghaddam:
**Description:**
In the train_step and test_step methods of the Transformer class, there's a reference to model.compute_loss() which should be self.compute_loss() instead. Since we're inside the class methods, we should use self to refer to the current instance of the model.

**Files Changed:**

[transformer_asr.py](https://github.com/keras-team/keras-io/blob/master/examples/audio/transformer_asr.py#L182)

**Changes Made:**

Replaced model.compute_loss() with self.compute_loss() in both train_step and test_step methods

**Impact:**
This correction ensures proper access to the parent class's compute_loss method through the correct instance reference. The original code would raise a NameError since model is not defined in this context, while self properly refers to the current Transformer instance.


mattdangerw:
Can you updated the generated files? markdown and ipynb?

PR #2080, opened 2025-04-03
 Add RematScope API to keras-io

laxmareddyp:


scripts/api_master.py
```diff
@@ -1964,6 +1964,13 @@
                         "keras.utils.Config",
                     ],
                 },
+                {
```
divyashreepathihalli, line 1967: can you also add the reference to - https://github.com/keras-team/keras/blob/e61e83df13a43301ff132a62ed998d534ee49089/keras/src/backend/common/remat.py#L137

scripts/api_master.py
```diff
@@ -1965,6 +1965,8 @@
                     ],
                 },
                 {
+                    # reference
```
divyashreepathihalli, line 1968: it is commented out

divyashreepathihalli:
oh I meant add the keras.remat api documentation too

scripts/api_master.py
```diff
@@ -1964,6 +1964,14 @@
                         "keras.utils.Config",
                     ],
                 },
+                {
+                    "path": "resource_management_utils",
+                    "title": "Resource management utilities",
```
divyashreepathihalli, line 1969: I would prefer for the title to be Rematerialization.
@hertschuh do you have any thoughts?

hertschuh: Yes, let's call it "Rematerialization"

divyashreepathihalli (approved):

hertschuh (approved):

divyashreepathihalli:
@laxmareddyp - please update this to 3.9.2 - https://github.com/keras-team/keras-io/blob/b34da6be73298ef2962465360415ab33386b198a/scripts/autogen.py#L35

PR #2083, opened 2025-04-08
Add KerasRS API documentation and examples

abheesht17:
~Do **not** review now.~

TODO:

- [x] Add other layers when they are added to the repo.
- [x] Fix images and links in examples.
- [x] Add generated files [they aren't here on this PR yet, because we can generate them once we release].


scripts/autogen.py
```diff
@@ -392,6 +394,12 @@ def sync_tutobook_templates(self):
                 os.makedirs(dst_dir)
                 copy_inner_contents(dir_path / "md", dst_dir, ext=".md")
 
+        shutil.copytree(
```
hertschuh, line 397: Add a TODO here?

scripts/master.py
```diff
@@ -42,5 +43,6 @@
         KERAS2_API_MASTER,
         TUNER_MASTER,
         HUB_MASTER,
+        RS_MASTER,
```
hertschuh, line 46: Isn't that the thing you should remove for now so that KERAS_RS doesn't appear in the menu?

scripts/autogen.py
abheesht17: No, this isn't a TODO. We have to copy over the examples to KerasRS

scripts/master.py
abheesht17: Nope, all of that is in the HTML files

hertschuh (approved):

PR #2085, opened 2025-04-12
Example (nl_image_search.py) update to work with newer version of TF in Keras 2 compat mode.

vinayak19th:
Multiple Bug fixes for the code to be usable with newer tf versions (tested with 2.19.0) under keras 2 compat mode :
Change log:

* Removed TF-Addons as it doesn't work with newer version and was only being used for the AdamW optimizer which is part of keras natively now
* [Fixed tf.nn.gelu call](https://github.com/keras-team/keras-io/commit/afc6065d062240062974af715631c116442b0104) : Changed tf.nn.gelu -> tf.keras.activations.gelu so that modern keras 2 backward compat can use it within the model api
* Updated the hub.KerasLayer calls to use the correct arguments
* [Update training code to include steps per epoch](https://github.com/keras-team/keras-io/commit/e60cd602e8ff3ebb5e65e2028894d070799b187e) : Updated the training code to have `steps_per_epoch` as a variable so that model.fit can show the total steps
* [Fixed file loading and delete paths](https://github.com/keras-team/keras-io/commit/ec1e16320dab183fa32b30d0eb2cd98492d48a03) : Fixed the file paths to reflect the effect of the newer unzip process

### DIFF
--- a/examples/vision/nl_image_search.py
+++ b/examples/vision/nl_image_search.py
@@ -241,7 +241,7 @@ def project_embeddings(
 ):
     projected_embeddings = layers.Dense(units=projection_dims)(embeddings)
     for _ in range(num_projection_layers):
-        x = tf.nn.gelu(projected_embeddings)
+        x = tf.keras.activations.gelu(projected_embeddings)
         x = layers.Dense(projection_dims)(x)
         x = layers.Dropout(dropout_rate)(x)
         x = layers.Add()([projected_embeddings, x])
@@ -298,7 +298,7 @@ def create_text_encoder(
         "https://tfhub.dev/tensorflow/bert_en_uncased_preprocess/2",
         name="text_preprocessing",
     )
-    # Load the pre-trained BERT model to be used as the base encoder.
+    # Load the pre-trained BERT model to be used as the base encoder with trainable set to false.
     bert = hub.KerasLayer("https://tfhub.dev/tensorflow/small_bert/bert_en_uncased_L-4_H-512_A-8/1",trainable=False,name="bert")
     # Receive the text as inputs.
     inputs = layers.Input(shape=(), dtype=tf.string, name="text_input")
--- a/examples/vision/nl_image_search.py
+++ b/examples/vision/nl_image_search.py
@@ -75,7 +75,7 @@ annotation_file = os.path.join(annotations_dir, "captions_train2014.json")
 
 Download caption annotation files
 if not os.path.exists(annotations_dir):
-    annotation_zip = tf.keras.utils.get_file(
+    annotation_zip = keras.utils.get_file(
         "captions.zip",
         cache_dir=os.path.abspath("."),
         origin="http://images.cocodataset.org/annotations/annotations_trainval2014.zip",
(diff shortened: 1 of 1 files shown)


examples/vision/nl_image_search.py
```diff
@@ -228,7 +241,7 @@ def project_embeddings(
 ):
     projected_embeddings = layers.Dense(units=projection_dims)(embeddings)
     for _ in range(num_projection_layers):
-        x = tf.nn.gelu(projected_embeddings)
+        x = tf.keras.activations.gelu(projected_embeddings)
```
divyashreepathihalli, line 244: use -  keras.activations.glu(x, axis=-1)

examples/vision/nl_image_search.py
```diff
@@ -285,13 +298,8 @@ def create_text_encoder(
...
     )
-    # Load the pre-trained BERT model to be used as the base encoder.
-    bert = hub.KerasLayer(
-        "https://tfhub.dev/tensorflow/small_bert/bert_en_uncased_L-4_H-512_A-8/1",
-        "bert",
-    )
-    # Set the trainability of the base encoder.
-    bert.trainable = trainable
+    # Load the pre-trained BERT model to be used as the base encoder with trainable set to false.
+    bert = hub.KerasLayer("https://tfhub.dev/tensorflow/small_bert/bert_en_uncased_L-4_H-512_A-8/1",trainable=False,name="bert")
```
divyashreepathihalli, line 302: format the code

examples/vision/nl_image_search.py
```diff
@@ -407,7 +415,7 @@ def test_step(self, features):
 )
 dual_encoder = DualEncoder(text_encoder, vision_encoder, temperature=0.05)
 dual_encoder.compile(
-    optimizer=tfa.optimizers.AdamW(learning_rate=0.001, weight_decay=0.001)
+    optimizer=tf.keras.optimizers.AdamW(learning_rate=0.001, weight_decay=0.001)
```
divyashreepathihalli, line 418: use keras 3  - keras.optimizers.AdamW

divyashreepathihalli (commented):
Can you update this example for keras 3?

vinayak19th:
> Can you update this example for keras 3?

I was having an issue with the BERT model, but I can try fixing that later

examples/vision/nl_image_search.py
vinayak19th: Updated in the branch

divyashreepathihalli (commented):
I still see a lot of tf ops in the file

vinayak19th:
> > Can you update this example for keras 3?
> 
> I was having an issue with the BERT model, but I can try fixing that later

This is the exact error message

```
ValueError: Exception encountered when calling layer 'text_preprocessing' (type KerasLayer).

A KerasTensor is symbolic: it's a placeholder for a shape an a dtype. It doesn't have any actual numerical value. You cannot convert it to a NumPy array.

Call arguments received by layer 'text_preprocessing' (type KerasLayer):
  • inputs=<KerasTensor shape=(None,), dtype=string, sparse=False, ragged=False, name=text_input>
  • training=None
```

PR #2093, opened 2025-04-30
Fix progress bars in tutobooks.

hertschuh:
Added better detection of blocks that are progress bars.

Added better handling of progress bars to only display the final state of each progress bar.


abheesht17 (approved):

mattdangerw (approved):
Thanks!

PR #2102, opened 2025-05-08
Add `pip install keras-rs` to all examples that need it.

hertschuh:
This is needed for the colabs to work.


abheesht17 (approved):
Thanks!

PR #2104, opened 2025-05-18
Fix DCN

abheesht17:


hertschuh (changes requested):
Wait, you forgot to add the modified `.py` file it looks like.

abheesht17:
> Wait, you forgot to add the modified `.py` file it looks like.

This is the modified file, right: https://github.com/keras-team/keras-io/blob/4098355ad09e11d8d93b0304dbec7befdfc2fd3a/examples/keras_rs/dcn.py?

hertschuh (approved):

PR #2113, opened 2025-05-29
Corrections for gradient centralization example.

albertoesmp:
The gradient centralization example used the model trained without gradient centralization (NGC) when applying gradient centralization (GC). Thus, the results were better because the GC model wasn't trained from scratch but started from the trained NGC model. This fix uses independent models for GC and NGC and thus yields a fair comparison between both strategies.


JyotinderSingh:
@albertoesmp Thank you for this fix, could you please resolve the CLA issue mentioned in the comment above?

hertschuh (approved):

PR #2114, opened 2025-05-30
Fix keras_rs examples in colab.

hertschuh:
`import keras_rs` with the JAX backend requires `jax-tpu-embedding` to be installed.

Also fixed rank issue in deep recommender example.


hertschuh:
@cantonios

abheesht17 (approved):

PR #2122, opened 2025-06-09
Use `keras-rs-nightly` for now.

hertschuh:
`DistributedEmbedding` as an issue in the docstring, which was fixed here: https://github.com/keras-team/keras-rs/pull/118 However, it is not picked up by keras.io because the pip wheel was not released after the fix.

The issue only affects keras.io, so instead of creating a release just for this, we're using `keras-rs-nightly` for now.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -16,4 +16,4 @@ keras-nlp
 keras-tuner
 tf-keras
 keras-hub
-keras-rs
+keras-rs-nightly


sachinprasadhs:
`keras-rs-nightly` doesn't seem to have release since April 30

hertschuh:
@sachinprasadhs 

> `keras-rs-nightly` doesn't seem to have release since April 30

Can you approve: https://github.com/keras-team/keras-rs/pull/121 ?

sachinprasadhs (approved):

PR #2128, opened 2025-06-19
Add image & numpy layers description

shashaka:
I have updated api_master.py for image augmentation and numPy ops.


sachinprasadhs (approved):
LGTM

PR #2129, opened 2025-06-23
Add Docs For Muon Optimizer

dhantule:
I have updated the `api_master.py` file to generate documentation for Muon optimizer which was introduced in this PR [#21037](https://github.com/keras-team/keras/pull/21037)


scripts/api_master.py
```diff
@@ -1491,6 +1491,11 @@
                         },
                     ],
                 },
+                {
+                  "path": "muon",
+                  "title": "Muon",
+                  "generate": ["keras.optimizers.Muon"],
+                }
```
sachinprasadhs: add comma after }

PR #2131, opened 2025-06-26
Added DLRM notebook and generated python script

kharshith-k:
Added DLRM.ipynb and generated dlrm.py


examples/keras_rs/dlrm.py
```diff
@@ -0,0 +1,406 @@
...
+"""
+## Introduction
+
+This tutorial demonstrates how to use Deep Learning Recommendation Model (DLRM) to
+effectively
+learn relationship between movies and ratings using DotInteraction
+
+The Deep Learning Recommendation Model (DLRM) architecture is designed for more effective
+learning of explicit and bounded-degree features. It comprises three main
+components: an input layer (typically an embedding layer), a dot interaction layer for
```
abheesht17: The input layer contains multiple embedding layers for every categorical feature, right?

examples/keras_rs/dlrm.py
```diff
@@ -0,0 +1,406 @@
+"""
+Title: Ranking with Deep Learning Recommendation Model using Dot Interaction
```
abheesht17: Let's remove "Dot Interaction". That's a part of DLRM anyway.

examples/keras_rs/dlrm.py
```diff
@@ -0,0 +1,406 @@
...
+"""
+
+"""
+## Introduction
+
+This tutorial demonstrates how to use Deep Learning Recommendation Model (DLRM) to
+effectively
+learn relationship between movies and ratings using DotInteraction
+
+The Deep Learning Recommendation Model (DLRM) architecture is designed for more effective
```
abheesht17: Please add the paper as a reference: https://arxiv.org/abs/1906.00091.

examples/keras_rs/dlrm.py
```diff
@@ -0,0 +1,406 @@
...
+learning of explicit and bounded-degree features. It comprises three main
+components: an input layer (typically an embedding layer), a dot interaction layer for
+modeling explicit feature interactions, and a deep network for capturing implicit
+interactions.
```
abheesht17: I think we should edit this a little. DLRM has the following components:

- Embedding layers for categorical features;
- A "top" MLP, through which dense features are passed;
- A dot interaction layer, which computes feature interactions between all embeddings and the output of the MLP on the dense features;
- Bottom MLP, through which everything from the above layers is passed.

Of course, you should rephrase what I've written above, it's rough :). Also, we should add a diagram of the architecture. You can see section 2.2 in the paper: https://arxiv.org/pdf/1906.00091.

examples/keras_rs/dlrm.py
```diff
@@ -0,0 +1,406 @@
+"""
+Title: Ranking with Deep Learning Recommendation Model using Dot Interaction
+Author: Harshith Kulkarni
+Date created: 2025/06/02
+Last modified: 2025/06/20
+Description: Rank movies with DLRM using Dot Interaction
```
abheesht17: Probably better to write "with KerasRS" instead of "using Dot Interaction"

examples/keras_rs/dlrm.py
```diff
@@ -0,0 +1,406 @@
...
+
+Now that we know a little bit about DLRM, let's start writing some code. We will
+first train a DLRM on a real-world dataset, and demonstrate that the model has indeed
+learnt important feature crosses.
+
+Let's set the backend to JAX, and get our imports sorted.
+"""
+
+"""shell
+!pip install  jax_tpu_embedding keras-rs
```
abheesht17: Is jax-tpu-embedding needed?

examples/keras_rs/dlrm.py
```diff
@@ -0,0 +1,406 @@
...
+        # Layers.
+
+        self.embedding_layers = []
+        for feature_name, vocabulary in vocabularies.items():
+            self.embedding_layers.append(
+                keras.layers.Embedding(
+                    input_dim=len(vocabulary) + 1,
+                    output_dim=embedding_dim,
+                )
+            )
```
abheesht17: One of the highlights of DLRM is that it can process both categorical and dense features. We should use some dense features present in the MovieLens dataset. Do we have any such features?

examples/keras_rs/dlrm.py
```diff
@@ -0,0 +1,406 @@
...
+products between different feature embeddings.
+
+The deep network is a standard feedforward multilayer perceptron
+(MLP). These two networks are then combined to form the DLRM.
+
+Now that we know a little bit about DLRM, let's start writing some code. We will
+first train a DLRM on a real-world dataset, and demonstrate that the model has indeed
+learnt important feature crosses.
+
+Let's set the backend to JAX, and get our imports sorted.
```
abheesht17: The introduction should contain some salient features of DLRM, such as its ability to handle both categorical and dense inputs, for example. Another good thing to talk about is how it is different from DCN (see section 2.3).

Feel free to borrow stuff from the paper.

abheesht17 (changes requested):
Thanks for the PR! Left some comments, let me know what you think :).

Here are some notes for posterity (we don't need to implement these now in this PR):

- We should, in the future, also think about the specific case of how how DLRM implements distributed training. The embeddings are sharded across devices, whereas the MLP layers are DDP'd.
- The paper mentions that they use `torch.nn.EmbeddingBag` instead of `torch.nn.Embedding`. Apparently, `torch.nn.EmbeddingBag` is more efficient than `torch.nn.Embedding` followed by some aggregation op. We should think about this at some later point (maybe, `keras_rs.embeddings.DistributedEmbedding` solves this). Anyway, we can worry about this later!

examples/keras_rs/dlrm.py
kharshith-k: Yes, the input layers contains multiple embedding layers, typically an embedding layer for each feature. I'll update the markdown regarding the same

examples/keras_rs/dlrm.py
kharshith-k: Sure!

examples/keras_rs/dlrm.py
kharshith-k: Sure, I'll add the paper as a reference.

examples/keras_rs/dlrm.py
kharshith-k: Sure, I'll edit the markdown content according to the layers defined in the code. However in current implementation, the "top" MLP layers are not there, the outputs from embedding layers are directly passed to dot interaction layer. Also I'll add architecture diagram from the paper.

examples/keras_rs/dlrm.py
kharshith-k: Sure, I'll replace "using Dot Interaction" with "with KerasRS"

examples/keras_rs/dlrm.py
kharshith-k: The movielens 100k-ratings dataset mostly contains categorical features, However it also has user age feature but its bucketized, which transforms it to be a categorical feature.

examples/keras_rs/dlrm.py
kharshith-k: Sure, I'll add some salient features that DLRM excels at by referring the research paper. Also I'll mention few points about how it is different from DCN.

examples/keras_rs/dlrm.py
kharshith-k: No, its not needed, looks like I had added that line when I was implementing the code and was experimenting with TPU runtime on colab then forgot to remove. I'm removing that line

abheesht17:
@kharshith-k - let me know when this is ready for another round of review. Thanks!

.DS_Store
abheesht17, line 1: Remove this file

examples/.DS_Store
abheesht17, line 1: Please remove this file. You can also add these files to .gitignore, BTW, to avoid these getting added by accident.

examples/keras_rs/.DS_Store
abheesht17, line 1: Please remove this file. You can also add these files to `.gitignore`, BTW, to avoid these getting added by accident.

examples/keras_rs/img/.DS_Store
abheesht17, line 1: Please remove

examples/keras_rs/ipynb/dlrm.ipynb
abheesht17, line 1: There are two copies of this notebook. Let's delete one of them

`examples/keras_rs/ipynb/DLRM.ipynb`
`examples/keras_rs/ipynb/dlrm.ipynb`

kharshith-k:
> @kharshith-k - let me know when this is ready for another round of review. Thanks!
@abheesht17 - The changes are ready for review. Please let me know if any more changes would be required. Thanks!

examples/keras_rs/dlrm.py
abheesht17: Hmmm, doesn't it have `raw_user_age` as a feature? The reason I'm insisting on this is because the two towers for dense and categorical features is a salient part of DLRM. And do you think we can use normalised timestamp as a feature? Will that help/does that make sense?

abheesht17 (commented):
Thanks! Great work, this looks good to me, overall. Just one major comment on using dense features (it isn't DLRM without the two blocks for dense and categorical features)

examples/keras_rs/dlrm.py
kharshith-k: Sure! like you mentioned, two towers for dense and categorical features is a salient part of DLRM. I'll have another look at the code and see if any dense features can be used. Timestamp looks like a suitable dense feature. I'll modify the code and try using that

abheesht17: And we can remove `user_bucketised_age` and use `raw_user_age`, maybe

kharshith-k: Sure! Will try that

kharshith-k:
Hi @abheesht17, The changes are ready for review. I've used raw_user_age and timestamp for continuous features. Please let me know if any more changes would be required. Thanks!

examples/keras_rs/dlrm.py
```diff
@@ -0,0 +1,480 @@
+"""
+Title: Ranking with Deep Learning Recommendation Model
+Author: Harshith Kulkarni
+Date created: 2025/06/02
+Last modified: 2025/08/18
+Description: Rank movies with DLRM with KerasRS
```
abheesht17: `Rank movies with DLRM with KerasRS` --> `Rank movies with DLRM using KerasRS`, maybe?

examples/keras_rs/dlrm.py
```diff
@@ -0,0 +1,480 @@
...
+
+This tutorial demonstrates how to use the Deep Learning Recommendation Model (DLRM) to
+effectively learn the relationships between items and user preferences using a
+dot-product interaction mechanism. For more details, please refer to the
+[DLRM](https://arxiv.org/pdf/1906.00091) paper.
+
+DLRM is designed to excel at capturing explicit, bounded-degree feature interactions and
+is particularly effective at processing both categorical and continuous (sparse/dense)
+input features. The architecture consists of three main components: dedicated input
+layers to handle diverse features (typically embedding layers for categorical features),
```
abheesht17, line 21: Let's remove this: typically embedding layers for categorical features

examples/keras_rs/dlrm.py
```diff
@@ -0,0 +1,480 @@
...
+
+
+Now that we have a foundational understanding of DLRM's architecture and key
+characteristics, let's dive into the code. We will train a DLRM on a real-world dataset
+to demonstrate its capability to learn meaningful feature interactions. Let's begin by
+setting the backend to JAX and organizing our imports.
+"""
+
+"""shell
+!pip install keras-rs -q
```
abheesht17: `!pip install keras-rs -q` --> `!pip install -q keras-rs `

examples/keras_rs/dlrm.py
```diff
@@ -0,0 +1,480 @@
...
+[basic ranking](/keras_rs/examples/basic_ranking/)
+tutorial. Let's load the dataset, and keep only the useful columns.
+"""
+
+ratings_ds = tfds.load(
+    "movielens/100k-ratings",
+    split="train",
+    download_and_prepare_kwargs={
+        "download_config": tfds.download.DownloadConfig(
+            verify_ssl=False,
```
abheesht17: Do we need this?

examples/keras_rs/dlrm.py
```diff
@@ -0,0 +1,480 @@
...
+    }
+)
+
+"""
+Let's normalize all the continuous features, so that we can use that for the MLP layers.
+"""
+
+normalization_layers = {}
+for feature_name in MOVIELENS_CONFIG["continuous_features"]:
+    normalization_layers[feature_name] = keras.layers.Normalization(axis=-1)
```
abheesht17, line 272: Sorry to be a pain here, but could you extract the hour of day, hour of week from the timestamp, and map it to a unit circle using sin, cos? That's a better way of encoding timestamp

examples/keras_rs/dlrm.py
```diff
@@ -0,0 +1,480 @@
...
+        # Layers for continuous features.
+        self.continuous_dense_layers = {}
+        for feature_name in MOVIELENS_CONFIG["continuous_features"]:
+            # Use a small MLP to process continuous features
+            self.continuous_dense_layers[feature_name] = keras.Sequential(
+                [
+                    keras.layers.Dense(mlp_dim, activation="relu"),
+                    keras.layers.Dense(mlp_dim),
+                ]
+            )
```
abheesht17: Generally, we concatenate all dense features together, and use just one MLP. Let's do that, instead of creating separate MLPs from every dense feature.

abheesht17 (commented):
Left a few comments

examples/keras_rs/dlrm.py
kharshith-k: Actually when I tried to download the dataset the other day, I was facing "server SSL certificate expired" error. So, as a work around I had to use this configuration. Should I keep it or remove it?

examples/keras_rs/dlrm.py
kharshith-k: Sure not a problem! I will do the changes accordingly

examples/keras_rs/dlrm.py
kharshith-k: Sure!

kharshith-k:
Hello @abheesht17 I've modified the DLRM code as per your suggestions. I've encoded the timestamp feature by splitting it into hour_of_day_sin, hour_of_day_cos, hour_of_week_sin and hour_of_week_cos using sin and cos functions in the preprocessing. Please let me know if its correct or needs any corrections. 

Thanks!

examples/keras_rs/dlrm.py
```diff
@@ -0,0 +1,498 @@
...
+            + MOVIELENS_CONFIG["categorical_str_features"]
+        ):
+            vocab_size = len(vocabularies[feature_name]) + 1  # +1 for OOV token
+            self.embedding_layers[feature_name] = keras.layers.Embedding(
+                input_dim=vocab_size,
+                output_dim=embedding_dim,
+            )
+
+        # A single MLP for all continuous features.
+        self.continuous_mlp = keras.Sequential(
```
abheesht17: Let's stick to what the original DLRM model uses. self.bottom_mlp?

examples/keras_rs/dlrm.py
```diff
@@ -0,0 +1,498 @@
...
+            [
+                keras.layers.Dense(mlp_dim, activation="relu"),
+                keras.layers.Dense(embedding_dim),  # Output must match embedding_dim
+            ]
+        )
+
+        self.dot_layer = keras_rs.layers.DotInteraction()
+
+        self.dense_layers = []
```
abheesht17: We can call this self.top_mlp

examples/keras_rs/dlrm.py
abheesht17: I think this was a temporary failure. Does it work without passing `verify_ssl=False` now?

examples/keras_rs/dlrm.py
kharshith-k: Sure!

examples/keras_rs/dlrm.py
kharshith-k: Sure!

examples/keras_rs/dlrm.py
kharshith-k: Yes, its working now and I've removed it

kharshith-k:
> This LGTM! Just one comment, please address it, re-generate the ipynb and md files, and we can go ahead and merge this. Thanks for all the work!

Hi @abheesht17, I've re-generated the ipynb and md files, can you please check and let me know if this can be merged or needs any more changes?

Thanks for constant support!

examples/keras_rs/dlrm.py
abheesht17: This is still self.dense_layers?

abheesht17: My bad, I was seeing an older version

Issue #1256, opened 2023-02-25
Misconception of the transformer_encoder in notebook https://github.com/keras-team/keras-io/blob/master/examples/timeseries/timeseries_classification_transformer.py


YacineKaci:
The normalization layer should be applied after the residual connection, not before. At least this is how it is done in the Transformer community.
So the code:
.
.
    x = layers.LayerNormalization(epsilon=1e-6)(x)
    res = x + inputs
.
.
    x = layers.LayerNormalization(epsilon=1e-6)(x)
    return x + res

Should be replaced by:
.
.
    x = x + inputs
    res  = layers.LayerNormalization(epsilon=1e-6)(x)
.
.
    x = x + res
    return layers.LayerNormalization(epsilon=1e-6)(x)

Kind regards

sonali-kumari1:
Hi @YacineKaci - 

Thanks for reporting this issue. While the [original Transformer paper](https://arxiv.org/abs/1706.03762) applies LayerNorm after the residual connection (Post-LN), the [Keras Transformer example](https://keras.io/examples/timeseries/timeseries_classification_transformer/) applies LayerNorm before the residual (Pre-LN) for a very specific reason: to improve training stability.

As explained in [this blog](https://sh-tsang.medium.com/review-pre-ln-transformer-on-layer-normalization-in-the-transformer-architecture-b6c91a89e9ab), the originally designed Post-LN Transformer can lead to large gradients near the output layer, often requiring a warm-up phase to avoid instability. On the other hand, the Pre-LN Transformer leads to better-behaved gradients and simplifies the training process.

Given these advantages, we'll maintain the Pre-LN approach in the [Keras Transformer example](https://keras.io/examples/timeseries/timeseries_classification_transformer/). Please feel free to close this issue if it resolves your concerns.

PR #2144, opened 2025-07-15
Fix: Transformer encoder - apply LayerNorm after residual

sonali-kumari1:
This PR fixes the Transformer encoder to properly apply Layer Normalization after the residual connections, in accordance with the specification from the [Attention is All You Need" (Vaswani et al., 2017)](https://arxiv.org/abs/1706.03762) paper. 

Previously, Layer Normalization was incorrectly applied before the residual connection, which violated the design outlined in Section 3.1 of "Attention is All You Need," where it states: "the output of each sub-layer is`LayerNorm(x + Sublayer(x))`". 

This update ensures that Layer Normalization is applied after the residual addition in both the attention and feed-forward sub-layers. 

Fixes : [#1256 ](https://github.com/keras-team/keras-io/issues/1256)

### DIFF
--- a/examples/timeseries/timeseries_classification_transformer.py
+++ b/examples/timeseries/timeseries_classification_transformer.py
@@ -75,15 +75,16 @@ def transformer_encoder(inputs, head_size, num_heads, ff_dim, dropout=0):
         key_dim=head_size, num_heads=num_heads, dropout=dropout
     )(inputs, inputs)
     x = layers.Dropout(dropout)(x)
-    x = layers.LayerNormalization(epsilon=1e-6)(x)
     res = x + inputs
+    x = layers.LayerNormalization(epsilon=1e-6)(res) 
 
     # Feed Forward Part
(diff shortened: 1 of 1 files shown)


sachinprasadhs:
There is no issue with the existing implementation.

The implementation in https://keras.io/examples/timeseries/timeseries_classification_transformer is also correct, it is an improvement to the original paper by having Layer Normalization inside the block so that the training will be stable.

Here is a blog explaining the difference between both the approaches https://sh-tsang.medium.com/review-pre-ln-transformer-on-layer-normalization-in-the-transformer-architecture-b6c91a89e9ab

May be you can just add a note explaining this variation.

sachinprasadhs:
Thanks, could you please make the changes to .md and .ipynb files as well, you don't have to render all the files, you can just make manual changes to those files.

sachinprasadhs:
Could you please run black format to the changed files

sachinprasadhs (approved):

PR #2152, opened 2025-07-30
highlights QAT in guide title

JyotinderSingh:


divyashreepathihalli:
@JyotinderSingh can you check the failing tests?

JyotinderSingh:
On a closer look, the guide doesn't implement true QAT. The "mixed precision" title is accurate. This change should not be required.

PR #2174, opened 2025-09-03
Add DistributedEmbedding example for TPU on TensorFlow.

hertschuh:
This was run on a cloud TPU v6e-1.


examples/keras_rs/distributed_embedding_tf.py
```diff
@@ -0,0 +1,315 @@
...
+    optimizer="adam",
+    placement="sparsecore",
+)
+
+FEATURE_CONFIGS = {
+    "movie_id": keras_rs.layers.FeatureConfig(
+        name="movie",
+        table=movie_table,
+        input_shape=(BATCH_SIZE,),
+        output_shape=(BATCH_SIZE, EMBEDDING_DIMENSION),
```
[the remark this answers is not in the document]
hertschuh: Comment was incorrect, fixed comment.

examples/keras_rs/distributed_embedding_tf.py
```diff
@@ -0,0 +1,315 @@
...
+    ),
+    "user_id": keras_rs.layers.FeatureConfig(
+        name="user",
+        table=user_table,
+        input_shape=(BATCH_SIZE,),
+        output_shape=(BATCH_SIZE, EMBEDDING_DIMENSION),
```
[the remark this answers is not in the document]
hertschuh: Comment was incorrect, fixed comment.

examples/keras_rs/distributed_embedding_tf.py
```diff
@@ -0,0 +1,315 @@
...
+                keras.layers.Dense(1),
+            ]
+        )
+
+    def call(self, features):
+        # Embedding lookup. Outputs have the same structure as the inputs.
+        embedding = self.embedding_layer(features)
+        return self.ratings(
+            keras.ops.concatenate(
+                [embedding["user_id"], embedding["movie_id"]],
```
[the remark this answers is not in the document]
hertschuh: What count is the key in the dict, feature name is just for information.

examples/keras_rs/distributed_embedding_tf.py
```diff
@@ -0,0 +1,315 @@
...
+use of `keras_rs.layers.DistributedEmbedding`, which makes use of SparseCore on
+TPU. This is the TensorFlow version of the tutorial. It needs to be run on TPU
+v5p or v6e.
+
+Let's begin by installing the necessary libraries. Note that we need
+`tensorflow-tpu` version 2.19. We'll also install `keras-rs`.
+"""
+
+"""shell
+pip install -U -q tensorflow-tpu==2.19.1
```
abheesht17, line 27: [knowledge] I know we've chatted about this offline a bit, but do you know if we intend this to work with 2.20+ in the future?

examples/keras_rs/distributed_embedding_tf.py
```diff
@@ -0,0 +1,315 @@
...
+pip install -q keras-rs
```
abheesht17, line 28: I think you'd mentioned this before: should we add the optional square bracket thing to KerasRS setup files, like so: `pip install -q keras-rs[tpu]`? Or `pip install -q keras-rs[dist-emb-tpu]` or something?

examples/keras_rs/distributed_embedding_tf.py
```diff
@@ -0,0 +1,315 @@
...
+"""
+
+"""
+We're using the PJRT version of the runtime for TensorFlow. We're also enabling
+the MLIR bridge. This requires setting a few flags before importing tensorflow.
```
abheesht17: [nit] `tensorflow` --> `TensorFlow`

examples/keras_rs/distributed_embedding_tf.py
```diff
@@ -0,0 +1,315 @@
...
+"""
+
+os.environ["KERAS_BACKEND"] = "tensorflow"
+
+import keras
+import keras_rs
+import tensorflow_datasets as tfds
+
+"""
+## Creating a TPUStrategy
```
abheesht17: [nit] TPUStrategy inside backticks?

examples/keras_rs/distributed_embedding_tf.py
```diff
@@ -0,0 +1,315 @@
...
+import libtpu
+
+os.environ["PJRT_DEVICE"] = "TPU"
+os.environ["NEXT_PLUGGABLE_DEVICE_USE_C_API"] = "true"
+os.environ["TF_PLUGGABLE_DEVICE_LIBRARY_PATH"] = libtpu.get_library_path()
+os.environ["TF_XLA_FLAGS"] = (
+    "--tf_mlir_enable_mlir_bridge=true "
+    "--tf_mlir_enable_convert_control_to_data_outputs_pass=true "
+    "--tf_mlir_enable_merge_control_flow_pass=true"
+)
```
abheesht17, line 46: [knowledge] These same settings work on GCP as well?

examples/keras_rs/distributed_embedding_tf.py
```diff
@@ -0,0 +1,315 @@
...
+
+import tensorflow as tf
+
+"""
+We're now ready to import `keras` and `keras-rs`. But we need to set the
+backend to TensorFlow.
```
abheesht17: [nit] Should we just say that "We now set the Keras backend to TensorFlow and import the necessary libraries". Because, we're importing TFDS as well

examples/keras_rs/distributed_embedding_tf.py
```diff
@@ -0,0 +1,315 @@
...
+
+To run TensorFlow on TPU, you need to use a `tf.distribute.TPUStrategy` to
+handle the distribution of the model.
+
+Note that the core of the model is replicated across TPU instances. Only the
+embedding tables handled by `DistributedEmbedding` are sharded across the
+SparseCore chips of all the available TPUs.
+"""
+
+resolver = tf.distribute.cluster_resolver.TPUClusterResolver(tpu="local")
```
abheesht17, line 76: [knowledge] I'm assuming this won't work in the multi-host case, right? Since `tpu = "local"`

examples/keras_rs/distributed_embedding_tf.py
```diff
@@ -0,0 +1,315 @@
...
+topology = tf.tpu.experimental.initialize_tpu_system(resolver)
+tpu_metadata = resolver.get_tpu_system_metadata()
+
+device_assignment = tf.tpu.experimental.DeviceAssignment.build(
+    topology, num_replicas=tpu_metadata.num_cores
+)
+strategy = tf.distribute.TPUStrategy(
+    resolver, experimental_device_assignment=device_assignment
+)
```
abheesht17, line 85: Curious why we have to do this instead of using `MirroredStrategy`? Maybe, we can add a note here for the reader as to why this is necessary? What do you think?

examples/keras_rs/distributed_embedding_tf.py
```diff
@@ -0,0 +1,315 @@
...
+We need to make sure the batch size is a multiple of the number of TPUs.
+"""
+
+PER_REPLICA_BATCH_SIZE = 256
+BATCH_SIZE = PER_REPLICA_BATCH_SIZE * strategy.num_replicas_in_sync
+
+"""
+## Preparing the dataset
+
+We're going to use the same Movielens data. The ratings are the objectives we
```
abheesht17: [nit] Movielens --> MovieLens

examples/keras_rs/distributed_embedding_tf.py
hertschuh: No, 2.19 is the last version of TensorFlow that supports TPUs (Monica linked a doc about this).

examples/keras_rs/distributed_embedding_tf.py
hertschuh: Yes, I was able to get this to work both in Colab and GCP.

examples/keras_rs/distributed_embedding_tf.py
hertschuh: Correct. Is it just a question of passing the name of the TPU cluster? If so, I can add a variable and a comment explaining how to do it. But I haven't tested it.

examples/keras_rs/distributed_embedding_tf.py
hertschuh: Well, that's what the text lines 64-69 are for. If you're on TPU, you have to use a `TPUStrategy`.

https://www.tensorflow.org/api_docs/python/tf/distribute/MirroredStrategy

> This strategy is typically used for training on one machine with multiple GPUs. For TPUs, use [tf.distribute.TPUStrategy](https://www.tensorflow.org/api_docs/python/tf/distribute/TPUStrategy)

I could add a link to the TensorFlow documentation.

examples/keras_rs/distributed_embedding_tf.py
hertschuh: That's a good question, we should experiment with this. Part of the complication is the combinatorial of backends and hardware. One other issue I have faced is that some packages clash in the version they want for their dependencies (`protobuf` or `keras` for instance).

I think we would have to do:
- `keras-rs[tf-tpu]` (adds `tensorflow-tpu==2.19.1`)
- `keras-rs[jax-tpu]` (adds `jax-tpu-embedding` and `jax[tpu]`)

examples/keras_rs/distributed_embedding_tf.py
abheesht17: Yeah, I think you can pass the GCP zone, project, etc. too. I just asked this question for knowledge. We don't have to add since we haven't tested it yet. What do you think?

examples/keras_rs/distributed_embedding_tf.py
abheesht17: > I could add a link to the TensorFlow documentation.

Yeah, let's add it?

hertschuh: Done

examples/keras_rs/distributed_embedding_tf.py
hertschuh: I can't get my hands on a multi-host v6e right now. We can submit as-is and revisit.

PR #2179, opened 2025-09-17
DistributedEmbedding for JAX example needs more recent JAX version.

hertschuh:
The preinstalled version of JAX on colab no longer works with the latest version of the `jax-tpu-embedding`.


abheesht17 (approved):
Thanks!

PR #2180, opened 2025-09-19
Update meta description tag to contain the specific title of the page.

hertschuh:
Every page had a fixed `<meta name="description" content="Keras documentation">` tag. It now contains the title of the page, which is more specific because it has the API name for instance.

Google search in particular looks at this tag, and this tag only, for indexing [Documentation](https://developers.google.com/search/docs/crawling-indexing/special-tags#meta-tags). The hope is that it will improve search results.

This also makes it consistent with other meta tags like `<meta property="og:title" content="...">`.

### DIFF
--- a/theme/base.html
+++ b/theme/base.html
@@ -5,7 +5,7 @@
 
   <meta charset="utf-8">
   <meta name="viewport" content="width=device-width, initial-scale=1">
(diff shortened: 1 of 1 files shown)


abheesht17 (approved):
Nice, thanks!

PR #2181, opened 2025-09-29
Fix Markdown Rendering for Handwriting Recognition Example

dhantule:
The Handwriting Recognition [Example](https://keras.io/examples/vision/handwriting_recognition/#model) had rendering issues due to missing backticks which caused the final image and the Final Remarks section to display incorrectly.

This fix adds the missing backticks  to resolve those layout problems

image


sachinprasadhs (approved):
LGTM, Thanks

PR #2187, opened 2025-10-07
migrate retinanet example to keras 3

Shi-pra-19:
This commit updates the RetinaNet example to be compatible with Keras 3. The following changes were made:
- Refactored the code to align with Keras 3's  API.
- Adjusted imports, model initialization, training, and callbacks as required by the new version of Keras.
This migration ensures that the example works seamlessly with the latest version of Keras.

This is my first PR to the repo, and I’m excited to contribute!

### DIFF
--- a/examples/vision/retinanet.py
+++ b/examples/vision/retinanet.py
@@ -29,12 +29,14 @@ the problem of the extreme foreground-background class imbalance.
 
 
 import os
+import glob
 import re
 import zipfile
 
 import numpy as np
 import tensorflow as tf
-from tensorflow import keras
+
+import keras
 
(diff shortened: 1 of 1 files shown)


Shi-pra-19:
Fixed the following - 
- Corrected the `prior_probability` typo.
- Added `from_logits=True`
- Kept `keras.optimizers.SGD`  instead of `keras.optimizers.sgd` (avoids `Attribute Error`).
- Pre-trained weights are not loaded because they are not in the format  required by Keras.

Shi-pra-19:
Could you please review this PR? @sachinprasadhs

PR #2188, opened 2025-10-07
Updates to "Customizing fit() with JAX" guide.

hertschuh:
The starting point for these changes was to use `stateless_compute_loss` instead of `compute_loss` for consistency, since we state that everything in `train_step` should be stateless.
- This in turn required plumbing `metrics_variable` and `sample_weight`.
- This required adding `keras.utils.unpack_x_y_sample_weight`.

These changes make the guide slightly longer, but the code is more generic and closer to the Keras default implementation.

Also fixed issue where `test_step` was not returning the metrics correctly.


SamanehSaadat:
> The starting point for these changes was to use `stateless_compute_loss` instead of `compute_loss` for consistency, since we state that everything in `train_step` should be stateless.
> 
> * This in turn required plumbing `metrics_variable` and `sample_weight`.
> * This required adding `keras.utils.unpack_x_y_sample_weight`.
> 
> These changes make the guide slightly longer, but the code is more generic and closer to the Keras default implementation.
> 
> Also fixed issue where `test_step` was not returning the metrics correctly.

I personally think it's okay to sacrifice a little bit of consistency for simplicity in a guide. One other option could be just adding a paragraph at the end explaining when `stateless_compute_loss` should be used? But your call.

hertschuh:
Thanks for the review @SamanehSaadat !

> I personally think it's okay to sacrifice a little bit of consistency for simplicity in a guide. One other option could be just adding a paragraph at the end explaining when `stateless_compute_loss` should be used? But your call.

After thinking about it, I'm going to submit it like this. Yufeng's video now uses the code with `stateless_compute_loss`, and people copying this code as a starting point would have a hard time figuring it out what to add to support `stateless_compute_loss`.

PR #2201, opened 2025-10-21
Removing the Developer survey banner and nav bar link from the Keras.io landing page

buildwithsuhana:
This PR removes the 2025 developer survey banner and survey link in the navbar as the survey is now closed.

The landing page now looks like:

image


sachinprasadhs:
There is a conflict, which needs to be resolved

Issue #2176, opened 2025-09-05
Keras Example: English-to-Spanish translation with KerasHub  Got error when building in decode_sequences function


alrahimi:
### Issue Type

Bug

### Source

binary

### Keras Version

TensorFlow Version: 2.19.0 keras Version: 3.11.3 keras_hub Version: 0.21.1

### Custom Code

No

### OS Platform and Distribution

colab ubuntu 

### Python version

3.12.11

### GPU model and memory

T4

### Current Behavior?

Keras Example: English-to-Spanish translation with a sequence-to-sequence Transformer
Got error when building in decode_sequences(input_sentences):
AttributeError                            Traceback (most recent call last)
[/tmp/ipython-input-257581123.py](https://localhost:8080/#) in <cell line: 0>()
     37 for i in range(2):
     38     input_sentence = random.choice(test_eng_texts)
---> 39     translated = decode_sequences([input_sentence])
     40     translated = translated.numpy()[0].decode("utf-8")
     41     translated = (

1 frames
[/usr/local/lib/python3.12/dist-packages/tensorflow/python/framework/tensor.py](https://localhost:8080/#) in __getattr__(self, name)
    258         tf.experimental.numpy.experimental_enable_numpy_behavior()
    259       """)
--> 260     self.__getattribute__(name)
    261 
    262   @property

AttributeError: 'tensorflow.python.framework.ops.EagerTensor' object has no attribute 'to_tensor'

### Standalone code to reproduce the issue or tutorial link

```shell
Just try to build the example below  by clicking "view in colab" .  
https://keras.io/examples/nlp/neural_machine_translation_with_keras_hub/
I built it as it is and got error for PIP install so I chhged to  keras-hub==0.21.1
!pip install -q --upgrade rouge-score
  !pip install -q --upgrade keras-hub==0.21.1
  !pip install -q --upgrade keras  # Upgrade to Keras 3.

even if I ignore the error and use keras-hub==0.22.1 I still get the same error in the decode_sequences
```

### Relevant log output

```shell

```

PR #2203, opened 2025-10-23
Fixes to `neural_machine_translation_with_keras_hub` example.

hertschuh:
This example was no longer running beause of:
- breaking change in `get_file`.
- use of `to_tensor` on a non-ragged tensor, which was replaced with the cross-backend `convert_to_tensor(ragged=False)`.
- decoding of string that was no longer needed.

Fixes https://github.com/keras-team/keras-io/issues/2176

### DIFF
--- a/examples/nlp/neural_machine_translation_with_keras_hub.py
+++ b/examples/nlp/neural_machine_translation_with_keras_hub.py
@@ -46,7 +46,6 @@ Before we start implementing the pipeline, let's import all the libraries we nee
 """shell
 pip install -q --upgrade rouge-score
 pip install -q --upgrade keras-hub
-pip install -q --upgrade keras  # Upgrade to Keras 3.
 """
 
 import keras_hub
@@ -57,9 +56,6 @@ import keras
 from keras import ops
 
 import tensorflow.data as tf_data
-from tensorflow_text.tools.wordpiece_vocab import (
-    bert_vocab_from_dataset as bert_vocab,
-)
 
 """
 Let's also define our parameters/hyperparameters.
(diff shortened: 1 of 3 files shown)


examples/nlp/neural_machine_translation_with_keras_hub.py
```diff
@@ -46,7 +46,7 @@
 """shell
 pip install -q --upgrade rouge-score
 pip install -q --upgrade keras-hub
-pip install -q --upgrade keras  # Upgrade to Keras 3.
+pip install -q --upgrade keras
```
abheesht17: Do we need to install Keras in examples at all now? Colab has Keras 3.x. I think we can remove it?

examples/nlp/neural_machine_translation_with_keras_hub.py
```diff
@@ -87,7 +84,7 @@
     origin="http://storage.googleapis.com/download.tensorflow.org/data/spa-eng.zip",
     extract=True,
 )
-text_file = pathlib.Path(text_file).parent / "spa-eng" / "spa.txt"
+text_file = pathlib.Path(text_file).parent / "spa-eng_extracted" / "spa-eng" / "spa.txt"
```
abheesht17: [Unrelated, somewhat]

So, `keras.utils.get_file` adds `_extracted` on its own in newer versions of Keras? Any reason why this was done in the first place?

examples/nlp/neural_machine_translation_with_keras_hub.py
hertschuh: Removed

examples/nlp/neural_machine_translation_with_keras_hub.py
hertschuh: See https://github.com/keras-team/keras/commit/dcefb139863505d166dd1325066f329b3033d45a

I made a change which makes it a bit shorter.

PR #2204, opened 2025-10-24
Add Depth Anything,MobilenetV5,ParSeq, Qwen3Moe, SmollLM3 Hub model API details

sachinprasadhs:
- Depth Anything
- MobilenetV5
- ParSeq
- Qwen3Moe
- SmollLM3

### DIFF
--- a/scripts/docstrings.py
+++ b/scripts/docstrings.py
@@ -226,7 +226,6 @@ def get_default_value_for_repr(value):
         repr(value).startswith("<")  # <Foo object at 0x00000000>
         and hasattr(value, "__class__")  # it is an object
         and hasattr(value, "get_config")  # it is a Keras object
-        and not inspect.isclass(value)  # ensure it's an instance, not a class
     ):
         config = value.get_config()
         init_args = []  # The __init__ arguments to render


scripts/docstrings.py
```diff
@@ -217,10 +217,16 @@ def __repr__(self):
...
+    if inspect.isclass(value):
+        # Render classes as module.ClassName to produce a valid python
+        # dotted-name expression in the fake signature (black can parse it).
+        return ReprWrapper(value.__module__ + "." + value.__name__)
+
     if (
         repr(value).startswith("<")  # <Foo object at 0x00000000>
         and hasattr(value, "__class__")  # it is an object
         and hasattr(value, "get_config")  # it is a Keras object
+        and not inspect.isclass(value)  # ensure it's an instance, not a class
```
hertschuh: Since you're returning right above, this is no longer needed.

scripts/docstrings.py
```diff
@@ -217,10 +217,16 @@ def __repr__(self):
         # Render the function name instead
         return ReprWrapper(value.__name__)
 
+    if inspect.isclass(value):
+        # Render classes as module.ClassName to produce a valid python
+        # dotted-name expression in the fake signature (black can parse it).
+        return ReprWrapper(value.__module__ + "." + value.__name__)
```
hertschuh, line 223: Oh can you show how it looks like?

scripts/docstrings.py
sachinprasadhs: This was giving issue in https://github.com/keras-team/keras-hub/blob/master/keras_hub/src/models/mobilenetv5/mobilenetv5_backbone.py#L160, since `SqueezeAndExcite2D` is a class.

scripts/docstrings.py
sachinprasadhs: You mean the model page?

sachinprasadhs: It renders like the highlighted mark here https://screenshot.googleplex.com/AMv3DsmNTr46hYo and above commented link is the argument list.

scripts/docstrings.py
hertschuh: I assume you added this before you added line 220. But now that you have line 220, this condition is no longer possible.

If `value` is a class, you will have returned line 223, and this condition is no longer possible.

sachinprasadhs: Done

hertschuh (approved):
Thanks!

PR #2205, opened 2025-10-24
Cleanup templates/ dir

sachinprasadhs:
Ran the` python autogen.py make` command and noticed many of the tutorials and examples having outdated examples/not synced or having extra length output for epochs.


hertschuh:
What about the `ipynb` files? Were they all up to date?

sachinprasadhs:
Noticed that these .md files are actually present in https://github.com/keras-team/keras-io/tree/master/examples/keras_rs, why do we have these again in templates directory?

sachinprasadhs:
This is the PR which caused this https://github.com/keras-team/keras-io/pull/2083
Was it intentional or accidental?

hertschuh:
This was an accident, the files in the `templates` folder should come from `autogen` and should not be committed in the repo (with some exceptions like `index.md`). Can you delete them all?

sachinprasadhs:
Got it, deleted the files under `temples/ except` `index.md` files, also updated the `.gitignore` file.

PR #2206, opened 2025-10-28
migrate metric learning example to keras 3

Shi-pra-19:
Refactor loss function and tensor handling in metric learning for compatibility with Keras 3.


Shi-pra-19:
I added `run_eagerly=True` because, without it, the training raised an **"OperatorNotAllowedInGraphError"** caused by an internal operation in the TensorFlow Similarity library.

PR #2226, opened 2025-12-19
Add guide for writing a custom kernel

jeffcarp:
Walks users through how to define a custom kernel using Pallas (only works with the JAX backend).

### DIFF
--- a/guides/define_custom_kernel.py
+++ b/guides/define_custom_kernel.py
@@ -4,7 +4,7 @@ Author: [jeffcarp](https://www.jeffcarp.com/)
 Date created: 2025/12/18
 Last modified: 2025/12/18
 Description: Write high-performance custom Keras layers for TPUs and GPUs.
-Accelerator: v5e-1 TPU
+Accelerator: TPU
 """
 
 """
--- a/guides/md/define_custom_kernel.md
+++ b/guides/md/define_custom_kernel.md
@@ -5,6 +5,11 @@
 **Last modified:** 2025/12/18<br>
 **Description:** Write high-performance custom Keras layers for TPUs and GPUs.
 
+
+<img class="k-inline-icon" src="https://colab.research.google.com/img/colab_favicon.ico"/> [**View in Colab**](https://colab.research.google.com/github/keras-team/keras-io/blob/master/guides/ipynb/define_custom_kernel.ipynb)  <span class="k-dot">•</span><img class="k-inline-icon" src="https://github.com/favicon.ico"/> [**GitHub source**](https://github.com/keras-team/keras-io/blob/master/guides/define_custom_kernel.py)
+
+
+
 # How to Write a Custom TPU or GPU Kernel in Keras
 
 Keras has [many pre-made layers to choose from](/api/layers/), and the
@@ -39,10 +44,7 @@ First, make sure you're running the latest version of `libtpu`:
 !pip install --upgrade -q "jax[tpu]" -f https://storage.googleapis.com/jax-releases/libtpu_releases.html
 ```
 
-    [33mDEPRECATION: The HTML index page being used (https://storage.googleapis.com/jax-releases/libtpu_releases.html) is not a proper HTML 5 document. This is in violation of PEP 503 which requires these pages to be well-formed HTML 5 documents. Please reach out to the owners of this index page, and ask them to update this index page to a valid HTML 5 document. pip 22.2 will enforce this behaviour change. Discussion can be found at https://github.com/pypa/pip/issues/10825[0m[33m
-    [0m
-
-
+    
 ```python
 from functools import partial
 import os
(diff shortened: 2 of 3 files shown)


guides/define_custom_kernel.py
```diff
@@ -0,0 +1,412 @@
...
+
+**Notes**
+
+1. Pallas is only compatible with the JAX backend.
+2. If you're running in Colab and get an error about `Pallas TPU requires a libtpu
+version that's at most a month old`, run this line to update `libtpu`:
+
+```
+!pip install --upgrade -q "jax[tpu]" -f
+https://storage.googleapis.com/jax-releases/libtpu_releases.html
```
hertschuh: Don't wrap this line so that copy-paste works.

Also, is this needed with the current version of colab? If so, make it a shell block:

```
"""shell
pip install --upgrade -q "jax[tpu]" -f https://storage.googleapis.com/jax-releases/libtpu_releases.html
"""
```

guides/define_custom_kernel.py
```diff
@@ -0,0 +1,412 @@
...
+
+Keras has [many pre-made layers to choose from](/api/layers/), and the
+ability to easily [create your
+own](/guides/making_new_layers_and_models_via_subclassing/) if you can't
+find the exact one you need. However, if you need to customize memory and computation
+behavior at the hardware level, or you have a need for speed, you may be interested in
+writing your own custom kernel.
+
+This guide will explore how to write a custom kernel and add it to your
+Keras model. We will utilize **Pallas**, a library that lets you write
```
hertschuh, line 24: Can you very explicitly say (early) on which TPUs and GPUs this colab will work? In particular which ones from the colab free tier. I'm assuming none of the TPUs in the colab free tier would be supported. What about the GPUs in the colab free tier?

guides/define_custom_kernel.py
```diff
@@ -0,0 +1,412 @@
...
+import os
+
+os.environ["KERAS_BACKEND"] = "jax"
+
+import jax
+from jax.experimental import pallas as pl
+import jax.numpy as jnp
+import keras
+
+assert keras.backend.backend() == "jax", "Must use JAX for this guide."
```
hertschuh: Not needed because of line 51.

guides/define_custom_kernel.py
```diff
@@ -0,0 +1,412 @@
...
+
+from functools import partial
+import os
+
+os.environ["KERAS_BACKEND"] = "jax"
+
+import jax
+from jax.experimental import pallas as pl
+import jax.numpy as jnp
+import keras
```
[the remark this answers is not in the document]
jeffcarp: Done.

guides/define_custom_kernel.py
```diff
@@ -0,0 +1,412 @@
...
+benchmark(standard_layer, input_data, "Standard Keras (Matmul + ReLU)")
+benchmark(pallas_layer, input_data, "Pallas Fused (Matmul + ReLU)")
+
+
+```
+Benchmarking Matrix Size: 8192x8192
+------------------------------
+Standard Keras (Matmul + ReLU) Average Latency: 7.879 ms
+Pallas Fused (Matmul + ReLU) Average Latency: 2.107 ms
+```
```
hertschuh: You will get that "for free" once you actually render the guide

guides/define_custom_kernel.py
```diff
@@ -0,0 +1,412 @@
+"""
```
hertschuh, line 1: You also have to render the guide by running

```
cd scripts
python autogen.py add_guide define_custom_kernel
```
and add the other files to the PR. You'll also need to add this guide to the index:

https://github.com/keras-team/keras-io/blob/80982a533b4e69141deb849047b639303b54c6f1/scripts/guides_master.py#L43

guides/define_custom_kernel.py
jeffcarp: Sounds good - Colab has v5e1 TPUs which support this, added a note.

guides/define_custom_kernel.py
jeffcarp: The guide for writing guides says to submit a PR with the .py file only:
https://github.com/keras-team/keras-io/blob/80982a533b4e69141deb849047b639303b54c6f1/README.md?plain=1#L113-L114
Should we update it?

Added to guides index. Working on rendering - though it may require TPU to render?

guides/define_custom_kernel.py
jeffcarp: Deleted, thanks

guides/define_custom_kernel.py
jeffcarp: I was able to render on CPU by temporarily disabling outputs - I'll check out a TPU VM and use that to render, hopefully that will work.

hertschuh: Oh I didn't even realized that's the recommend flow.

Yes, you need a TPU VM to really render it. There is a way to do it in colab too, although it's a bit contrived.

hertschuh: I'm a bit confused about what the CPU rendering did because I don't see the output blocks.

jeffcarp: Re: CPU rendering - I manually disabled outputs just to test if everything else worked.

I was able to render with a TPU VM. I manually uploaded to Colab to verify, looks like everything worked?
https://colab.research.google.com/drive/16IrGN_oUG0OOs0uWq75LR9JWVhnWnQcJ?usp=sharing

guides/md/define_custom_kernel.md
```diff
@@ -0,0 +1,414 @@
...
+    avg_time = (end_time - start_time) / iterations * 1000  # convert to ms
+    print(f"{name} Average Latency: {avg_time:.3f} ms")
+
+
+# 2. Run Comparison
+print(f"Benchmarking Matrix Size: {N}x{N}\n" + "-" * 30)
+benchmark(standard_layer, input_data, "Standard Keras (Matmul + ReLU)")
+benchmark(pallas_layer, input_data, "Pallas Fused (Matmul + ReLU)")
+
+```
```
hertschuh, line 376: I don't understand. The outputs are missing in your rendered colab. They should appear in blocks like this: https://github.com/keras-team/keras-io/blob/master/examples/keras_rs/md/sequential_retrieval.md?plain=1#L311

(they are correctly appearing in the `ipynb` file though)

jeffcarp: My bad, I was manually running `nbconvert` on TPU and missed that. Switched to running autogen.py on TPU. That should fix it.

PR #2229, opened 2026-01-03
Fix: Replace manual unzip shell command with Python zipfile

jaytiwarihub:
**Description**
The current tutorial relies on a shell command (`unzip`) to extract the dataset. This fails on Windows environments where `unzip` is not installed by default, and it interrupts the flow for users who haven't manually configured the Kaggle CLI.

**Changes**
- Replaced the manual shell instructions with Python's built-in `zipfile` module.
- Added a check to verify if the dataset exists before attempting extraction.
- Updated the code to be cross-platform (works on Windows, Linux, and macOS).

**Verification**
- I tested the script locally.
- Verified that the zip file is correctly extracted to the `16000_pcm_speeches` directory automatically.


jaytiwarihub:
@googlebot I signed it!

jaytiwarihub:
@googlebot I signed it!

jaytiwarihub:
@googlebot I signed it!

jaytiwarihub:
@googlebot I signed it!

PR #2236, opened 2026-01-09
Fix: Unsafe path handling and directory pollution in Time Series examples

jaytiwarihub:
## Description
This PR fixes two path handling issues in the `examples/timeseries/` folder that cause fragility on Windows and cache directory pollution.

## Changes

1. **`timeseries_weather_forecasting.py`**
   - **Issue:** The script was using `ZipFile.extractall()` without a path, dumping dataset files into the user's current working directory instead of the Keras cache. It also used a relative path (`"jena_climate..."`) which fails if the script is run from a different directory.
   - **Fix:** Now extracts files directly to the `~/.keras/datasets/` cache directory using `os.path.dirname(zip_path)` and accesses them using an absolute path.

2. **`timeseries_traffic_forecasting.py`**
   - **Issue:** The script used `data_dir = data_dir.rstrip("PeMSD7_Full.zip")`. This is unsafe because `rstrip` removes *any* character from the set, potentially mangling the directory path if it ends with characters like `p`, `z`, `i`, etc.
   - **Fix:** Replaced with `os.path.dirname(zip_path)` for robust, cross-platform parent directory resolution.

## Test Plan
- Verified that datasets download and extract correctly to the Keras cache.
- Verified that file paths resolve correctly on Windows.

### DIFF
--- a/examples/timeseries/timeseries_traffic_forecasting.ipynb
+++ b/examples/timeseries/timeseries_traffic_forecasting.ipynb
@@ -1,0 +1,920 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {
+    "colab_type": "text"
+   },
+   "source": [
+    "# Traffic forecasting using graph neural networks and LSTM\n",
+    "\n",
+    "**Author:** [Arash Khodadadi](https://www.linkedin.com/in/arash-khodadadi-08a02490/)<br>\n",
+    "**Date created:** 2021/12/28<br>\n",
+    "**Last modified:** 2023/11/22<br>\n",
+    "**Description:** This example demonstrates how to do timeseries forecasting over graphs."
+   ]
+  },
+  {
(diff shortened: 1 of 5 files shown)


sachinprasadhs:
Thanks, the changes looks good, but let's not delete the files.
Like I have mentioned in the other PRs, you only make the changes to the .py files and for each file generate .ipynb and .md files by following the example here https://github.com/keras-team/keras-io/blob/master/README.md#previewing-a-new-example
Also, it's good practice to make the changes to the modified date field to let users know when the example was updated.

sachinprasadhs:
Resolve the conflicts and apply the suggestion for your others PRs as well.

jaytiwarihub:
Closing this one out! .... The git history got a bit tangled here with some accidental deletions, so I'm moving to a fresh branch to keep things clean
Moving this work to PR #2272

PR #2241, opened 2026-01-11
Fix: Use secure HTTPS and automatic extraction for MovieLens dataset

jaytiwarihub:
## Description
Updated the MovieLens dataset loader to use the secure `https` protocol (replacing `http`) and replaced manual `zipfile` extraction with the built-in `keras.utils.get_file(..., extract=True)` utility.

## Changes
1. Changed dataset URL from `http://` to `https://` for security.
2. Removed redundant `ZipFile` extraction code.
3. Enabled `extract=True` in `get_file` to handle caching and extraction automatically.

## Test Plan
- Verified that the `ml-latest-small.zip` file downloads and extracts correctly to `~/.keras/datasets/`.

### DIFF
--- a/examples/structured_data/collaborative_filtering_movielens.ipynb
+++ b/examples/structured_data/collaborative_filtering_movielens.ipynb
@@ -1,0 +1,379 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {
+    "colab_type": "text"
+   },
+   "source": [
+    "# Collaborative Filtering for Movie Recommendations\n",
+    "\n",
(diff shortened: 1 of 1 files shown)


examples/structured_data/collaborative_filtering_movielens.py
```diff
@@ -48,23 +48,16 @@
 
 # Download the actual data from http://files.grouplens.org/datasets/movielens/ml-latest-small.zip"
 # Use the ratings.csv file
+# FIX: Use HTTPS and let Keras handle extraction automatically
```
sachinprasadhs, line 51: You can remove this comment.

PR #2244, opened 2026-01-15
Fix: Support Sequential models in Grad-CAM example

jaytiwarihub:
This PR fixes a compatibility issue in examples/vision/grad_cam.py where the make_gradcam_heatmap function would fail for Sequential models.

Currently, the function uses model.inputs to construct the gradient model. While this works for Functional API models (like Xception in the example), it often raises errors or behaves unexpectedly with simple Sequential models that haven't explicitly built their graph with an InputLayer.

Changes:

Changed model.inputs to model.input in make_gradcam_heatmap.

This ensures the function works robustly for both Sequential and Functional models, as model.input is the standard way to access the single input tensor for image classifiers.

### DIFF
--- a/examples/vision/ipynb/grad_cam.ipynb
+++ b/examples/vision/ipynb/grad_cam.ipynb
@@ -327,7 +327,7 @@
   "accelerator": "GPU",
   "colab": {
    "collapsed_sections": [],
-   "name": "C:\\Users\\om777\\keras-io\\examples\\vision\\grad_cam",
+   "name": "grad_cam",
    "private_outputs": false,
    "provenance": [],
    "toc_visible": true


divyashreepathihalli:
A corresponding .md file will need to be generated. Instructions here - https://github.com/keras-team/keras-io?tab=readme-ov-file#adding-a-new-code-example

jaytiwarihub:
@divyashreepathihalli Thanks for the review! I have updated the grad_cam.md file as requested. I also included a small fix in scripts/tutobooks.py. The autogen script was crashing on Windows due to encoding errors (UnicodeDecodeError), so I added explicit encoding='utf-8' handling to fix it. Everything is syncing correctly now.

jaytiwarihub:
@sachinprasadhs My intention with this change was to make make_gradcam_heatmap more robust. Many users copy-paste this function to debug their own custom models, and if they use a Sequential model, the current implementation fails because of how inputs are handled.
However, I understand that for the purpose of this specific tutorial (which uses Xception), the extra logic might be unnecessary complexity.
Would you prefer I revert this specific input-handling change and keep the function strictly optimized for Functional/Single-Input models?

PR #2259, opened 2026-01-22
Remove Keras 2 guides from rendering

sachinprasadhs:
- **Remove Keras 2 Examples from Rendering**
This PR updates the documentation generation system to filter Keras 2 examples, ensuring that only Keras 3 content is displayed on keras.io. By adding a Keras_2 tag, legacy guides remain in the repository for future migration reference but will no longer be rendered on the website.
- **Moved all of the existing examples from others section to it's appropriate section.**

### DIFF
--- a/scripts/autogen.py
+++ b/scripts/autogen.py
@@ -469,16 +469,11 @@ class KerasIO:
                     md_content_lines = md_content.split("\n")
                     for i, line in enumerate(md_content_lines):
                         if "View in Colab" in line:
-                            md_content_lines.insert(
-                                i,
-                                f"<div class='example_version_banner keras_{version}'>ⓘ This example uses Keras {version}</div>",
-                            )
-                            # Add warning for Keras 2 examples
                             if version == 2:
-                                md_content_lines.insert(
-                                    i + 1,
-                                    "<div class='example_keras2_warning'><strong>Warning:</strong> This example uses Keras 2 and may not be compatible with the latest version of Keras. Please check out all of our <a href=\"https://keras.io/examples/\">Keras 3 examples here</a>.</div>",
-                                )
+                                banner_text = f"<div class='example_version_banner keras_{version}'>ⓘ This example uses Keras {version}. This example may not be compatible with the latest version of Keras. Please check out all of our <a href=\"https://keras.io/examples/\">Keras 3 examples here</a>.</div>"
+                            else:
+                                banner_text = f"<div class='example_version_banner keras_{version}'>ⓘ This example uses Keras {version}</div>"
+                            md_content_lines.insert(i, banner_text)
                             break
                     md_content = "\n".join(md_content_lines) + "\n"
                     with open(md_path, "w") as f:


scripts/autogen.py
```diff
@@ -92,7 +92,13 @@ def make_examples_master(self):
                         f.close()
                         assert title_line.startswith("Title: ")
                         title = title_line[len("Title: ") :]
-                        children.append({"path": example_path, "title": title.strip()})
+                        children.append(
+                            {
+                                "path": example_path,
+                                "title": title.strip(),
+                                "keras_3": True,
```
jeffcarp, line 100: Instead of separate keys for `keras_2` and `keras_3`, it feels cleaner to have a `keras_version` key that can be 2 or 3

jeffcarp (commented):
Thanks! Have we confirmed that all of the Keras 2 guides are not able to be updated to Keras 3? If we're not certain, it might be better to add a warning at the top of the Keras 2 guides until we can confirm. Also re: Fabien's point, I think we should keep rendering the pages, but maybe just remove them from the navigation?

sachinprasadhs:
Thanks for the review. 
Warning sounds good, does something like "**Warning: This example is legacy and may be removed in future updates. It is no longer guaranteed to function correctly. Please refer to the Keras 3 examples instead.**" works, right under where we have "This example uses Keras 2 like this https://keras.io/examples/vision/near_dup_search/?

When you mentioned remove from navigation, you mean to remove from here https://keras.io/examples/?
If we remove from examples_master.py for these examples, this also will remove from rendering the pages as per the current implementation.

scripts/autogen.py
sachinprasadhs: The reason I introduced `Keras_2` is, we already have a  `Keras_3` key, based on that the examples are loading the  information like this https://keras.io/examples/nlp/multimodal_entailment/

Assuming this `Keras_2` key is only temporary, till we figure out the fate of each examples, either to remove it or migrate it.

jeffcarp: Ah gotcha, if a guide is not `keras_3` does that mean 100% the time it's Keras 2? In that case do we even need a new key?

jeffcarp:
Sorry for the delay - ah I see, so it looks like we already automatically apply a warning to Keras 2 guides:
https://github.com/keras-team/keras-io/blob/f4ebbf62007ee5a840847a22b646742cdb84885f/scripts/autogen.py#L468

In that case, expanding on the warning sounds good. How about something like:

```python
if version == 2:
  md_content_lines.insert(
    ... (
        "Warning: This example uses Keras 2 and may not be compatible with the latest version of Keras. "
        "Please check out all of our <a href="https://keras.io/examples/">Keras 3 examples here</a>."
    )
```

> If we remove from examples_master.py for these examples, this also will remove from rendering the pages as per the current implementation.

Would it be possible to decouple this so the pages still render but they don't show up in the navigation? That sounds like a smoother step forward.

sachinprasadhs:
Added the warning message as suggested, and now it only renders keras 3 examples in landing page and in the navigation keras 2 examples will be loaded and it will show the warning.

Added the below screenshots for reference.

Screenshot 2026-01-30 at 12 36 19 AM

Keras 2 example with warning:
Screenshot 2026-01-30 at 12 40 47 AM

scripts/autogen.py
jeffcarp: Is this still necessary given the most recent changes? Logically it seems like we only need the `keras_2` key.

jeffcarp (commented):
Thanks Sachin! Looks good, just left a comment about potential simplification.

sachinprasadhs:
This block is necessary as a fallback mechanism. If a new example is added to keras.io but isn't explicitly updated in examples_master.py, this script ensures it is still pulled from the src directory and listed under the [Others](https://keras.io/examples/) category. By setting `Keras_3` to `True`, we ensure these new examples are correctly tagged as `Keras 3` by default, preventing the previous issue where they were incorrectly labeled as `Keras 2`."

scripts/autogen.py
```diff
@@ -467,6 +473,12 @@ def sync_tutobook_templates(self):
                                 i,
                                 f"<div class='example_version_banner keras_{version}'>ⓘ This example uses Keras {version}</div>",
                             )
+                            # Add warning for Keras 2 examples
+                            if version == 2:
+                                md_content_lines.insert(
+                                    i + 1,
+                                    "<div class='example_keras2_warning'><strong>Warning:</strong> This example uses Keras 2 and may not be compatible with the latest version of Keras. Please check out all of our <a href=\"https://keras.io/examples/\">Keras 3 examples here</a>.</div>",
```
jeffcarp: I just realized won't this be duplicative with the warning above? I.e. it will show:
`This example uses Keras 2`
`Warning: This example uses Keras 2 and`

Can you make it so that there's only one message for 3 and one message for 2?

jeffcarp (commented):
Got it - thanks for the clarification. Left one more comment

scripts/autogen.py
sachinprasadhs: There is no extra message for Keras 3, this warning message is only for Keras 2 tutorials.
Shall I rephrase the warning message like 

"Warning: This example is legacy and may be removed in future updates. It is no longer guaranteed to function correctly. Please refer to our <a href="https://keras.io/examples/">Keras 3 examples here"

jeffcarp: Right, but for Keras 2 there are still two separate messages, right? Can you restructure so there's only one message for both? Something like:

```
if version == 3:
  md_content_lines.insert( {keras 3 message} )
elif version == 2:
  md_content_lines.insert( {keras 2 message} )
```

sachinprasadhs: Just to clarify. 

Currently this is how we are rendering for Keras 3 example( Banner after description): https://keras.io/examples/vision/image_classification_from_scratch/

For Keras 2 example: 
https://keras.io/examples/vision/retinanet/ 

With the changes in this PR, for this banner, the only extra thing is getting added is a warning message(after Keras 2 banner) for Keras 2 examples like below image

image



So, if I'm understanding it correctly, you are suggesting to replace Keras 2 warning message from "ⓘ This example uses Keras 2" to 
```
(
        "Warning: This example uses Keras 2 and may not be compatible with the latest version of Keras. "
        "Please check out all of our <a href="https://keras.io/examples/">Keras 3 examples here</a>."
    )
```

If we are replacing this above message in banner, it would be too big, we may have to shorten the message.

jeffcarp: Thanks for the visual explanation - yes that sounds good, I think it makes sense to move the 2nd Keras 2 message up into the yellow text box so it's more prominent to users. Would the text be able to fit if we added line breaks (`<br/>`) in the right places?

sachinprasadhs: It looks like this 

Shall I commit this change? 
image

jeffcarp: Looks good, thanks!

sachinprasadhs: Pushed the change, please review it.

jeffcarp:
Thanks for your patience! Lgtm

Issue #2230, opened 2026-01-05
Keras Code Example for image similarity estimation: model's final layer sigmoid activation is inconsistent with contrastive loss (with integer margin)


chanangaza:
### Issue Type

Performance

### Source

binary

### Keras Version

3.13

### Custom Code

Yes

### Python version

3.12

### Current Behavior?

in [Image similarity estimation using a Siamese Network with a contrastive loss](https://keras.io/examples/vision/siamese_contrastive/):

Shouldn't we use Binary-Cross-Entropy Loss given the final dense classification head with Sigmoid activation? the integer 'margin' param makes no sense... either remove final dense layer and have model output raw Eucleadian distance against the contrastive-loss (integer margin) or use the Binary-Cross-Entropy loss (but then this is not an example of contrastive loss)...

### Standalone code to reproduce the issue or tutorial link

```shell
### from example...
# Provided two tensors t1 and t2
# Euclidean distance = sqrt(sum(square(t1-t2)))
def euclidean_distance(vects):
    """Find the Euclidean distance between two vectors.

    Arguments:
        vects: List containing two tensors of same length.

    Returns:
        Tensor containing euclidean distance
        (as floating point value) between vectors.
    """

    x, y = vects
    sum_square = ops.sum(ops.square(x - y), axis=1, keepdims=True)
    return ops.sqrt(ops.maximum(sum_square, keras.backend.epsilon()))


input = keras.layers.Input((28, 28, 1))
x = keras.layers.BatchNormalization()(input)
x = keras.layers.Conv2D(4, (5, 5), activation="tanh")(x)
x = keras.layers.AveragePooling2D(pool_size=(2, 2))(x)
x = keras.layers.Conv2D(16, (5, 5), activation="tanh")(x)
x = keras.layers.AveragePooling2D(pool_size=(2, 2))(x)
x = keras.layers.Flatten()(x)

x = keras.layers.BatchNormalization()(x)
x = keras.layers.Dense(10, activation="tanh")(x)
embedding_network = keras.Model(input, x)


input_1 = keras.layers.Input((28, 28, 1))
input_2 = keras.layers.Input((28, 28, 1))

# As mentioned above, Siamese Network share weights between
# tower networks (sister networks). To allow this, we will use
# same embedding network for both tower networks.
tower_1 = embedding_network(input_1)
tower_2 = embedding_network(input_2)

merge_layer = keras.layers.Lambda(euclidean_distance, output_shape=(1,))(
[...]

PR #2261, opened 2026-01-22
Fix logic error in Siamese Network: Remove inconsistent sigmoid activation for Contrastive Loss

maitry63:
This PR fixes a mathematical conflict between the model architecture and the loss function in the Siamese Network tutorial.

The model previously used a sigmoid activation on the final output. Because sigmoid squashes values between 0 and 1, the model could never output a distance greater than 1.0. Since the Contrastive Loss uses a margin of 1.0, the model was not able to push dissimilar images far enough apart to satisfy the loss function.

Removed the final 'Dense(1, activation="sigmoid")' layer so the model can output raw Euclidean distance.

Ref : https://github.com/keras-team/keras-io/issues/2230

### DIFF
--- a/examples/vision/ipynb/siamese_contrastive.ipynb
+++ b/examples/vision/ipynb/siamese_contrastive.ipynb
@@ -10,7 +10,7 @@
     "\n",
     "**Author:** Mehdi<br>\n",
     "**Date created:** 2021/05/06<br>\n",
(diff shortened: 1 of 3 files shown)


sachinprasadhs:
Change the last modified date field in .py and generate .md and .ipynb files.

PR #2277, opened 2026-02-05
[Code Example] Fix Rendering and Ordering of Brain Tumor Segmentation

innat:


innat:
@divyashreepathihalli Thanks for the feedback. Please check when you're available.

innat:
@divyashreepathihalli 

About one thing, I need your assistance. I added a `gif` file in [this](https://github.com/innat/keras-io/blob/8cc74480ed724b1fe05ac588ecccaaca60e4e276/examples/vision/brain_tumor_segmentation.py#L912-L915) location. I can see it render properly in `ipynb` file locally. But I couldn't see this gif file in [examples/vision/img/brain_tumor_segmentation](https://github.com/innat/keras-io/tree/8cc74480ed724b1fe05ac588ecccaaca60e4e276/examples/vision/img/brain_tumor_segmentation) also couldn't see it render properly in the `ipynb` [here](https://github.com/innat/keras-io/blob/8cc74480ed724b1fe05ac588ecccaaca60e4e276/examples/vision/ipynb/brain_tumor_segmentation.ipynb). This is my first time adding example and so I might overlooked something. Thanks.

sachinprasadhs:
@innat , May be try adding like this  `Gif name` instead of ``

innat:
@sachinprasadhs Just tried, same. I think it will render when finally push to keras-io, hopefully.

divyashreepathihalli (approved):

PR #2284, opened 2026-02-07
Fix VMEM OOM and benchmark issue in Pallas guide

jeffcarp:
Running the [custom kernel guide](https://keras.io/guides/define_custom_kernel/) as-is on TPU v5e will encounter a VMEM compiler OOM in newer versions of JAX:

image

This is caused by newer versions of Pallas in JAX >0.8.0 enabling [double VMEM buffering](https://docs.jax.dev/en/latest/pallas/pipelining.html#pallas-pipelining-api) by default. The guide was rendered with an older version of JAX so it didn't hit this issue.

Additionally, in the guide's benchmark, the non-Pallas vanilla JAX version is unintentionally faster:

```
Benchmarking Matrix Size: 8192x8192
------------------------------
Standard Keras (Matmul + ReLU) Average Latency: 7.811 ms
Pallas Fused (Matmul + ReLU) Average Latency: 35.039 ms
```

(Thanks to [Aditya Kane](https://www.linkedin.com/in/aditya-kane/) for pointing this out)

This PR updates the guide to tile the contracting dimension in the matmul, avoiding the VMEM OOM. After re-rendering on TPU, the benchmark correctly shows the fused Pallas kernel as being faster:

```
Benchmarking Matrix Size: 8192x8192
------------------------------
Standard Keras (Matmul + ReLU) Average Latency: 7.807 ms
Pallas Fused (Matmul + ReLU) Average Latency: 2.240 ms
```

### DIFF
--- a/guides/define_custom_kernel.py
+++ b/guides/define_custom_kernel.py
@@ -203,9 +203,9 @@ def fused_matmul(a, b):
     _, n = b.shape
 
     # Define tile sizes
-    tile_m, tile_n = 128, 128
+    tile_m, tile_k, tile_n = 128, 128, 128
     assert (
-        m % tile_m == 0 and n % tile_n == 0
+        m % tile_m == 0 and k % tile_k == 0 and n % tile_n == 0
     ), "Inputs must be multiples of 128 for this demo"
 
     return pl.pallas_call(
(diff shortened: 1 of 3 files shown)


jeffcarp:
Ah re: Gemini's comments - I didn't realize that's how BlockSpec works, let me investigate and revise a bit. Thanks!

jeffcarp:
Re: the benchmark, I wonder if it's not realistic to demonstrate this on v5e. I increased the VMEM size scoped for this Pallas kernel to circumvent the OOM via:

```
os.environ['JAX_XLA_FLAGS'] = '--xla_tpu_scoped_vmem_limit_kib=20480'
# And within pl.pallas_call:
compiler_params=pl.tpu.CompilerParams(vmem_limit_bytes=20480*1000),
```

And the Pallas kernel is still slower:

```
Benchmarking Matrix Size: 8192x8192
------------------------------
Standard Keras (Matmul + ReLU) Average Latency: 8.009 ms
Pallas Fused (Matmul + ReLU) Average Latency: 26.169 ms
```

It may be that it's impossible to beat XLA at its own game here. I may pivot the tutorial to instead show how to define an MoE layer, which should have a more robust speedup between dense and sparse versions. I'll close this PR until I have that ready.

PR #2303, opened 2026-02-26
Stylegan migration

kharshith-k:
Please find my PR for migrating StyleGAN tutorial from keras2 to keras3. Full training of the model needs to be started after the review

[Colab Notebook](https://colab.research.google.com/gist/kharshith-k/4e53898ebe1a60d651540ca9895fc171/stylegan.ipynb)


examples/generative/stylegan.py
```diff
@@ -533,87 +634,69 @@ def metrics(self):
...
+        epsilon = keras.random.uniform((batch_size, 1, 1, 1))
+        interpolates = epsilon * real_images + (1 - epsilon) * fake_images
+        pred_fake_grad = self.discriminator([interpolates, alpha])
+        gradient_penalty = 0.0
+        drift_loss = self.loss_weights["drift"] * keras.ops.mean(
+            keras.ops.concatenate([pred_fake, pred_real], axis=0) ** 2
+        )
+        d_loss = loss_fake + loss_real + gradient_penalty + drift_loss
+        self.add_loss(g_loss)
+        self.add_loss(d_loss)
```
sachinprasadhs: This is wrong approach to combine both Generator and Discriminator losses.
GANs require alternating updates for the Discriminator and Generator using separate optimizers. So this approach will block that behavior.

examples/generative/stylegan.py
```diff
@@ -2,9 +2,8 @@
 Title: Face image generation with StyleGAN
 Author: [Soon-Yau Cheong](https://www.linkedin.com/in/soonyau/)
 Date created: 2021/07/01
-Last modified: 2021/07/01
+Last modified: 2026/03/06
 Description: Implementation of StyleGAN for image generation.
```
sachinprasadhs, line 6: Why accelerator is removed?

examples/generative/stylegan.py
```diff
@@ -654,7 +746,7 @@ def call(self, inputs: dict()):
 """
 
 START_RES = 4
-TARGET_RES = 128
+TARGET_RES = 16  # Training to 16x16 for quick demonstration
```
sachinprasadhs: How much difference in training we will observe if we do 32x32 or more?
16x16 is significant reduce!

examples/generative/stylegan.py
```diff
@@ -238,24 +254,88 @@ class AdaIN(layers.Layer):
...
     def call(self, inputs):
         x, w = inputs
-        ys = tf.reshape(self.dense_1(w), (-1, 1, 1, self.x_channels))
-        yb = tf.reshape(self.dense_2(w), (-1, 1, 1, self.x_channels))
+        ys = keras.ops.reshape(self.dense_1(w), (-1, 1, 1, self.x_channels))
+        yb = keras.ops.reshape(self.dense_2(w), (-1, 1, 1, self.x_channels))
         return ys * x + yb
 
 
+class SpectralNormalization(layers.Wrapper):
```
sachinprasadhs: Why this normalization is added now? Was there any training stability issue without this?

examples/generative/stylegan.py
```diff
@@ -671,18 +763,75 @@ def train(
...
+    early_stopping_patience=3,  # Stop if no improvement after N phases
+    save_best_only=True,  # Only save checkpoints when model improves
+    d_updates_per_g_update=2,  # Update discriminator every N generator updates (1:2 ratio)
 ):
-    opt_cfg = {"learning_rate": 1e-3, "beta_1": 0.0, "beta_2": 0.99, "epsilon": 1e-8}
+    # Stability improvements:
+    # - Spectral normalization on all discriminator conv layers (NEW - addresses fundamental instability)
+    # - Discriminator update ratio (1:2) - balances G/D training dynamics
+    # - Combined approach: spectral norm provides architectural stability + update ratio balances training
+    # Previous attempts: 1:2 ratio alone delayed divergence 3-4x but still diverged by step 1500
```
sachinprasadhs: This looks like some debug comments from multiple previous runs, avoid such comments and keep only those comments which will be beneficial for the tutorial.

sachinprasadhs (commented):
Added few comments and also Gemini comments are valid. Please address those. Thanks

examples/generative/stylegan.py
kharshith-k: Sure, I've fixed the code to compute generator and discriminator losses separately. Please review and let me know if its correct

examples/generative/stylegan.py
kharshith-k: I think I've removed it by mistake

examples/generative/stylegan.py
kharshith-k: with steps_per_epoch set to 100, training takes an hour for 64x64 resolution

examples/generative/stylegan.py
kharshith-k: I removed it as it was not needed after modifying the approach of updating the losses

hertschuh (changes requested):
There are many unrelated changes to 100+ files. Can you fix them?

Thanks!

examples/generative/stylegan.py
```diff
@@ -238,21 +254,22 @@ class AdaIN(layers.Layer):
...
         w_shape = input_shapes[1]
-
         self.w_channels = w_shape[-1]
         self.x_channels = x_shape[-1]
-
-        self.dense_1 = EqualizedDense(self.x_channels, gain=1)
-        self.dense_2 = EqualizedDense(self.x_channels, gain=1)
+        if self.dense_1 is None:
+            self.dense_1 = EqualizedDense(self.x_channels, gain=1)
+            self.dense_2 = EqualizedDense(self.x_channels, gain=1)
```
hertschuh, line 267: You shouldn't need the `if` here, `build` is only called once.

examples/generative/ipynb/stylegan.ipynb
```diff
@@ -828,349 +960,217 @@
...
...v1In8wkjRdhaP8umb3U3bPfug2Q+8azu/vbrWKC7YR/vpsdHlepu2A3mAZkn+44/GtBPjhuYrLv5Tbzu8C5YoCfNi75j57B+3ty1GTKf7obLPHu87qoNwi9MAAAssGACAGCBBRMAAAssmAAAWGDBBADAguO6zXaW6ZNOoR4dV6nzPbojyrh/eNrPRu3R+wNuj9UbfW4uekHmaQu+0vUahoS2u00NeI0rAzoQg96v4M5c3z6d+jU2Btz2vB5uosQWp38V9O7xaDLlDTo4ujNxVXWWzMc9u1iXO+Df27W3rHfiqu6SvbN+sMydqIAu4CveemN6Fct65b2fkpfPHKY7WZc+oztfTWToG50YsJdspVmr723WyXyt0Xuf7vbNGcfoSZDTQ15u6n/U+ZgWuus9zRT75uhjetKN0x3GzjXdcV28dazMM337kNYFfCY6B33W9urYDAv8zHrfz5yAz3yfHH19t0UyTu33hsw3d5wT+iC99Gt8M1V3wWct3ibzCyUJMo+c6J+n4bJeWYRq4TWmUwc9F+NnTJW5O2+5b86skvUWh/S1Nxk4S/9WSygaJnPjfump98mgMFlv7KWf5eWO3o7czPpFd9UWuRXsJQsAwL/FggkAgAUWTAAALLBgAgBggQUTAAALze8l60zQ+cKAvRzz82V83ej9TG/yHW+Pe0SO22j0ffvM1Y9RPHeozKep5rj6g/omRncCH+m6Q+Z9b0sIuI9PQOdfS/0Pvk3UmXp9n+yHdP66r9zxKj3uuaMybjv8HpmXZpTLPMUf/Kq7Ydu16yhzp01AN2xSnM59amtv6BMBnaJDPtN7bJqbO+l8fmhUqRvrjNk3ScYz63Vu9D+5t/ZdjM7LC/fLvHq27nJNC/MeO8v0vqCTC/w7iDbZEDte5gen2u3tentgp77eX9lM0RuiRgfcx9//HDe2ixy35/szMp/+gf4+WRb+tX6+JaHR1Z/00FbRjTJ3TYLMV8/Re59Onug97v3qNTnu+BbdDXtJbx9sOp9arU+Y5d7D+xLkqKy9/m/5Js4l3bbqzn84oJ7XuG9WBZxpCMj19/xuc8Wq3t/4hQkAgAUWTAAALLBgAgBggQUTAAALLJgAAFj4p71kAQCA4RcmAABWWDABALDAggkAgAUWTAAALLBgAgBggQUTAAALfwJRSaJ0RiM7ywAAAABJRU5ErkJggg==\n",
```
hertschuh, line 932: You'll need to generate the images too.

PR #2307, opened 2026-03-02
Suppress training output warnings and remove intermediate epoch PNG visualizations in Focal Modulation Network tutorial

LakshmiKalaKadali:


sachinprasadhs (commented):
class TrainMonitor is unused, if it is not required then remove it. 
Rest all looks good.

sachinprasadhs:
You just removed from .py file only.
Can you just regenerate files again. Thanks!

sachinprasadhs (approved):

PR #2320, opened 2026-03-31
Migration of Finetuning Stable diffusion via textual inversion from Keras 2 to Keras 3 

kharshith-k:
Please find my PR for migrating Finetuning Stable Diffusion using textual inversion tutorial from keras2 to keras3

[Colab Gist](https://colab.research.google.com/gist/kharshith-k/14d32d7ea3bc23edb0ef689e5a5e033d/fine_tune_via_textual_inversion.ipynb)


laxmareddyp (changes requested):
I don't see this in the examples, and there doesn't seem to be a corresponding .py file.

I could find only this tutorial which you attached in the PR description. https://github.com/keras-team/keras-io/blob/master/examples/generative/fine_tune_via_textual_inversion.py

Why the new .ipynb and .md files are created in this PR   , if it is just migration PR ?

PR #2321, opened 2026-04-01
[Examples] Add Context-Aware Music Retrieval tutorial using Yambda and JAX DataParallel

rishiraj:
**Description**
This PR adds a new, production-oriented retrieval example to the KerasRS examples gallery: **Context-Aware Music Retrieval with Yambda and JAX Data Parallelism**. 

While existing examples demonstrate standard user-item Two-Tower models, real-world recommender systems must dynamically adapt to user context. This tutorial tackles this problem using the recently released [Yambda-5B](https://huggingface.co/datasets/yandex/yambda) industrial-scale music dataset, leveraging its explicit `is_organic` flag to alter recommendations based on whether a user is actively exploring (organic) or passively listening (algorithmic).

**Key Educational Takeaways for Keras Users:**
* **Contextual Query Fusion:** Demonstrates how to dynamically shift a user's embedding by fusing it with categorical context embeddings in the Query Tower.
* **Overcoming the "Curse of Scale" in Validation:** Shows how to decouple the `BruteForceRetrieval` index from the `call()` method to prevent validation-loop freezing and OOM crashes when evaluating against massive candidate pools.
* **Vocabulary Compression:** Uses `keras.layers.IntegerLookup` to compress sparse 7-digit item IDs into a contiguous range, dropping the TPU HBM footprint from >15GB to just megabytes and preventing XLA memory exhaustion.
* **High-Performance `tf.data` Pipelines:** Utilizes Hugging Face's `to_tf_dataset()` with `drop_remainder=True` to stream data directly from Apache Arrow files, maintaining the strict static shapes required for JAX/TPU XLA compilation.
* **TPU Scaling:** Implements `keras.distribution.DataParallel` for seamless multi-core TPU training using the JAX backend.

**Testing Performed:**
- [x] Verified end-to-end execution on a Google Colab TPU v5e-1 environment.
- [x] Confirmed XLA compilation successfully caches the graph (Epoch 1: ~150s, Epoch 2+: ~3s).
- [x] Verified that model inference outputs distinct item recommendations for the exact same user based solely on toggling the `is_organic` context flag.
- [x] Validated reverse-lookup mapping from internal compressed IDs back to original Yambda IDs.


rishiraj:
Thanks for the automated review! I've incorporated the suggestions regarding the loss reduction (`reduction="sum_over_batch_size"`) and removed the unused `jax-tpu-embedding` dependency to keep the tutorial clean. 

Regarding the `build()` method refactor: I've implemented the suggestion to let `super().build(input_shape)` handle the dictionary unpacking for the layers used in `call()`. However, there's a subtle architectural detail in this Two-Tower setup—the candidate tower (`self.item_embedding`) is only ever invoked inside `compute_loss()`, not within the forward `call()` pass. 

Because it is decoupled from the main forward pass, `super().build()` cannot automatically trace and build it. Attempting to access `self.item_embedding.embeddings` immediately after `super().build()` results in a `RuntimeError: You must build the layer before accessing embeddings`.

To resolve this while keeping the code compliant with Keras 3 standards, I explicitly build the candidate embedding with a 1D tuple shape `(None,)` before passing its weights to the `BruteForceRetrieval` layer, and then let `super().build(input_shape)` cleanly handle the rest of the query tower components. 

The tutorial now trains and evaluates flawlessly on the TPU! @abheesht17 @sachinprasadhs @hertschuh Could you please take a look at this PR when you have a moment? Thanks!

hertschuh (approved):
This is great, thanks!

Are you able to generate the colab notebook and the rendered md file?

This example would also benefit from using KerasRS' [DistributedEmbedding](https://keras.io/keras_rs/api/embedding_layers/distributed_embedding/). It should allow you to load the full embedding and train on the full dataset. Something to consider for an update to this as a separate PR.

rishiraj:
@hertschuh generated the colab notebook and the rendered md file. also thanks for the suggestion, i will work on it in a future PR.

hertschuh (approved):
Thanks!

PR #2329, opened 2026-04-02
Migration of Consistency training with supervision in keras 3

maitry63:
This PR migrates the Consistency training with supervision tutorial from Keras 2 to Keras 3. The implementation is now fully backend-agnostic, allowing it to run seamlessly on JAX, PyTorch, and TensorFlow.

Notebook - [Colab](https://colab.sandbox.google.com/gist/maitry63/965b301b967f3e679822b2844a1f6529/copy_consistency-training-with-supervision.ipynb)


examples/vision/consistency_training.py
```diff
@@ -216,34 +230,58 @@ def get_training_model(num_classes=10):
...
+        """
+        Update SWA weights after each epoch
+        """
+        if self.swa_weights is None:
+            self.swa_weights = [ops.cast(w, "float32") for w in model.get_weights()]
+            self.n = 1
+        else:
+            for i, w in enumerate(model.get_weights()):
+                self.swa_weights[i] = (self.swa_weights[i] * self.n + w) / (self.n + 1)
+            self.n += 1
```
hertschuh: This won't work. It will never get called as `Adam` doesn't have a `apply_swa` callback.

examples/vision/consistency_training.py
```diff
@@ -216,34 +230,58 @@ def get_training_model(num_classes=10):
...
-SWA = tfa.optimizers.SWA
+
+class SWA(Adam):
+    """
+    Simple Stochastic Weight Averaging wrapper for Adam optimizer.
+    """
+
+    def __init__(self, optimizer, start_epoch=0, **kwargs):
+        super().__init__(**kwargs)
+        self.base_optimizer = optimizer
```
hertschuh: There's both a super class of Adam and a `self.base_optimizer` of Adam. But `self.base_optimizer` is never used.

examples/vision/consistency_training.py
```diff
@@ -252,121 +290,77 @@ def get_training_model(num_classes=10):
...
-        # Update weights
-        self.optimizer.apply_gradients(zip(gradients, trainable_vars))
+# Backend-agnostic SelfTrainingModel
+class SelfTrainingModel(keras.Model):
+    def __init__(self, student, teacher, temperature=10):
+        super().__init__()
+        self.student = student
+        self.teacher = teacher
+        self.temperature = temperature
+        self.teacher.trainable = False  # freeze teacher
```
hertschuh: We have API for model distillation now, please use that: https://keras.io/api/models/distillation/distiller/

hertschuh (changes requested):

examples/vision/consistency_training.py
maitry63: Updated the Distiller Class

examples/vision/consistency_training.py
maitry63: Updated

examples/vision/consistency_training.py
maitry63: corrected

examples/vision/consistency_training.py
hertschuh: But you should just use the keras Distiller class instead of reimplementing one. (Unless there is a reason why you cannot).

maitry63: Thank you for your feedback. Actually, I used my own `DistillationModel` because there is no conventional way to perform knowledge distillation with this configuration.

In this tutorial, the teacher receives `clean images`, whereas the student receives `noisy images` from the same set of inputs. Hence, we need to dual inputs and also maintain consistency across various representations of the same sample, which is not possible via the use of Keras `Distiller` class.
Even if I used `Distiller`, I had to override its `train_step `function.

hertschuh (approved):
Thanks!

Can you generate the ipynb and md files?

hertschuh (approved):
Thanks!

PR #2382, opened 2026-06-23
[keras.io] ConvLSTM next-frame prediction: use causal padding to prevent data leakage

### LINKED ISSUE #1003: Data leakage in Next-Frame Video Prediction with Convolutional LSTMs
In the Keras tutorial "Next-Frame Video Prediction with Convolutional LSTMs" (https://keras.io/examples/vision/conv_lstm/) I've noticed that there seems to be a data leakage problem. Namely, the 3D convolution has direct access to the frame it is trying to predict, so the results are unreasonably good. The correct way to implement this would be to use causal padding, which would inhibit the model's ability to "cheat" by not providing him directly with the value that it should predict. Let me know if I haven't missed something.

pctablet505:
Fixes keras-team/keras-io#1003

The final `Conv3D` layer in the ConvLSTM next-frame prediction example used `padding="same"`, which allowed the kernel to access the target frame it was supposed to predict and caused data leakage.

This change replaces that layer with a `ZeroPadding3D` + `Conv3D(padding="valid")` pair that applies causal padding along the time axis:

- `ZeroPadding3D(padding=((2, 0), (1, 1), (1, 1)))` pads only past frames in time and symmetrically in space.
- `Conv3D(filters=1, kernel_size=(3, 3, 3), activation="sigmoid", padding="valid")` keeps the same output shape while no longer looking at the frame being predicted.

The Model Construction markdown block now explains the use of causal padding.

Verification:
- `python -m py_compile examples/vision/conv_lstm.py` passed.
- `python scripts/autogen.py add_example vision/conv_lstm` was run with a 10-minute timeout and timed out while training the first epoch (reached batch 111/180). The architecture compiled successfully and training began, but the full example could not finish on CPU within the timeout. Generated `.md`, `.ipynb`, and image files were not committed.

### DIFF
--- a/examples/vision/conv_lstm.py
+++ b/examples/vision/conv_lstm.py
@@ -2,7 +2,7 @@
 Title: Next-Frame Video Prediction with Convolutional LSTMs
 Author: [Amogh Joshi](https://github.com/amogh7joshi)
 Date created: 2021/06/02
-Last modified: 2023/11/10
+Last modified: 2026/06/23
 Description: How to build and train a convolutional LSTM model for next-frame video prediction.
 Accelerator: GPU
 """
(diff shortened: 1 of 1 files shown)


pctablet505:
Agent review feedback has been incorporated: clarified the causal-padding explanation to say 'past and current frames' instead of just 'past frames'.

Note: this PR branch is based on a fork master that contains unrelated LiteRT guide work and a README fork note, so the GitHub diff shows those files as well. The actual change for issue #1003 is limited to `examples/vision/conv_lstm.py`. A maintainer may want to rebase/cherry-pick commit `b5b0d5c3` from this branch onto current upstream master to get a clean diff.

pctablet505:
Closing in favor of a clean PR branched from current keras-team/keras-io master.

pctablet505:
Reopened. A clean rebase onto current keras-team/keras-io master is blocked because the GitHub token used here lacks the `workflow` scope required to push branches that include updated `.github/workflows` files. The fix itself (commit `b5b0d5c3`) is ready; it just needs to be pushed from a fork/account that can sync with upstream.

pctablet505:
Rebased onto current `keras-team/keras-io:master`. The diff is now limited to `examples/vision/conv_lstm.py` only.

PR #2393, opened 2026-07-13
Migrate EfficientNet fine-tuning tutorial to Keras 3

### LINKED ISSUE #2385: Refactoring needed for `unfreeze_model(model)` function in "Image classification via fine-tuning with EfficientNet" tutorial, "Transfer learning from pre-trained weights" section, second step
### Bugs and code smells

Hi, this function ([source code][1]) has a few things wrong:

| ❌ What's wrong | ✅ What should be done about it |
| - | - |
| - **Bug:**  The `model` parameter on which operations are applied is not returned at the end. | - The `model` parameter should be returned at the end, and assigned to a new object. |
| - **Code smells:** The number of layers to unfreeze is a magic number, which is a notorious code smell. Other magic values are: learning rate, loss function, metrics. | - All these should be exposed as function parameters. |

### And a feature request

Another nice feature to have would be if the function could unfreeze not only a slice of layers but also a part of their name. See existing issue #995 left in July 2022.

### ❌ How the function looks now

```py
def unfreeze_model(model):
    # We unfreeze the top 20 layers while leaving BatchNorm layers frozen
    for layer in model.layers[-20:]:
        if not isinstance(layer, layers.BatchNormalization):
            layer.trainable = True

    optimizer = keras.optimizers.Adam(learning_rate=1e-5)
    model.compile(
        optimizer=optimizer, loss="categorical_crossentropy", metrics=["accuracy"]
    )


unfreeze_model(model)
```

### ✅ How it should look

```py
def unfreeze_model(
    model: keras.Model,
    layers_to_unfreeze: int | str = 20,
    learning_rate: float = 1e-5,
    loss_func_name: str = "categorical_crossentropy",
    metrics: list[str] | None = None,
) -> keras.Model:
    """Unfreeze part of `model` and recompile it for fine-tuning.

    Args:
        model: A `keras.Model` instance to unfreeze in place.
        layers_to_unfreeze: Either an `int` giving the number of layers,
            counted from the end of `model.layers`, to unfreeze, or a `str`
            substring to match against layer names -- the first matching
            layer and every layer after it (in `model.layers` order) are
            unfrozen. Use a string like `"block7"` to respect EfficientNet's
            residual block boundaries instead of an arbitrary layer count.
            Defaults to `20`.
        learning_rate: Learning rate for the fine-tuning `Adam` optimizer.
            Defaults to `1e-5`.
        loss: Loss function passed to `model.compile()`. Defaults to
            `"categorical_crossentropy"`.
        metrics: List of metrics passed to `model.compile()`. Defaults to
            `["accuracy"]`.

    Returns:
        `model`, with the selected layers unfrozen (except
        `BatchNormalization` layers, which are always kept frozen) and
        recompiled with the new optimizer/loss/metrics.
    """
    metrics = metrics or ["accuracy"]

    # Auto-detect a nested backbone; fall back to `model` itself if flat
    base_model = next((L for L in model.layers if isinstance(L, keras.Model)), model)
    # container flag must be True for any child to train:
    if base_model is not model:
        base_model.trainable = (
            True
        )

    # Reset to a known state so repeated calls are predictable
    for layer in base_model.layers:
        if not isinstance(layer, layers.BatchNormalization):
            layer.trainable = False

    if isinstance(layers_to_unfreeze, int) and layers_to_unfreeze > 0:
        selected = base_model.layers[-layers_to_unfreeze:]
    elif isinstance(layers_to_unfreeze, str):
        selected = [L for L in base_model.layers if layers_to_unfreeze in L.name]
        if not selected:
            raise ValueError(f"No layer name contains {layers_to_unfreeze!r}.")
    else:
        raise TypeError(
            f"layers_to_unfreeze must be a positive int or str, got {layers_to_unfreeze!r}"
        )

    for layer in selected:
        if not isinstance(layer, layers.BatchNormalization):
            layer.trainable = True

    model.compile(
        optimizer=keras.optimizers.Adam(learning_rate=learning_rate),
        loss=loss_func_name,
        metrics=metrics,
    )
    return model


model = unfreeze_model(model)
```

[1]: https://github.com/keras-team/keras-io/blob/transfer_learning/examples/vision/image_classification_efficientnet_fine_tuning.py#L393-L405

kharshith-k:
- Replace TensorFlow-specific code with backend-agnostic Keras 3 APIs
- Remove tensorflow_datasets dependency, implement direct HTTP download from Stanford servers
- Replace tf.data.Dataset pipeline with keras.utils.PyDataset classes
- Replace tf.image.resize and tf.one_hot with keras.ops equivalents
- Enhance unfreeze_model() function with parameterization (fixes #2385)
  - Add parameters: layers_to_unfreeze, learning_rate, loss, metrics
  - Support int (last N layers) and str (block name) patterns
  - Add return statement and update call site
  - Add comprehensive docstring with input validation
- Update all documentation to reflect Keras 3 backend-agnostic approach
- Standardize header format and update dates to 2026/07/13
- Apply changes consistently across .ipynb, .py, and .md files

Here's the [Github Gist](https://colab.research.google.com/gist/kharshith-k/b47f8fb9904d1b5e2421dcde277a48d2/image_classification_efficientnet_fine_tuning.ipynb) for the migration

This PR Also implements the fixes for this [issue](https://github.com/keras-team/keras-io/issues/2385)


examples/vision/md/image_classification_efficientnet_fine_tuning.md
kirisakow, line 1573: Empty line at the end of the fenced block should be dropped, or it will cause the following line (`</div>`) to be rendered as code

kirisakow, line 1579: No need to skip so many blank lines. If you need more than one blank line below, use one or more `<br/>`.

kirisakow (changes requested):

kirisakow:
Plus [an extended comment](https://github.com/keras-team/keras-io/issues/2385#issuecomment-5006551276) regarding `unfreeze_model()` function

examples/vision/image_classification_efficientnet_fine_tuning.py
```diff
@@ -391,53 +552,111 @@ def build_model(num_classes):
...
+    Returns:
+        `model`, with the selected layers unfrozen (except
+        `BatchNormalization` layers, which are always kept frozen) and
+        recompiled with the new optimizer/loss/metrics.
+    """
+    if metrics is None:
+        metrics = ["accuracy"]
+
+    # Access the nested EfficientNetB0 base model
+    base_model = model.get_layer("efficientnetb0")
```
maitry63, line 589: I think, this works now because the model is named "efficientnetb0". But if the name changes later, `unfreeze_model() `will break.

@kharshith-k , Could you please avoid using the name  hardcoded here, make it specific to this EfficientNetB0 model?

examples/vision/image_classification_efficientnet_fine_tuning.py
```diff
@@ -391,53 +552,111 @@ def build_model(num_classes):
...
+        unfreeze_from = None
+        for i, layer in enumerate(base_model.layers):
+            if layers_to_unfreeze in layer.name:
+                unfreeze_from = i
+                break
+        if unfreeze_from is None:
+            raise ValueError(f"No layer name contains {layers_to_unfreeze!r}.")
+        layers_to_process = base_model.layers[unfreeze_from:]
+    elif isinstance(layers_to_unfreeze, int):
+        n_layers_to_unfreeze = min(layers_to_unfreeze, len(base_model.layers))
```
maitry63, line 606: Here, If `layers_to_unfreeze=0`, `base_model.layers[-0:]` means all layers are unfrozen. Negative values can also cause unexpected results.

Could you check that `layers_to_unfreeze` is an integer greater than 0 before slicing?

examples/vision/image_classification_efficientnet_fine_tuning.py
```diff
@@ -391,53 +552,111 @@ def build_model(num_classes):
...
+        loss: Loss function passed to `model.compile()`. Defaults to
+            `"categorical_crossentropy"`.
+        metrics: List of metrics passed to `model.compile()`. Defaults to
+            `["accuracy"]`.
+
+    Returns:
+        `model`, with the selected layers unfrozen (except
+        `BatchNormalization` layers, which are always kept frozen) and
+        recompiled with the new optimizer/loss/metrics.
+    """
```
maitry63, line 584: In this comment , it says the number is counted from the end of `model.layers,` but the code uses `base_model.layers` instead.

Can you update the docstring to match the code?

kirisakow:
Thank you for the update. 

Can you guys please tell me how long is it going to take for the update to be published [online](https://keras.io/examples/vision/image_classification_efficientnet_fine_tuning/)?

@kharshith-k @maitry63

examples/vision/image_classification_efficientnet_fine_tuning.py
laxmareddyp, line 64: Since this tutorial is being migrated to Keras 3, update these documentation code examples to use the Keras 3 import path.

examples/vision/image_classification_efficientnet_fine_tuning.py
```diff
@@ -123,10 +123,13 @@
 ## Setup and data loading
 """
 
+import os
```
laxmareddyp: un used import, remove it.

examples/vision/image_classification_efficientnet_fine_tuning.py
```diff
@@ -139,40 +142,130 @@
...
+with protobuf version incompatibilities).
 """
 
-dataset_name = "stanford_dogs"
-(ds_train, ds_test), ds_info = tfds.load(
-    dataset_name, split=["train", "test"], with_info=True, as_supervised=True
-)
-NUM_CLASSES = ds_info.features["label"].num_classes
+# Download and extract Stanford Dogs dataset
+import scipy.io
```
laxmareddyp: move inline imports to top.

examples/vision/image_classification_efficientnet_fine_tuning.py
laxmareddyp: The link  points to a TensorFlow TPU repo that may be stale. Consider whether this link is still useful or should be removed/updated?

laxmareddyp: arbitray → arbitrary

examples/vision/image_classification_efficientnet_fine_tuning.py
```diff
@@ -139,40 +142,130 @@
...
-ds_test = ds_test.map(lambda image, label: (tf.image.resize(image, size), label))
+def download_and_extract(url, extract_to):
+    filename = url.split("/")[-1]
+    filepath = data_dir / filename
+    if not filepath.exists():
+        print(f"Downloading {filename}...")
+        urllib.request.urlretrieve(url, filepath)
+        print(f"Extracting {filename}...")
+        with tarfile.open(filepath, "r") as tar:
+            tar.extractall(extract_to)
```
laxmareddyp: without a filter argument triggers a DeprecationWarning on Python ≥ 3.12.   Add filter="data" for safe extraction

 tar.extractall(extract_to, filter="data")

laxmareddyp (changes requested):
Thanks for the PR..! Left few comments. Once  address comemnts good to merge.

laxmareddyp (approved):
Thanks..! LGTM

PR #2400, opened 2026-07-23
Fix zizmor unpinned-uses warnings

kharshith-k:
This PR fixes security warnings raised by [`zizmor`](https://github.com/woodruffw/zizmor).

Specifically, this resolves the `unpinned-uses` warnings. While our external actions were already safely pinned via immutable commit SHAs, tools like `zizmor` and OSSF Scorecard also require the accompanying version comments to follow a strict Semantic Versioning format (e.g., `# v6.0.0` instead of `# v6`). If they are not strictly formatted, the linters consider them imprecise/mutable tags.

### Changes Made:
Updated the major-version comments to strict `.0.0` semver format across all 7 workflow files:
- `continuous_integration.yml`
- `auto-assignment.yaml`
- `scorecard.yml`
- `csat.yaml`
- `upload.yml`
- `zizmor.yml`
- `stale-issues-pr.yml`

### DIFF
--- a/.github/workflows/auto-assignment.yaml
+++ b/.github/workflows/auto-assignment.yaml
@@ -16,7 +16,7 @@ jobs:
       issues: write
       pull-requests: write
     steps:
-      - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.0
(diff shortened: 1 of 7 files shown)


laxmareddyp (approved):
LGTM, Thanks

